PR #7, opened 2017-11-14
Resolve #1 Add wait condition on create and update to make sure the s…

FranOis:
…ervice is RUNNING

PR #10, opened 2017-11-14
Kafka resource

FranOis:
Had to base this PR from https://github.com/jelmersnoeck/terraform-provider-aiven/pull/7 Will update once #7 is merged.

This is to give an idea what it would look like. There is most likely a way to set this up so that we can leverage some form of inheritance/compositing. Right now it's mostly copy paste, but we've got plans to add the kafka specific functions such as specifying the default ACLs or enabling or disabling various kafka services. This layer will convert the terraform Schema into the map provided to the create service field `user_config` so this will not leak into the aiven client library.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -23,10 +23,11 @@ func Provider() *schema.Provider {
 		},
 
 		ResourcesMap: map[string]*schema.Resource{
-			"aiven_project":      resourceProject(),
-			"aiven_service":      resourceService(),
-			"aiven_database":     resourceDatabase(),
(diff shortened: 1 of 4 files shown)


FranOis:
@jelmersnoeck I'm back working on this at work. Any news for this PR?

FranOis:
@benjigoldberg @jelmersnoeck  if there is anything you like / want from that PR, feel free to steal it, I'm not part of motivateco anymore, so I can't update that PR.

PR #17, opened 2018-07-20
Add support for User Config and Kafka User Config checking.

benjigoldberg:
@jelmersnoeck integrates support for user config into the Aiven provider. Before merging this I'll have to either update the dep to point to master of the aiven repo, or we'll need to cut a `0.1.0` or `0.0.3` in the Aiven repo.

One other note -- I had to add some type transformations on the user config because [Terraform converts bools to strings](https://www.terraform.io/docs/configuration/variables.html#booleans). The Aiven API uses JSON Schema or something similar to type check values, and was thus rejecting the API requests without the type munging you see here.

Its also worth noting that Aiven doesn't document whats in the `user_config` for each service. We're reaching out to them to see if we can get a list of whats officially documented without having to reverse-engineer the API w/ list/get on the service.

### DIFF
--- a/resource_service.go
+++ b/resource_service.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/jelmersnoeck/aiven"
@@ -14,12 +15,11 @@ func resourceService() *schema.Resource {
 		Update: resourceServiceUpdate,
 		Delete: resourceServiceDelete,
 
-		// TODO: add user config
 		Schema: map[string]*schema.Schema{
 			"project": {
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: "Target cloud",
+				Description: "Target project",
 				ForceNew:    true,
 			},
 			"cloud": {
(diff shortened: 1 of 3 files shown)


benjigoldberg:
We'll also need these changs merged before the tag - https://github.com/jelmersnoeck/aiven/pull/15

resource_service.go
```diff
@@ -178,3 +184,46 @@ func resourceServiceWait(d *schema.ResourceData, m interface{}) error {
...
+	if ok {
+		newUserConfig["schema_registry"] = stringToBool(schemaRegistryInterface.(string))
+	}
+	return newUserConfig
+}
+
+func stringToBool(maybeBool string) bool {
+	if maybeBool == "1" {
+		return true
+	} else if strings.ToLower(maybeBool) == "true" {
```
FranOis: detail, but I prefer || rather than else if

FranOis:
If you update dependencies to point to the right sha, it should build now.

benjigoldberg:
@FranOis thanks for the review! Should we cut a new release in the aiven repo and have me point to that? Or would you prefer I use the SHA?

FranOis:
I'll let @jelmersnoeck decide on the cutting of new versions. So if you need this ASAP, I'd use the SHA and keep on developing and then a new release could be made of both the client and the provider.

benjigoldberg:
@FranOis sounds good 👍! I cleaned up the bool checker and pinned the version as suggested.

Thanks for the quick review

Issue #3, opened 2017-11-14
Service specific resources


FranOis:
This is more up for a general design discussion. While I like the flexibility of having a very generic service resource that allows this provider to support any new service that aiven could add to their offering without having to write more code. The downside to this is that you have to manage all service type in the same file or do some kind of logic switch between type plus managing a giant map of possible configuration that depends on the service type. This causes issues with clarity of use and makes it hard to understand exactly what's possible to do and what is not possible to do on a service basis.

My proposal is as follows.
1. Keep the very generic service resource for backward compatibility and for future work by aiven and to support any service that are not yet implemented in this provider.
2. Implement a service type specific resource for each offering of aiven.

For example:
You would still have the service resource as it exists now, but we would also have one for Kafke, Postgres, ElasticSearch etc.

jelmersnoeck:
Thanks for this (and all the PRs!).

I will have a deeper look into this and get back to you about it, but in general I like the idea.

rikonen:
A new implementation that was merged today keeps the generic service resource for all services but there are `<service_type>_user_config` maps that define any attributes that are specific to certain service, which is (mostly) automatically generated and doesn't need manual maintenance when new service type specific attributes are added (apart from re-running a script to regenerate some files). This would seem like decent enough compromise. The service resource, which now supports all service types and attributes, is less than 600 lines of code.

PR #20, opened 2018-07-26
Wait for Kafka auxillary services to start up

briancharous:
@jelmersnoeck @FranOis 

When Aiven creates a Kafka service with Kafka Connect, and/or, Kafka Rest, and/or schema registry enabled, Aiven will report that Kafka is ready potentially before any of those peripheral services are actually ready. This PR ensures that when a resource is a Kafka service the resource shouldn't be considered ready until all the peripheral services are healthy; if a service is reachable, it is considered ready. This is a feature @benjigoldberg and I need so that resources that are dependent on Kafka Connect being ready don't fail immediately after Terraform creates the Kafka service in Aiven.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea
 terraform-provider-aiven
 terraform.tfvars
 *.tfstate*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 terraform-provider-aiven
 terraform.tfvars
 *.tfstate*
--- a/service_change.go
+++ b/service_change.go
@@ -40,10 +40,8 @@ func (w *ServiceChangeWaiter) RefreshFunc() resource.StateRefreshFunc {
 		// If this is a Kafka service, wait for its component parts (i.e. Kafka Connect,
 		// Kafka REST, and Schema Registry) to be ready
 		state := service.State
-		if state == aivenTargetState && service.Type == "kafka" {
-			if !kafkaServicesReady(service) {
-				state = aivenKafkaServicesStartingState
-			}
+		if !kafkaServicesReady(service) {
+			state = aivenKafkaServicesStartingState
 		}
 
 		return service, state, nil
@@ -53,6 +51,14 @@ func (w *ServiceChangeWaiter) RefreshFunc() resource.StateRefreshFunc {
 // If any of Kafka Rest, Schema Registry, and Kafka Connect are enabled, refresh
 // their state to check if they're ready
 func kafkaServicesReady(service *aiven.Service) bool {
+	// Check if the service is a Kafka service and Kafka itself is ready
+	if service.Type != "kafka" {
+		return true
+	}
+	if service.State != aivenTargetState {
+		return false
+	}
+
 	kafkaRestEnabled, schemaRegistryEnabled, kafkaConnectEnabled := false, false, false
 	userConfig := service.UserConfig.(map[string]interface{})
 	if enabled, ok := userConfig["kafka_rest"]; ok {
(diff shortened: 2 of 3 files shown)


service_change.go
```diff
@@ -30,15 +37,62 @@ func (w *ServiceChangeWaiter) RefreshFunc() resource.StateRefreshFunc {
...
+	if enabled, ok := userConfig["kafka_connect"]; ok {
+		kafkaConnectEnabled = enabled.(bool)
 	}
+
+	if !kafkaRestEnabled && !schemaRegistryEnabled && !kafkaConnectEnabled {
+		return true
+	}
+
+	if kafkaRestEnabled {
+		if _, err := http.Get(service.ConnectionInfo.KafkaRestURI); err != nil {
```
FranOis: Not sure what Go will do with that `_`. From the doc: `Caller should close resp.Body when done reading from it.` Seems like assigning that to `_` will make the `rest.Body` leak.

https://golang.org/pkg/net/http/#Get

service_change.go
```diff
@@ -30,15 +37,62 @@ func (w *ServiceChangeWaiter) RefreshFunc() resource.StateRefreshFunc {
 			return nil, "", err
 		}
 
-		return service, service.State, nil
+		// If this is a Kafka service, wait for its component parts (i.e. Kafka Connect,
+		// Kafka REST, and Schema Registry) to be ready
+		state := service.State
+		if state == aivenTargetState && service.Type == "kafka" {
```
FranOis: @jelmersnoeck I'm not against this but this seems to tie into Issue #3 Feel like if we want this kind of behaviour that it should live inside a Kafka specific resource and this service resource has limited functionality but works with everything that is not more fleshed out in the terraform provider.

Any thoughts? cc: @briancharous && @benjigoldberg

jelmersnoeck: +1 on making a Kafka specific resource, but I am not opposed in having this logic in the general resource either.

I would suggest taking out the conditional here and putting all of that in the `kafkaServiceReady` function. In that function, we can do a conditional to validate if this is a Kafka resource, if not, we return true.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
jelmersnoeck: Seeing that this differs per person, it should actually live in a [global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) and not in this repository.

service_change.go
jelmersnoeck: The connection will already have been made if there is no error and will thus be open, regardless of us ignoring it or not. Therefor, one should still close the body.

A simple `rsp.Body.Close()` to signal that the connection can be closed after the error check will do.

jelmersnoeck:
Thanks for the patch @briancharous ! A few minor changes I'd like to see.

As @FranOis mentioned, this would be ideal to live in a Kafka specific resource as well. But seeing that the common resource should also support this behaviour it can be added in here as well.

service_change.go
briancharous: Yeah, I agree with this. I think we'll make a follow up PR possibly next week to make this change, but I wanted to get this fix in first.

briancharous:
Thanks for the review @jelmersnoeck and @FranOis! I've made a few updates, let me know what you think.

service_change.go
```diff
@@ -30,15 +37,72 @@ func (w *ServiceChangeWaiter) RefreshFunc() resource.StateRefreshFunc {
...
+	}
+
+	if !kafkaRestEnabled && !schemaRegistryEnabled && !kafkaConnectEnabled {
+		return true
+	}
+
+	// Ping each Kafka add-on service's url to see if they are alive
+	if kafkaRestEnabled {
+		if resp, err := http.Get(service.ConnectionInfo.KafkaRestURI); err != nil {
+			resp.Body.Close()
```
FranOis: `resp.Body.Close()` should be called outside of the if statement. Since we need to close it if `err` is `nil`.

You could also create a utility method that encapsulate all this logic and simply takes a URI as param.

I would also get rid of some of the booleans in here. It would look something like this:

```
func kafkaRestReady(service *aiven.Service) bool {
 	if enabled, ok := userConfig["kafka_rest"]; ok && enabled.(bool) {
 	 	return uriReachable(service.ConnectionInfo.KafkaRestURI)
 	}
       return true
}

func uriReachable(uri string) bool {
       resp, err := http.Get(uri)
       defer resp.Body.Close()
       return err == nil
}
```

service_change.go
```diff
@@ -30,15 +37,72 @@ func (w *ServiceChangeWaiter) RefreshFunc() resource.StateRefreshFunc {
 			return nil, "", err
 		}
 
-		return service, service.State, nil
+		// If this is a Kafka service, wait for its component parts (i.e. Kafka Connect,
```
FranOis: Feels redundant now that all this is encapsulated in a method.

briancharous:
@FranOis thanks again, I've made those changes, let me know what you think

briancharous:
@jelmersnoeck Any chance we can get this merged in?

FranOis:
@briancharous sorry for the wait.

PR #21, opened 2018-07-27
Feature ca crt project data source

benjigoldberg:
@jelmersnoeck @FranOis Integrates the project CA source into the provider. Also creates a data source for the project (sadly we can't reuse the provider resource, b/c data sources aren't allowed to have anything other than `Read`)

### DIFF
--- a/resource_project.go
+++ b/resource_project.go
@@ -28,6 +28,42 @@ func resourceProject() *schema.Resource {
 				Required:    true,
 				Description: "Project name",
 			},
+			"ca_crt": {
+				Type:        schema.TypeString,
+				Sensitive:   true,
+				Computed:    true,
+				Description: "Project CA Certificate",
+			},
+		},
+	}
+}
+
+func dataSourceProject() *schema.Resource {
+	return &schema.Resource{
+		Read: resourceProjectRead,
+
+		Schema: map[string]*schema.Schema{
+			"card_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Credit card ID",
+			},
+			"cloud": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Target cloud",
+			},
+			"project": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Project name",
+			},
+			"ca_crt": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Sensitive:   true,
+				Description: "Project CA Certificate",
+			},
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


resource_project.go
```diff
@@ -28,6 +28,42 @@ func resourceProject() *schema.Resource {
...
+		},
+	}
+}
+
+func dataSourceProject() *schema.Resource {
+	return &schema.Resource{
+		Read: resourceProjectRead,
+
+		Schema: map[string]*schema.Schema{
+			"card_id": {
```
FranOis: I don't think card_id is used at all.

resource_project.go
```diff
@@ -28,6 +28,42 @@ func resourceProject() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Credit card ID",
+			},
+			"cloud": {
```
FranOis: Same for cloud, doesn't seem to be used at all.

resource_project.go
```diff
@@ -28,6 +28,42 @@ func resourceProject() *schema.Resource {
 				Required:    true,
 				Description: "Project name",
 			},
+			"ca_crt": {
```
FranOis: why not `ca_certificate`? I hate shortened names when it's not necessary :) But I'm not allergic to them either.

rikonen:
New version that was merged today supports `ca_cert` property for project resource. It does not introduce data source for it but it's not clear to me what would be the benefit of having a data source for this.

benjigoldberg:
@rikonen  this is required if you need to add the certificate your bundle of trusted certificates afaik, otherwise you have to have your tls clients trust that the issuer is valid.

rikonen:
Yes the `ca_cert` itself is definitely required (and as mentioned provided for the resource now). What I meant was that it's not clear why it should be exposed via a datasource instead of accessing it directly using the resource?

benjigoldberg:
@rikonen because for accounts like ours Aiven bills us directly by the project, we don't have the flexibility to use terraform to actually create the projects. If we create a project via terraform we have to reach out to the support team to set up billing on it (direct deposit and all that). So in that scenario, we have to use a data source to pull the information out of a preexisting project. For any existing clients I think this would likely useful as well, right?

rikonen:
@benjigoldberg You can import an existing project into terraform. Just add the project resource to your Terraform script and run import. Something like

```
resource "aiven_project" "myprojectname" {
    project = "myprojectname"
    ... other attributes ...
}
```

And then do

```
terraform import aiven_project.myprojectname myprojectname
```

and you're all set. Also, you can create new projects that inherit billing details from an existing project by specifying `copy_from_project = "myotherproject"` attribute for the new project. That makes the new project use billing if the old one did.

PR #38, opened 2018-12-07
Support service maintenance window and cloud provider VPC peering connection ID

rikonen:


resource_vpc_peering_connection.go
```diff
@@ -117,6 +139,58 @@ func copyVPCPeeringConnectionPropertiesFromAPIResponseToTerraform(
...
+}
+
+// Conf sets up the configuration to refresh.
+func (w *VPCPeeringBuildWaiter) Conf() *resource.StateChangeConf {
+	state := &resource.StateChangeConf{
+		Pending: []string{"APPROVED"},
+		Target: []string{
+			"ACTIVE",
+			"REJECTED_BY_PEER",
+			"PENDING_PEER",
```
mwfrojdman: How does this thing work? I guess in the normal flow
1. user would create peering connection with the Aiven API
2. wait for peering connection to reach state `PENDING_PEER`
3. accept the connection in his AWS account/create symmetric peering in his GCP project
4. wait for peering connection to reach state `ACTIVE`

I suppose this would wait for either state `PENDING_PEER` or `ACTIVE` to be reached, but not be able to distinguish between them, ie. differentiate between states 2 and 4

rikonen: The point of this wait is not to mark the VPC peering connection resource as created until its state has changed from `APPROVED` into something else. This does not relate to the flow beyond that point.

Once the resource has been created some other resource such as the peer connection may be created as the peering connection ID is available for that operation. That will cause the connection state to change to `ACTIVE` but it will not be reflected in the Terraform state automatically but only after user runs refresh. This cannot wait for other states as the operation would block indefinitely and dependent resources could not be created.

rikonen: And I maybe actually answered beside your original question... The `Target` property specifies any state that may be reached in order for the operation to be considered complete. In practice it can hardly ever reach `ACTIVE` here but that's included in the list for the sake of completeness.

PR #41, opened 2019-01-21
Switch from gosimple linter to staticcheck

rikonen:
gosimple has been deprecated and removed. staticcheck is the replacement
linter so start using that and also fix errors reported by it.

### DIFF
--- a/user_config.go
+++ b/user_config.go
@@ -165,13 +165,13 @@ func generateTerraformUserConfigSchema(key string, definition map[string]interfa
 }
 
 func getAivenSchemaType(value interface{}) (string, bool) {
-	switch value.(type) {
+	switch res := value.(type) {
 	case string:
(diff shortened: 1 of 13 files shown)


user_config.go
```diff
@@ -354,16 +354,16 @@ func convertTerraformUserConfigValueToAPICompatibleFormat(
 			omit = true
 		}
 	case "number":
-		switch value.(type) {
+		switch res := value.(type) {
 		case float64:
 		case float32:
```
ojarva: You could use `case float32, int64, int32, int:` here (I think) to avoid repeating `convertedValue = float64(res)` on every separate case. Not sure whether that would be better, though.

rikonen: Doesn't work, it causes `res` to be treated as `interface{}`

PR #65, opened 2019-05-14
Add Kafka 2.2 support

Richard-Mathie:
Kafka 2.2 has been available since April

https://aiven.io/blog/kafka-2-2-now-available-on-aiven/

is this all we need to do to enable it in terraform?


Richard-Mathie:
wondering why the go client dosn't get the schema from the API its self though as that would keep it in sync with what was available...

```bash
$ http get https://api.aiven.io/v1/service_types | jq '.service_types.kafka.user_config_schema'
```
```json
....
"kafka_version": {
      "enum": [
        "2.1",
        "2.2"
      ],
      "title": "Kafka major version",
      "type": [
        "string",
        "null"
      ]
    }
...
```

rikonen:
Automatically getting configuration would be possible for the service user configuration options. Some other options are currently not available via the API. Would need at least simple caching to avoid constantly calling the API.

PR #67, opened 2019-05-14
Add support for rsyslog integrations

briancharous:


rikonen:
Looks good, thanks. We'll still try this out before merging.

briancharous:
Thanks @rikonen, I've tried this out and it works for our configuration, but I haven't tried using it with client certificate auth.

Issue #56, opened 2019-04-17
Support for crossregion VPC peering


Ormod:
Add support for crossregion VPC peering. Currently the provider only support single-region VPC peering.

mteiste:
Available in version 1.0.10

PR #70, opened 2019-06-05
add cross region vpc peering

mteiste:
fix issue #56

### DIFF
--- a/resource_vpc_peering_connection.go
+++ b/resource_vpc_peering_connection.go
@@ -41,6 +41,15 @@ func resourceVPCPeeringConnection() *schema.Resource {
 				Required:    true,
 				Type:        schema.TypeString,
 			},
+			"peer_region": {
+				Description: "AWS region of the peered VPC (if not in the same region as Aiven VPC)",
+				ForceNew:    true,
+				Optional:    true,
+				Type:        schema.TypeString,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return new == ""
+				},
+			},
 			"state": {
 				Computed:    true,
 				Description: "State of the peering connection",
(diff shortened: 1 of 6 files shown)


resource_vpc_peering_connection.go
```diff
@@ -41,6 +41,18 @@ func resourceVPCPeeringConnection() *schema.Resource {
 				Required:    true,
 				Type:        schema.TypeString,
 			},
+			"peer_region": {
+				Description: "AWS region of the peered VPC (if not in the same region)",
```
rikonen: If not in the same region as what.

resource_vpc_peering_connection.go
```diff
@@ -41,6 +41,18 @@ func resourceVPCPeeringConnection() *schema.Resource {
...
+				ForceNew:    true,
+				Optional:    true,
+				Type:        schema.TypeString,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					if new != "" {
+						return false
+					}
+					return true
```
rikonen: `return new == ""` would work as well I suppose

resource_vpc_peering_connection.go
```diff
@@ -58,15 +70,34 @@ func resourceVPCPeeringConnection() *schema.Resource {
...
-			PeerVPC:          d.Get("peer_vpc").(string),
-		},
+	peerRegion := d.Get("peer_region").(string)
+
+	var (
+		pc  *aiven.VPCPeeringConnection
+		err error
 	)
 
+	if peerRegion != "" {
```
rikonen: Only having single `Create` in the API client and something like

```
apiPeerRegion := nil
if peerRegion != "" {
    apiPeerRegion = &peerRegion
}
pc, err := client.VPCPeeringConnections.Create(...)
```

would seem more convenient.

resource_vpc_peering_connection.go
mteiste: fixed

resource_vpc_peering_connection.go
mteiste: fixed

mteiste:
manually tested
- created vpc and one peering in the same region and one peering in an other region
- applied and verified from Aiven console that the vpc and peerings were created
- refreshed
- tested plan (to test that the same region peering will stay unchanged even though the refresh will fill the peer_region
- removed peerings from the configuration
- applied and verified from Aiven console that the peerings were deleted
- removed vpc from configuratoin
- applied and verified from Aiven console that the vpc status was changed to deleted

Issue #85, opened 2019-08-06
Terraform plan reveals sensitive data for Kafka service


milpog:
I am using `1.0.11` version of `aiven` provider and when I am running plan for `Kafka` service, sensitive data are revealed, as show on example below.

```hcl
resource "aiven_service" "kafka" {
        cloud_name              = "aws-eu-central-1"
        id                      = "project-name/service-name"
        maintenance_window_dow  = "monday"
        maintenance_window_time = "02:00:00"
        plan                    = "startup-2"
        ...
        service_type            = "kafka"
        service_uri             = (sensitive value)
        state                   = "RUNNING"
        termination_protection  = true

        kafka {
            access_cert         = (sensitive value)
            access_key          = (sensitive value)
            connect_uri         = "https://avnadmin:password@service-name-project-name.aivencloud.com:14000"
            rest_uri            = "https://avnadmin:password@service-name-project-name.aivencloud.com:14001"
            schema_registry_uri = "https://avnadmin:password@service-name-project-name.aivencloud.com:14002"
        }
       ...
}
```

Expected behaviour would be:

```hcl
resource "aiven_service" "kafka" {
        cloud_name              = "aws-eu-central-1"
        id                      = "project-name/service-name"
        maintenance_window_dow  = "monday"
        maintenance_window_time = "02:00:00"
        plan                    = "startup-2"
        ...
        service_type            = "kafka"
        service_uri             = (sensitive value)
        state                   = "RUNNING"
        termination_protection  = true

        kafka {
            access_cert         = (sensitive value)
            access_key          = (sensitive value)
            connect_uri         = (sensitive value)
            rest_uri            = (sensitive value)
            schema_registry_uri = (sensitive value)
        }
       ...
}
```

milpog:
This is similar to https://github.com/aiven/terraform-provider-aiven/issues/77

milpog:
Tested #86 manually and it works as expected.

PR #86, opened 2019-08-06
Mark Kafka uri attributes as sensitive

milpog:
Currently, the definition of some `uri` attribute on Kafka `aiven_service` resource doesn't state it as sensitive. `uri` attribute usually contains the user+password combination of the default/admin user to connect to the service.

Fixes https://github.com/aiven/terraform-provider-aiven/issues/85.


rikonen:
Thank you for the contribution!

PR #101, opened 2019-09-17
Add topic and acl caching

eversC:
Cc @mteiste 

Apologies it took me so long to get this PR in..

For anyone checking this, the TL;DR is that we have a large number (thousands) of resources in Aiven that we've Terraformed in a single GitHub repo. Running `tf plan` and `tf apply` commands are taking an extremely long time for Topics and ACLs due to an API call being made for each.

This change introduces an ACL and a Topic cache, which are populated using a single List API call for each. It's cut our tf plan/applies down to sub-2mins, and I hope it could reduce the impact on your API too?

We also added a sleep into each resource, as without them we were seeing most plan/applies failing with a gRPC error: `Error: rpc error: code = Canceled desc = context canceled`. Unfortunately we don't have much gRPC expertise currently, so we've not been able to track down the route cause yet, but adding sleeps seemed to reduce the frequency of the failures.

### DIFF
--- a/aiven/kafka_topic_change.go
+++ b/aiven/kafka_topic_change.go
@@ -5,6 +5,8 @@ import (
 	"time"
 
 	"github.com/aiven/aiven-go-client"
+	"github.com/aiven/terraform-provider-aiven/pkg/cache"
+
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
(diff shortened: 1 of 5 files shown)


eversC:
👋 has there been any movement with this?

we added a lock around the underlying maps (and removed the temporary sleeps) in the topic and acl caches, in order to fix the following issue I reported in the OP:

> We also added a sleep into each resource, as without them we were seeing most plan/applies failing with a gRPC error: Error: rpc error: code = Canceled desc = context canceled.

If there's anything we can do here at OVO then let me know. We're really keen to get this merged upstream so we no longer have to build your provider from our own repo.

PR #102, opened 2019-10-15
add APPROVED as valid vpc_peering connection status

mteiste:


mteiste:
APPROVED is transitional state, should not actually be needed.

PR #107, opened 2019-11-21
Enable topic caching

eversC:
This supplements #101 (caching was only enabled for ACLs, it should have been enabled for Topics as well).


mteiste:
looks ok, could you combine the commits into one?

eversC:
Messed up my git history in an attempt to squash commits, raised a new PR instead: #111

PR #109, opened 2019-11-28
schemas: update config schemas to match latest API

mteiste:
This also enables creation of Elasticsearch version 7, Kafka version 2.4
and Postgresql version 12 services.

### DIFF
--- a/aiven/templates/service_user_config_schema.json
+++ b/aiven/templates/service_user_config_schema.json
@@ -3,7 +3,9 @@
     "additionalProperties": false,
     "properties": {
       "ip_filter": {
-        "default": ["0.0.0.0/0"],
+        "default": [
+          "0.0.0.0/0"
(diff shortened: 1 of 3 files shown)


aiven/templates/service_user_config_schema.json
```diff
@@ -1278,24 +1598,34 @@
...
+          "9.5",
+          "9.6",
+          "10",
+          "11",
+          "12"
+        ],
         "title": "PostgreSQL major version",
-        "type": ["string", "null"]
-      },
-      "variant": {
```
rikonen: This shouldn't be dropped. See internal pull request related to this.

mteiste: It is not dropped, the automatic schema generation script just moved it to an other location (line 1723)

PR #113, opened 2019-12-05
Add support of the Aiven terraform plugin on Windows

ivan-savciuc:
Binaries for 64-bit and 32-bit versions of Windows will be generated during the `make release` and manually added to the next release of Aiven terraform provider.


.gitignore
```diff
@@ -6,3 +6,4 @@ vendor/*
 aiven/aiven-packr.go
 packrd/*
 plugins/
+.idea/
```
mteiste: I would recommend adding developer's ide (or enviroment specific) ignores to your global .gitignore

ivan-savciuc: good idea, thanks

PR #134, opened 2020-01-20
fix <service>_user_config Boolean fields without default value

ivan-savciuc:
For all the boolean configuration options without a default value use type string instead of boolean for internal representation. We will be using string instead of boolean to track whether a user is setting any values for such configuration options or not since Terraform boolean type cannot provide such indication. 

This solution is based on the fact that Terraform seamlessly for a user converts any simple type value to a string, for example:

`action_destructive_requires_name = true` if resource filed `action_destructive_requires_name` has a type `string` true value will be converted to `"true"` string value.

### DIFF
--- a/aiven/user_config.go
+++ b/aiven/user_config.go
@@ -4,6 +4,7 @@ package aiven
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/gobuffalo/packr/v2"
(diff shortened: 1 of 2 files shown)


aiven/user_config.go
```diff
@@ -381,6 +398,20 @@ func convertTerraformUserConfigValueToAPICompatibleFormat(
...
 		}
 	case "boolean":
+		switch value.(type) {
+		case string:
+			if value == "<<value not set>>" {
+				omit = true
+			} else {
+				b, err := strconv.ParseBool(value.(string))
+				if err != nil {
+					panic(fmt.Sprintf("Invalid %v user config key type %T for %v, expected string", serviceType, value, key))
```
rikonen: Error message looks incorrect. The type is string but value is not something that can be parsed as bool.

aiven/user_config.go
```diff
@@ -381,6 +398,20 @@ func convertTerraformUserConfigValueToAPICompatibleFormat(
...
+				}
+				convertedValue = b
+			}
+		default:
```
rikonen: Should this branch rather be executed if type is bool and default action should be raising an error about incorrect type?

ivan-savciuc: sounds like a good idea

PR #140, opened 2020-01-23
test-acc: project

ivan-savciuc:
use `make testacc` to run the acceptance tests

`AIVEN_TOKEN` environment variable is required and should contain a valid Aiven API token

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,11 @@ test: vendor
 	CGO_ENABLED=0 go test -v ./...
 
 testacc: vendor
-	TF_ACC=1 CGO_ENABLED=0 go test -v --cover ./...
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -54,6 +54,16 @@ release: bins
...
 
+testacc: vendor
+	TF_ACC=1 CGO_ENABLED=0 go test -v --cover ./... -timeout 120m
+
+sweep:
+	@echo "WARNING: This will destroy infrastructure. Use only in development accounts."
+	go test -v ./... -sweep=global -timeout 60m
+
+fmt: vendor
+	! go fmt ./... 2>&1 | read
```
rikonen: Is putting this target back intentional? It was just removed in an earlier commit

ivan-savciuc: a good catch, it shouldn't be there

PR #151, opened 2020-01-30
testacc: kafka topic

ivan-savciuc:


aiven/resource_service_user_test.go
```diff
@@ -75,6 +75,10 @@ func testAccServiceUserResource(name string) string {
 		  service_type = "pg"
 		  maintenance_window_dow = "monday"
 		  maintenance_window_time = "10:00:00"
+		  
+          pg_user_config {
```
rikonen: Not that it really matters but this block is mixing tabs and spaces and use of those isn't quite consistent.

ivan-savciuc: update using tabs for consistency

PR #153, opened 2020-01-31
testacc: kafka acl

ivan-savciuc:


aiven/datasource_kafka_acl.go
```diff
@@ -31,7 +31,7 @@ func datasourceKafkaACLRead(d *schema.ResourceData, m interface{}) error {
 	for _, acl := range acls {
 		if acl.Topic == topic && acl.Username == userName {
 			d.SetId(buildResourceID(projectName, serviceName, acl.ID))
-			return copyKafkaACLPropertiesFromAPIResponseToTerraform(d, acl, projectName, serviceName)
+			return resourceKafkaACLRead(d, m)
```
rikonen: What's the point of this change? `resourceKafkaACLRead` ends up calling `KafkaACLHandler.Get`, which in turn calls `KafkaACLHandler.List` and when gets the response it in turn calls `copyKafkaACLPropertiesFromAPIResponseToTerraform` so as far as I can tell this adds a lot of extra processing to get the same end result.

ivan-savciuc: Reasons I had in mind:
- unify reading methods, in case we will change something in the future in `resourceKafkaACLRead` we would not need to copy&paste similar code to a data source
- we are using `cache.ACLCache{}.Read` to read stuff from API inside resourceKafkaACLRead, consequently not necessary we will end-up calling `KafkaACLHandler.List ` especially for the data source
- it is standard practice to call `READ` at the end of all modifications, or data source read to synchronize immediately and avoid any diff. Let us go even further and replace all the calls of `copyKafkaACLPropertiesFromAPIResponseToTerraform` inside Update. Delete and other methods to be compliant with Terraform best practices before we apply for https://www.terraform.io/guides/terraform-provider-development-program.html 

Does it sound reasonable?

rikonen: OK, seems fair

PR #162, opened 2020-02-07
Fix user config optional bool

ivan-savciuc:


aiven/user_config.go
```diff
@@ -302,7 +302,10 @@ func convertAPIUserConfigToTerraformCompatibleFormat(
 			case string:
 				terraformConfig[key] = apiValue
 			case bool:
-				terraformConfig[key] = strconv.FormatBool(apiValue.(bool))
+				_, ok := schemaDefinition["api_type"]
+				if ok {
```
rikonen: I'm not sure I follow this logic. If api_type value is not present then we just silently don't set any value here?

ivan-savciuc: sry, I did partial push, added WIP to this PR

PR #163, opened 2020-02-10
Add accounts support

ivan-savciuc:
In this PR, added support for:
- accounts
- account teams
- account team members and invitations
- account projects 

Account Authentication methods support will be in a separate PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ the services.
 resource "aiven_project" "myproject" {
     project = "<PROJECT_NAME>"
     card_id = "<FULL_CARD_ID/LAST4_DIGITS>"
+    account_id = aiven_account_team.<ACCOUNT_RESOURCE>.account_id
 }
 ```
 
@@ -113,6 +114,9 @@ UUID is not shown in the UI it is typically easier to use the last 4 digits to i
 the card. This can be omitted if `copy_from_project` is used to copy billing info from
 another project.
 
+`account_id` is an optional property to link a project to already an existing account by 
+using account ID
+
 `copy_from_project` is the name of another project used to copy billing information and
 some other project attributes like technical contacts from. This is mostly relevant when
 an existing project has billing type set to invoice and that needs to be copied over to a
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -113,6 +114,9 @@ UUID is not shown in the UI it is typically easier to use the last 4 digits to i
 the card. This can be omitted if `copy_from_project` is used to copy billing info from
 another project.
 
+`account_id` is an optional property to lick a project to already an existing account by
```
skx: "lick" is probably a mistake.

README.md
```diff
@@ -125,6 +129,62 @@ Kafka. This value cannot be set, only read.
 
 Aiven ID format when importing existing resource: name of the project as is.
 
+### Resource Account
+```
+resource "aiven_account" "foo" {
```
rikonen: myaccount or somesuch might more readable sample value than foo. Sample for other similar cases.

README.md
```diff
@@ -125,6 +129,62 @@ Kafka. This value cannot be set, only read.
...
+`account_id` is an auto-generated unique account id.
+
+### Resource Account Team
+```
+resource "aiven_account_team" "foo" {
+    account_id = aiven_account.<ACCOUNT_RESOURCE>.account_id
+    name = "account_team1"
+    
+    # invite users to a team
+    invite_users = "user+1@example.com,user+2@example.com"
```
rikonen: This differs from how invitations work for projects. Unless there's some technical reason why team invites must work differently I'd use the same model for both.

This list of invite users seems potentially problematic. What happens when user actually accepts the invite? Then the user isn't invited anymore so Terraform state does not match actual state. Does removing an invite mean that you're removing the team member if the members has already accepted the invite? Having invites handled via team member resource that just has computed accepted property seems better.

aiven/resource_account.go
```diff
@@ -0,0 +1,161 @@
...
+		Optional:    true,
+		Computed:    true,
+	},
+	"tenant_id": {
+		Type:        schema.TypeString,
+		Description: "Tenant id",
+		Optional:    true,
+		Computed:    true,
+	},
+	"billing_enabled": {
```
rikonen: I don't think this property exists.

README.md
ivan-savciuc: Yes, indeed, it is different from project invitations and might not be straightforward for a user. 

First of all, I missed that we have a `Delete` method for invitations in Aiven API
```
"/account/<account_id>/team/<team_id>/invites/<user_email>",
method="DELETE"
```

And now it seems that the best idea would be to create a separate TF resource for Account Tema Invitations `account_team_invitation`, and when a user accepts an invitation, we will also check the member's list to make sure invitation was accepted. Perhaps even add a computed field `status` (sent, accepted) which will be calculated based on two records: invitations and members list. 

And `aiven_account_team.member` field will stay as the computed list of members. 

Or do you have something else in mind?

rikonen: Should be `account_team_member` instead of `account_team_invitation`. The `aiven_account_team.member` that I would consider not adding initially and if there's demand add it later. We don't have that for project and I don't think anyone has ever asked for it.

ivan-savciuc: Added `aiven_account_team_member` resource and deleted `aiven_account_team.member` and `aiven_account_team.invite`.

README.md
```diff
@@ -125,6 +129,77 @@ Kafka. This value cannot be set, only read.
...
+`account_id` is an auto-generated unique account id.
+
+### Resource Account Team
+```
+resource "aiven_account_team" "account_team1" {
+    account_id = aiven_account.<ACCOUNT_RESOURCE>.account_id
+    name = "account_team1"
+    
+    # invite users to a team
+    invite_users = "user+1@example.com,user+2@example.com"
```
rikonen: This invite_users hasn't been completely removed yet, I can see at least 5 references.

README.md
```diff
@@ -125,6 +129,77 @@ Kafka. This value cannot be set, only read.
...
+`account_id` is an unique account id.
+
+`invite_users` a list of comma-separated emails of the users that need to be invited 
+to an account team.
+
+`team_id` is an auto-generated unique account team id.
+
+`invite` a list of email invitations sent to the users defined in `invite_users` property.
+
+`member` a list of account team members.
```
rikonen: member and invite above should be removed

aiven/resource_account_team_member.go
```diff
@@ -0,0 +1,223 @@
...
+		Description: "Team invited by user email",
+		Optional:    true,
+		Computed:    true,
+	},
+	"create_time": {
+		Type:        schema.TypeString,
+		Description: "Time of creation",
+		Optional:    true,
+		Computed:    true,
+	},
```
rikonen: I would add `accepted` property like we have for project user resource for consistency.

rikonen: This conversation was marked as resolved but I don't see the value

ivan-savciuc:
added `aiven_account_team_member.accepted` Boolean property

PR #171, opened 2020-02-17
add account authentication support

ivan-savciuc:


aiven/resource_account_authentication.go
```diff
@@ -0,0 +1,225 @@
...
+	"account_id": {
+		Type:        schema.TypeString,
+		Description: "Account id",
+		Required:    true,
+	},
+	"type": {
+		Type:         schema.TypeString,
+		Description:  "Account authentication id",
+		Required:     true,
+		ValidateFunc: validation.StringInSlice([]string{"internal", "saml"}, false),
```
ojarva: What happens when we add the next authentication type? Does this fail when response (authentication methods list) contains entries with something like `openid`? We are not currently planning on adding OpenID support, but I'm fairly certain we'll be adding new authentication method types, and old official clients and libraries shouldn't break when that happens (if user creates a new method directly through aiven console).

ivan-savciuc: When we add new authentication type, we have first to update go-client and add all corresponding fields, and then TF provider. I believe if we add a new authentication type on API side, it won't work out of the box in Terraform, and it will require new code to be added in both go-client and TF provider and tested.

ivan-savciuc:
Can you please have a look at this PR? There is no activity for almost two weeks.

rikonen:
The console support for this is still not finalized afaik so the comment from Olli that this shouldn't be merged stands.

mteiste:
rebase (and convert to draft if this is still pending on changes on Aiven side)

ivan-savciuc:
Last accounts-related changes already in the master on the Aiven side, please review this PR again

Issue #170, opened 2020-02-17
Cannot create a Postgres read replica


nrf110:
I'm receiving the following error when creating a Postgres read-replica: 
`error waiting for Aiven service to be RUNNING: timeout while waiting for state to become 'RUNNING' (last state: 'WAITING_FOR_SERVICES', timeout: 20m0s)`

The service is created correctly, and is correctly configured to be a read replica of our primary database.  The ServiceWaiter is checking for the existence of backups:  https://github.com/aiven/terraform-provider-aiven/blob/master/aiven/service_change.go#L62.

A backup does not seem to be automatically created at service creation time like it does for a Postgres master, therefore it keeps polling until the timeout, despite the service actually being in a state of RUNNING.

Should the Aiven API be taking a backup of a read replica, or should there be additional logic in the terraform provider to skip the check for backups if this is a read replica?

rikonen:
Read-replicas don't have backups. The backup wait should just be removed in case the service is a read replica. @ivan-savciuc fyi, same affects MySQL probably

ivan-savciuc:
This issue should be fixed in V1.2.2 release https://github.com/aiven/terraform-provider-aiven/releases/tag/v1.2.0

PR #172, opened 2020-02-18
fix a creation problem for read-only replica type DB services

ivan-savciuc:
Should fix this issue https://github.com/aiven/terraform-provider-aiven/issues/170

### DIFF
--- a/aiven/service_change.go
+++ b/aiven/service_change.go
@@ -59,19 +59,24 @@ func backupsReady(service *aiven.Service) bool {
 		return true
 	}
 
+	// no backups for read replicas type of service
+	for _, i := range service.Integrations {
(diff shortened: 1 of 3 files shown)


aiven/service_change.go
```diff
@@ -59,19 +59,24 @@ func backupsReady(service *aiven.Service) bool {
 		return true
 	}
 
+	// no backups for read replicas type of service
+	for _, i := range service.Integrations {
+		if i.IntegrationType == "read_replica" {
```
rikonen: I think this needs to check the direction of the integration. The read_replica integration exists for the source service too. The service is only read-replica if destination service id matches the services own id.

ivan-savciuc: thanks, I will change that

Issue #154, opened 2020-02-02
Deletion protection


steshaw:
I like the idea of the `termination_protection` for services, but since this cannot prevent disasters for production services, I was wondering if there's a way to enable deletion protection for resources such as databases, topics, etc.

ivan-savciuc:
Hi,

yes, it is possible to do so by utilizing Terraform `prevent_destroy` - This flag provides extra protection against the destruction of a given resource. When this is set to true, any plan that includes a destroy of this resource will return an error message.

Example: 
```
resource "aiven_service" "kafka-service1" {
  project = aiven_project.project1.project
  cloud_name = "google-europe-west1"
  plan = "business-4"
  service_name = "kafka1"
  service_type = "kafka"

  kafka_user_config {
    kafka_version = "2.4"
    kafka {
      group_max_session_timeout_ms = 70000
      log_retention_bytes = 1000000000
    }
  }

  lifecycle {
    prevent_destroy = true
  }
}

```

Ivan

steshaw:
Hi Ivan, Terraform has a problem that `prevent_destroy` does not warn of the impending deletion of resources in all cases.

https://github.com/hashicorp/terraform/issues/17599

We host one of our most crucial databases on Aiven. I want a to find a way to configure it so that Terraform cannot destroy it for any reason — i.e. that destruction, if required, needs manual intervention. As context, our organisation has had disasters caused by Terraform's deletion and recreation of resources in the past and so are rightfully wary of taking up Terraform in the present. I want to put safety barriers in place so that we can safely adopt it once more.

ivan-savciuc:
Hi,

Unfortunately, I cannot offer anything better than using `termination_protection` and `prevent_destroy=true` right now, and it looks like you are already using both of them. And we do not have anything like that on the pipeline, and I will create a ticket in our internal issue tracking system. And such a feature should be implemented on the Aiven API side and will be exposed through our Terraform provider when available. 

Ivan

ivan-savciuc:
Well, I was wrong and did some experiments, it is possible to prevent the deletion of a resource. We will create a separate resource field, and if it is set to the true resource would be impossible to delete via terraform but of course, can be removed via UI. Or perhaps we can enable this logic based on `termination_protection` value. I will come back to you soon with a solution.

steshaw:
@ivan-savciuc That's really great news. Thank you!

ivan-savciuc:
[...]

PR #177, opened 2020-02-20
testacc: service deletion protection

ivan-savciuc:


aiven/resource_service.go
```diff
@@ -574,6 +575,16 @@ func resourceServiceDelete(d *schema.ResourceData, m interface{}) error {
 	client := m.(*aiven.Client)
 
 	projectName, serviceName := splitResourceID2(d.Id())
+
+	service, err := client.Services.Get(projectName, serviceName)
+	if err != nil {
+		return err
+	}
+
+	if service.TerminationProtection == true {
```
rikonen: What's the reason for doing this check on client side? I don't know if there's any harm in doing that really but since the server should anyway reject the deletion this seems a bit unnecessary and was just wondering why it's been added.

ivan-savciuc: It was requested here https://github.com/aiven/terraform-provider-aiven/issues/154 to have such protection, and from the name of the field itself, it wasn't clear that it will prevent from deletion of a service. After doing writing acceptance tests for this feature, I understood that `termination_protection` does already what the customer wants, and there are not that many good reasons to have this in TF provider code.

The few I can think of are:
- test during the run of `make testacc` weather `termination_protection` works on API side and block deletion. 
- if by any reason API `termination_protection` deletion protection fails we have a second line of defence in TF provider.

I still decided to make this PR since code was ready, knowing that it might not get merged in a current form of at all. Therefore if you do not see any benefits of having this in TF provider feel free to decline this PR. Or maybe we can just leave ACC tests to validation this functionality working on API side. 

Let me know WDYT?

rikonen: I think having the test makes sense but I'd drop this check from the provider and just have the test ensure that the server actually does prevent deleting the service.

ivan-savciuc: only ACC tests left which validate API deletion protection with `termination_protection` set true

PR #178, opened 2020-02-21
testacc and termination protection

ivan-savciuc:
Spent a lot of time searching for a way to test `termination_protection` accurately, but TF testing framework does not provide a way to interrupt or replace delete plan cycle. Therefore only test plan for use case when `termination_protection` is enabled.

### DIFF
--- a/aiven/resource_database.go
+++ b/aiven/resource_database.go
@@ -56,10 +56,12 @@ var aivenDatabaseSchema = map[string]*schema.Schema{
 		DiffSuppressFunc: handleLcDefaults,
 	},
 	"termination_protection": {
-		Type:        schema.TypeBool,
-		Optional:    true,
-		Default:     false,
-		Description: "Prevent database from being deleted. It is recommended to have this enabled for all services.",
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+		Description: `It is a Terraform client-side deletion protections, which prevents the database
+			from being deleted by Terraform. It is recommended to enable this for any production
+			databases containing critical data.`,
 	},
 }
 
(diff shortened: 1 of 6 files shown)


aiven/resource_database.go
```diff
@@ -55,13 +55,20 @@ var aivenDatabaseSchema = map[string]*schema.Schema{
 		ForceNew:         true,
 		DiffSuppressFunc: handleLcDefaults,
 	},
+	"termination_protection": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "Prevent database from being deleted. It is recommended to have this enabled for all services.",
```
rikonen: Would be good to mention this is client-side only functionality so the database can be deleted from web console or by direct API call. Also `It is recommended to have this enabled for all services.` is a bit vague recommendation. If that was the case we should just default to true. Maybe something like `It is recommended to enable for this any production databases containing critical data.` instead.

aiven/resource_kafka_topic.go
```diff
@@ -65,6 +65,12 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 		Description: "Topic cleanup policy. Allowed values: delete, compact",
 		ForceNew:    true,
 	},
+	"termination_protection": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "Prevent kafka topic from being deleted. It is recommended to have this enabled for all services.",
```
rikonen: Same comment here as for database.

rikonen:
The descriptions are still lacking the bit of mentioning this is client side functionality only. Current description may lead to someone mistakenly thinking the databases / topics cannot be deleted from console or via API either.

ivan-savciuc:
Docs were updated clearly stating TF client-side nature of `termination_protection` fields.

PR #184, opened 2020-03-04
project vpc graceful deletion

ivan-savciuc:
project vpc graceful deletion and acceptance tests

### DIFF
--- a/aiven/resource_project_vpc.go
+++ b/aiven/resource_project_vpc.go
@@ -211,8 +211,8 @@ func (w *ProjectVPCDeleteWaiter) RefreshFunc() resource.StateRefreshFunc {
 		if vpc.State != "DELETING" && vpc.State != "DELETED" {
 			err := w.Client.VPCs.Delete(w.Project, w.VPCID)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


aiven/resource_project_vpc.go
```diff
@@ -154,13 +178,61 @@ func (w *ProjectVPCActiveWaiter) RefreshFunc() resource.StateRefreshFunc {
...
+			}
+
+			return nil, "", err
+		}
+
+		if vpc.State != "DELETING" && vpc.State != "DELETED" {
+			err := w.Client.VPCs.Delete(w.Project, w.VPCID)
+			if err != nil {
+				// VPC cannot be deleted while there are services migrating from it
+				// and coses 409 error API response
```
rikonen: Typo in `coses` but also the typical reason for failing to delete VPC is that services in it were deleted very recently and the cloud resources related to the VMs themselves have not yet been deleted, which prevents removing the VPC. Maybe "VPC cannot be deleted while there are services migrating from it" => "VPC cannot be deleted while there are services migrating from it or service deletion is still in progress".

PR #187, opened 2020-03-05
testacc: service integration

ivan-savciuc:
Acceptance tests for service integration and service integration endpoint TF resources

### DIFF
--- a/aiven/resource_service_integration_endpoint_test.go
+++ b/aiven/resource_service_integration_endpoint_test.go
@@ -11,11 +11,10 @@ import (
 )
 
 func TestAccAivenServiceIntegrationEndpoint_basic(t *testing.T) {
-	t.Parallel()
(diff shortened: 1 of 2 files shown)


aiven/resource_service_integration_endpoint_test.go
```diff
@@ -0,0 +1,133 @@
...
+				}
+			}
+		}
+		
+		resource "aiven_service_integration_endpoint" "bar" {
+			project = aiven_project.foo.project
+			endpoint_name = "test-acc-ie-%s"
+			endpoint_type = "datadog"
+			datadog_user_config {
+				datadog_api_key = "Jwx4dl20zOfyYsJGvuv2fiV2VZzCgsuK"
```
rikonen: I guess this is just some bogus string. Perhaps some other integration type like Prometheus would be better in long term since this will cause us to make some failing requests to their API.

PR #189, opened 2020-03-09
doc: add VPC peering connection states documentation

ivan-savciuc:


rikonen:
Didn't read the actual text in much detail (@mwfrojdman can you please have a look?) but as a general note something like "and the customer needs to perform a cloud provider-specific steps in their cloud account to activate it" sounds weird because this documentation is intended for customers. Maybe "and you need to perform cloud provider specific steps in your cloud account to active it".

README.md
```diff
@@ -747,8 +747,47 @@ peered with.
...
+
+- `APPROVED` is the initial state after the user does a successful creation of a 
+peering connection resource via Terraform. 
+
+- `PENDING_PEER` appears when Aiven system has created one or more cloud resources from 
+the Aiven cloud account to the user's VPC/network that was specified by the peering 
+connection. This state means a VPC peering connection resource was successfully created 
+and in the pending state, and the user needs to perform a cloud provider-specific steps 
+in their cloud account to activate it. The state can change to `ACTIVE`, `DELETED`, 
+`DELETED_BY_PEER` and `REJECTED_BY_PEER`.
```
mwfrojdman: This could say that "The connection enters the `PENDING_PEER` state from `APPROVED` once the Aiven platform has created a connection to the specified peer successfully in the cloud, but the connection is not active until the user completes the setup in their cloud account. The steps needed in the user cloud account depend on the used cloud provider"

README.md
```diff
@@ -747,8 +747,47 @@ peered with.
...
+
+- `ACTIVE` stands for successful VPC peering connection that has been established, and all 
+related cloud resource is in a state that allows traffic to flow to the peered network, 
+a successful VPC peering connection. VPC peering connection stays in this state in until it 
+gets deleted. State transitions are to `DELETED` and `DELETED_BY_PEER`.
```
mwfrojdman: Perhaps "is deleted" instead of "gets deleted"

README.md
```diff
@@ -747,8 +747,47 @@ peered with.
...
+- `PENDING_PEER` appears when Aiven system has created one or more cloud resources from 
+the Aiven cloud account to the user's VPC/network that was specified by the peering 
+connection. This state means a VPC peering connection resource was successfully created 
+and in the pending state, and the user needs to perform a cloud provider-specific steps 
+in their cloud account to activate it. The state can change to `ACTIVE`, `DELETED`, 
+`DELETED_BY_PEER` and `REJECTED_BY_PEER`.
+
+- `ACTIVE` stands for successful VPC peering connection that has been established, and all 
+related cloud resource is in a state that allows traffic to flow to the peered network, 
+a successful VPC peering connection. VPC peering connection stays in this state in until it
```
mwfrojdman: Maybe something like "`ACTIVE` stands for a VPC peering connection whose setup has been completed`. We should document the rest of the implications elsewhere, because the details may change over time and do not depend on just the Terraform provider.

README.md
```diff
@@ -747,8 +747,47 @@ peered with.
...
+
+- `DELETED_BY_PEER` appears when a user deleted the VPC peering connection through their cloud 
+account. That is, the user deleted the peering cloud resource in their account. There are no 
+transitions from this state
+
+- `REJECTED_BY_PEER` an AWS specific state, when VPC peering connection was in the `PENDING_PEER` state, 
+and the user rejected the AWS peering connection request.
+
+- `INVALID_SPECIFICATION` a VPC peering connection was in state APPROVED state. Still, something 
+the user controls broke a successful connection, for example, the peer cloud account of VPC
```
mwfrojdman: Perhaps something more along the lines of "`INVALID_SPECIFICATION` is a VPC peering connection that was in the `APPROVED` state but could not be successfully created because of something in the user's control" and then the examples

README.md
```diff
@@ -747,8 +747,42 @@ peered with.
...
+
+- `APPROVED` is the initial state after the user does a successful creation of a 
+peering connection resource via Terraform. 
+
+- `PENDING_PEER` the connection enters the `PENDING_PEER` state from `APPROVED` once the 
+Aiven platform has created a connection to the specified peer successfully in the cloud, 
+but the connection is not active until the user completes the setup in their cloud account. 
+The steps needed in the user cloud account depend on the used cloud provider.
+
+- `ACTIVE` ACTIVE stands for a VPC peering connection whose setup has been completed
```
mwfrojdman: Is the second `ACTIVE` extra?

README.md
```diff
@@ -747,8 +747,42 @@ peered with.
...
+account. That is, the user deleted the peering cloud resource in their account. There are no 
+transitions from this state
+
+- `REJECTED_BY_PEER` an AWS specific state, when VPC peering connection was in the `PENDING_PEER` state, 
+and the user rejected the AWS peering connection request.
+
+- `INVALID_SPECIFICATION` is a VPC peering connection that was in the `APPROVED` state but could not be  
+successfully created because of something in the user's control, for example, the peer cloud account of VPC 
+doesn't exist, overlapping IP ranges, or the Aiven cloud account doesn't have permissions to peer 
+there. Usually, `state_info` field contains more details about the particular issue.
```
mwfrojdman: This should be without `Usually, `, the `state_info` always has more details (if nothing else then the `type` and `message` properties in that JSON object are always set

PR #190, opened 2020-03-10
preparation for 1.2.2 release, change log updates

ivan-savciuc:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [1.2.2] - 2020-03-10
+- Grafana service waits until Grafana is reachable publicly (in case `public_access.grafana`
+ is set to true) during resource creation or update.
```
rikonen: The clarification is a bit misleading, though I don't know if it's easy to write it in very clear manner. The service is accessible "if (it is not in a VPC or public_access.grafana is true) and there's no IP filter that prevents accessing it". Come to think of it, the current logic does not properly handle the IP filter case. Presumably if IP filter is anything but 0.0.0.0/0 we should not try connecting to Grafana as it's likely to fail.

ivan-savciuc: Updated 1.2.2 version description, also IP filter related business logic was added https://github.com/aiven/terraform-provider-aiven/pull/191

rikonen:
The changelog text doesn't look any different than it did before.

ivan-savciuc:
Sorry, had an Internet connection issue: 
```
ssh: Could not resolve hostname github.com: Name or service not known
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Please check now; changes should be available.

PR #198, opened 2020-04-07
remove CNAME configuration option from Kafka resource

ivan-savciuc:


ivan-savciuc:
Why do we still have it in the Aiven API `user_config_schema` in the first place?

mteiste:
Closing as the fix actually needs to be on the API side, as a workaround for now I'll modify the updater to support some kind of blacklisting as a workaround for now.

Issue #204, opened 2020-04-14
Custom Timeouts for Resources


encima:
Implement custom timeouts for all Aiven resources

> As an Aiven terraform user I would like the "aiven_vpc_peering_connection" resource (as well as all resources) to support custom timeouts so that I can change the default timeout when it is set too short.

ivan-savciuc:
Available in https://github.com/aiven/terraform-provider-aiven/releases/tag/v1.2.4

PR #210, opened 2020-04-17
add vpc custom timeout properties

ivan-savciuc:
As requested here https://github.com/aiven/terraform-provider-aiven/issues/204


aiven/resource_project_vpc.go
```diff
@@ -35,6 +36,14 @@ var aivenProjectVPCSchema = map[string]*schema.Schema{
 		Description: "State of the VPC (APPROVED, ACTIVE, DELETING, DELETED)",
 		Type:        schema.TypeString,
 	},
+	`timeout`: {
```
rikonen, line 39: Maybe the name could be something less likely to conflict with a server side property we might add at a later time? Like client_create_wait_timeout or somesuch.

Also, this uses different quotation marks than all other properties.

Same comments for the other timeout.

Issue #212, opened 2020-04-21
Poor performance when refreshing kafka topics


hklasa:
Hello,

Yesterday we upgraded aiven terraform provider from version v1.1.3 to v1.2.3 and after that we observed 7 times longer freshness time for our HCL code. I can see that majority of it is spend on refreshing kafka topics.

Are You aware of this performance issue? Can it be fixed in some way?

ivan-savciuc:
Hello,

It is the first time when we here about this issue and perhaps it is related to a new Kafka Topic Cache layer that was implemented in v1.2.1. Can you please share more details about your Kafka Topics set up, for example, how many Kafka clusters and topics do you have?

damienleger:
Hello,

I can confirm this behavior by upgrading 1.0.19 -> 1.2.3

`time terraform plan` command with both versions:

*1.0.19*
real    0m26.103s
user    0m8.782s
sys     0m1.788s

*1.2.3*
real    2m26.216s
user    0m9.015s
sys     0m1.806s

My terraform code consists of 36 aiven_service_user, 108 aiven_kafka_acl, 52 aiven_kafka_topic and some negligible resources.

But the `terraform plan` seems to hang on some aiven_kafka_topic resources but not all that is weird. Like 5-10s on some line "aiven_kafka_topic (...) Refreshing state..."

damienleger:
> perhaps it is related to a new Kafka Topic Cache layer that was implemented in v1.2.1.
I have just tested upgrade in 1.2.0 -> no problem, but with 1.2.1 -> problem. Your guess is right @ivan-savciuc

ivan-savciuc:
Improvement is available in https://github.com/aiven/terraform-provider-aiven/releases/tag/v1.2.4

damienleger:
I confirm performance issue is gone in 1.2.4, thanks! 👍

PR #216, opened 2020-04-30
speed up kafka topic availability waiter

ivan-savciuc:
Improve Kafka Topic availability performance:
- remove waiter delay which sets TF timeout before start checking availability
- avoid multiple calls to Aiven API to refresh Topic Cache, since TF calls multiple waiters at the same time solution would be to make the first waiter refresh cache and put all others in the waiting position.

https://github.com/aiven/terraform-provider-aiven/issues/212

### DIFF
--- a/aiven/kafka_topic_availability.go
+++ b/aiven/kafka_topic_availability.go
@@ -3,6 +3,7 @@ package aiven
 import (
 	"fmt"
 	"github.com/aiven/terraform-provider-aiven/pkg/cache"
+	"golang.org/x/sync/semaphore"
 	"log"
 	"time"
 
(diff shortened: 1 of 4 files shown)


aiven/kafka_topic_availability.go
```diff
@@ -19,6 +20,16 @@ type KafkaTopicAvailabilityWaiter struct {
 	TopicName   string
 }
 
+var renewing int32
+
+func isRenewing() bool {
```
mteiste, line 25: isRenewing / finishRenewing as names do not pair that well, also isXX() kind of only implies a check not state change, so I would recommend changing isRenewing to something that makes it more obvious that on success the state has changed

ivan-savciuc: Indeed bad names, I did struggle to find an appropriate one and decided to use semaphores instead to make code a bit more understandable.

aiven/kafka_topic_availability.go
```diff
@@ -76,6 +70,28 @@ func (w *KafkaTopicAvailabilityWaiter) RefreshFunc() resource.StateRefreshFunc {
 	}
 }
 
+func (w *KafkaTopicAvailabilityWaiter) refresh() error {
+	if !kafkaTopicAvailabilitySem.TryAcquire(1) {
+		log.Printf("[TRACE] Kafka Topic Availability cache refresh function is bussy ...")
```
mteiste, line 75: busy

PR #222, opened 2020-05-14
Fix Jenkins job

ivan-savciuc:
Updated makefile using latest go and packr2 version, also disable `go vendor` to make it work properly with the modules and hopefully, it will fix a Jenkins job.


mteiste:
I'll do a test in a clean machine to verify the resulting binaries and if they are ok I'll merge.

PR #228, opened 2020-05-20
Adjust provider schema to make api_token optional

ebachle:
Terraform hinting in VSCode indicates that when you leave off the `api_token` field that it's invalid.  Because there's a `DefaultFunc`, this will get read in from the environment variable `AIVEN_TOKEN` executed successfully regardless of the hinting, but this fixes that hinting.

Tested with `make test` and by running my terraform code.

### DIFF
--- a/aiven/provider.go
+++ b/aiven/provider.go
@@ -22,7 +22,6 @@ func Provider() terraform.ResourceProvider {
 			"api_token": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Default:     "",
 				Sensitive:   true,
 				DefaultFunc: schema.EnvDefaultFunc("AIVEN_TOKEN", nil),
 				Description: "Aiven Authentication Token",
--- a/aiven/provider.go
+++ b/aiven/provider.go
@@ -21,7 +21,8 @@ func Provider() terraform.ResourceProvider {
 		Schema: map[string]*schema.Schema{
 			"api_token": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
+				Default:     "",
 				Sensitive:   true,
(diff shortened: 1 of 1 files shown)


ivan-savciuc:
If we make `api_token` as an optional for the provider, it might cause more harm than good. First of all, you need a token either set in a TF file or via environment variable to communicated with Aiven API.

For example, If we set it to optional and a user did not set it in any way (TF, env) then when a user tries to run something like `terraform refresh` he/she will get an error:
``` 
{"errors":[{"message":"Invalid token","status":401}],"message":"Invalid token"}
```

Therefore we need to make sure on the provider side that `api_token` is set before we start any further actions.

ivan-savciuc:
The idea with the `DefaultFunc` is that if you like to leave `api_token` empty in a TF file please set an ENV variable or visa versa, but it should be set in some way, and TF client should communicate it to the user.

ebachle:
@ivan-savciuc I'm not sure I'm on the same page as you there.  Other providers provide this as optional in the schema (https://github.com/terraform-providers/terraform-provider-datadog/blob/master/datadog/provider.go#L27-L30)

From there, they use the `configureFunc` to validate that there is a value set for it.  https://github.com/terraform-providers/terraform-provider-datadog/blob/master/datadog/provider.go#L97-L100

Would you be open to a more comprehensive PR that implements that in the configureFunc so that the schema validation that happens in IDEs matches what is actually required?  

One interesting caveat is that the Provider Docs (https://www.terraform.io/docs/providers/datadog/index.html#argument-reference) still reference that field as being required, but with the caveat that it can be provided in the `provider` block or in an environment variable, somewhat similar to what is in the README here.

ivan-savciuc:
@ebachle I think we are on the same page if add `api_token` validation logic to `ConfigureFunc` we can make `api_token` field optional.

PR #231, opened 2020-05-27
add TF native timeouts support

ivan-savciuc:
`client_timeout `is marked as deprecated and instead, and users welcome to use native Terraform `timeouts` functionality

### DIFF
--- a/aiven/provider.go
+++ b/aiven/provider.go
@@ -203,14 +203,15 @@ func validateDurationString(v interface{}, k string) (ws []string, errors []erro
 
 // generateClientTimeoutsSchema generates client_timeout Terraform schema
 // based on name of the timeout and default duration
+// Deprecated: generateClientTimeoutsSchema exists for historical compatibility
+// and should not be used. To set timeouts use native TF timeouts functionality.
 func generateClientTimeoutsSchema(timeouts map[string]time.Duration) *schema.Schema {
 	schemaTimeouts := map[string]*schema.Schema{}
-	for name, duration := range timeouts {
+	for name := range timeouts {
 		schemaTimeouts[name] = &schema.Schema{
 			Type:         schema.TypeString,
 			Description:  name + " timeout",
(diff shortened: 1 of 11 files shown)


rikonen:
I presume resource_pg.go should not be included in this pull request

ivan-savciuc:
> I presume resource_pg.go should not be included in this pull request

Yes, sorry about that, this slipped through from another PR on which I'm working right now

aiven/provider.go
```diff
@@ -219,22 +220,30 @@ func generateClientTimeoutsSchema(timeouts map[string]time.Duration) *schema.Sch
...
 		Elem: &schema.Resource{
 			Schema: schemaTimeouts,
 		},
 	}
 }
 
 // getTimeoutHelper is a helper which extract from a resource data client timeout
+// Deprecated: getTimeoutHelper exists for historical compatibility
+// and should not be used. To set timeouts use native TF timeouts functionality.
 func getTimeoutHelper(d *schema.ResourceData, name string, defaultDuration time.Duration) (time.Duration, error) {
```
rikonen, line 235: Should defaultDuration be dropped and zero returned instead if a particular timeout isn't defined. Right now if you specify "client_timeout" for "read" and "timeout" for "update" the default value will be used for "update" instead of what was set in the config. That's of course a bit unlikely config but still doesn't seem like there's any benefit of defining default value here and there is some harm.

ivan-savciuc: I think this should do the trick since `client_timeout` has no default value in TF resource defenition:
```
clientTimeouts, ok := d.GetOk("client_timeout")
	if !ok || clientTimeouts.(*schema.Set).Len() == 0 {
		return 0, nil
	}```

But I think it is a good idea to remove `defaultDuration` for better readability and maybe I'm missing something.

ivan-savciuc: Did reread your commend - indeed you are right!

PR #233, opened 2020-05-31
Allow wildcard for username in Kafka ACL

encima:
Default Kafka ACL is `*`. Alphanumeric check does not allow this to be removed before adding new ACLs. Regexp has been changed to allow either a single `*` or an alphanumeric string.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,5 +10,4 @@ Description: The Aiven Terraform Provider is a way to access all of your Aiven s
 search_enabled: true
 exclude:
   - scripts/
-
 theme: jekyll-theme-minimal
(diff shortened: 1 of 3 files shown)


ivan-savciuc:
After some testing, it looks like we should extend the regexp to allow alpha-numeric characters, underscores, dashes, dots and glob characters '*' and '?'. Username `test-.?*` is legit and API accepts it.

encima:
It is legit and the API accepts it, but Kafka won't. Only `*` is allowed OR a string. We could make the second capture group more open, though, is that what you mean?

ivan-savciuc:
> It is legit and the API accepts it, but Kafka won't. Only `*` is allowed OR a string. We could make the second capture group more open, though, is that what you mean?

Indeed Kafka won't accept it. Then let's just update tests with `*` use-case, and if you think it might be useful, we can extend the second capture group.

PR #239, opened 2020-06-12
JSON Schema documentation for user_config objects

encima:
Modifications are quite big here. Introduced two libs:
1. Grafana JSON Schema Go lib - Generate markdown from JSON schema files
2. `weaming/json-schema-generator` - Generate JSON Schemas from User config responses.

Addition of second library adds a build step to Travis but it means docs are auto generated now so the overhead is worth it.


ivan-savciuc:
Can we avoid downgrading Golang from 1.3.x to 1.2.x (I'm planning to upgrade it even further to 1.4 at some point in time)?

encima:
> Can we avoid downgrading Golang from 1.3.x to 1.2.x (I'm planning to upgrade it even further to 1.4 at some point in time)?

It is still using 1.3.x (tests are run in 1.3 and go.mod specifies 1.3 also). It is only docs that are built in the step with 1.2.x. I think there is a reason for it but running it through Travis now to see if I remember wrong.

PR #240, opened 2020-06-12
user configuration options cache + packr2 refactoring 

ivan-savciuc:
The idea is to simplify code and dependencies, instead of using packr2 and caching the content of in-memory JSON files we can use `go generate`. It will generate Golang files with corresponding inline JSON and update `templates` package level variable with schemas content.


ivan-savciuc:
A little bit more regarding how it will work if we decide to merge this PR:
When a user runs `go generate` Golang triggers `aiven/templates/gen.sh` shell script to generate `*_user_config_schema.go` files with variables containing unmarshaled JSON from `*_user_config_schema.json` files. We include autogenerated files to the VCS as it is recommended in Golang community (though we don't have to do that).

mteiste:
Looks ok, rebase and I'll merge this

PR #245, opened 2020-06-23
Add guide for adding syslog (or similar) service integrations

encima:
Based on customer request. Basic example with sample script.

### DIFF
--- a/docs/guides/syslog-service-integration.md
+++ b/docs/guides/syslog-service-integration.md
@@ -21,7 +21,7 @@ In this example, we already have a project (`my-proj`) and a service (`kafka-ser
 Then, we define the `aiven_service_integration_endpoint`.
 
 The important things here are:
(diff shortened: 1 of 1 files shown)


docs/guides/syslog-service-integration.md
```diff
@@ -0,0 +1,74 @@
...
+2. Creating an `aiven_service_integration` - This will link the endpoint to a running Aiven service.
+
+## Example
+
+In this example, we already have a project (`my-proj`) and a service (`kafka-service1`) running. This means we can use the `datasource` to pull in those objects from Aiven.
+
+Then, we define the `aiven_service_integration_endpoint`.
+
+The important things here are:
+* endpoint_type - The type of integration (e.g. `rsyslog`, `prometheus`, `rsyslog`)
```
ivan-savciuc, line 24: `rsyslog` is mentioned two times here, maybe we can replace one with `datadog` for example.

encima: Thanks, done

PR #265, opened 2020-07-23
add support for transit gateway vpc attachment

ivan-savciuc:


rikonen:
I haven't checked the server side implementation but is there anything in the properties returned via the API that distinguishes transit gateway from regular peering connection? Was just thinking if e.g. import should check that the resource is in fact transit gateway vpc attachment.

ivan-savciuc:
The only difference is `user_peer_network_cidrs` in comparison with regular VPC peering connection. My understanding is that Aiven is using the same API endpoint for both types but logically distinguishes between them by `user_peer_network_cidrs` if it present in the request and response we are talking about VPC attachment.

aiven/resource_vpc_peering_connection.go
```diff
@@ -262,6 +269,12 @@ func copyVPCPeeringConnectionPropertiesFromAPIResponseToTerraform(
 		return err
 	}
 
+	if len(peeringConnection.UserPeerNetworkCIDRs) > 0 {
+		if err := d.Set("user_peer_network_cidrs", peeringConnection.UserPeerNetworkCIDRs); err != nil {
+			return err
+		}
+	}
```
rikonen, line 276: Should this not clear old cidrs in case list is empty?

ivan-savciuc:
And there is a separate API endpoint to manage CIDR`s after https://api.aiven.io/doc/#operation/Project.VpcPeeringConnectionUpdate after VPC creation

aiven/resource_transit_gateway_vpc_attachement.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}
+
+		if forDeletion {
+			deleteCIDRs = append(deleteCIDRs, old)
+		}
+	}
+
+	_, err = client.TransitGatewayVPCAttachment.Update(projectName, vpcID, aiven.TransitGatewayVPCAttachmentRequest{
+		Add:    nil,
```
rikonen, line 131: Should be `Add: add`?

aiven/resource_vpc_peering_connection.go
ivan-savciuc: Indeed, I was trying here to distinguish between Gateway Attachment and regular VPC by checking this parameter empty, but it introduces this problem.

PR #270, opened 2020-08-07
V1 - new configuration options

ivan-savciuc:
Add support for latest user configuration options for version 1


ivan-savciuc:
All deprecated service versions still are available in user configuration options to support existing customers using them.

PR #276, opened 2020-08-12
Kafka ACL regex modification

encima:
Add check that * is not first char of strings of length > 1
Some tests (KafkaACLWildcardPrefix) were reporting as being skipped so I will check if this is happening in CI as well.

### DIFF
--- a/aiven/resource_kafka_acl_test.go
+++ b/aiven/resource_kafka_acl_test.go
@@ -44,7 +44,12 @@ func TestAccAivenKafkaACL_basic(t *testing.T) {
 				ExpectError: regexp.MustCompile("invalid value for username"),
 			},
 			{
-				Config:             testAccKafkaACLWildcardResource(rName),
+				Config:      testAccKafkaACLInvalidCharsResource(rName),
+				PlanOnly:    true,
+				ExpectError: regexp.MustCompile("invalid value for username"),
+			},
+			{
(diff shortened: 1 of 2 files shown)


aiven/resource_kafka_acl_test.go
```diff
@@ -44,7 +44,12 @@ func TestAccAivenKafkaACL_basic(t *testing.T) {
 				ExpectError: regexp.MustCompile("invalid value for username"),
 			},
 			{
-				Config:             testAccKafkaACLWildcardResource(rName),
```
ivan-savciuc, line 47: Can we keep `testAccKafkaACLWildcardResource` test as well?

aiven/templates/endpoint_user_config_schema.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by go generate; DO NOT EDIT.
-// This file was generated at Tue 14 Jul 10:56:39 CEST 2020
```
ivan-savciuc, line 2: These changes are not relevant to Kafka ACL regex modification

aiven/resource_kafka_acl_test.go
encima: It is still in there, line 52, just the name was modified

aiven/templates/endpoint_user_config_schema.go
encima: Good point, removed

PR #285, opened 2020-08-18
Add Pg example for demo video

davidespo:
Example creates 3 pg clusters on multiple clouds in multiple regions.


encima (changes requested):
Could you also update the root `README` with a link to the example and an outline of what it does?

Everything else LGTM

encima (approved):

PR #286, opened 2020-08-19
Accounts example

davidespo:
This should have been forked from Master, but includes PG changes from PR #285.

Also, not sure how we feel about hosting assets inside this repo. The README includes relative image links to screenshots. 

https://github.com/aiven/terraform-provider-aiven/tree/accounts-ex/examples/account


encima:
> This should have been forked from Master, but includes PG changes from PR #285.
> 
> Also, not sure how we feel about hosting assets inside this repo. The README includes relative image links to screenshots.
> 
> https://github.com/aiven/terraform-provider-aiven/tree/accounts-ex/examples/account

My personal view is that it is fine (as long as we do not need LFS :)) 

We do not have a fixed place to store them anyway and storing them outside of the repo makes it hard to view offline

encima (changes requested):
README here is much more in-depth (I know the PG #285  has video context)

Won't merge until #285 is merged. Same request applies for the root `README` but let me know if you want me to add these.

encima (approved):

PR #288, opened 2020-08-24
add service components

ivan-savciuc:


aiven/datasource_service_component.go
```diff
@@ -0,0 +1,116 @@
...
+	componentName := d.Get("component").(string)
+	route := d.Get("route").(string)
+
+	service, err := client.Services.Get(projectName, serviceName)
+	if err != nil {
+		return fmt.Errorf("service %s/%s not found: %w", projectName, serviceName, err)
+	}
+
+	for _, c := range service.Components {
+		if c.Component == componentName && c.Route == route && c.Usage == "primary" {
```
rikonen, line 80: Usage should be possible to define as data source property so that one can get MySQL/PostgreSQL/Redis master and standby info as well. Possible values are primary, syncing (node is still being initialized but already accessible) and replica.

aiven/datasource_service_component.go
```diff
@@ -0,0 +1,116 @@
...
+			if err := d.Set("port", c.Port); err != nil {
+				return err
+			}
+			if err := d.Set("usage", c.Usage); err != nil {
+				return err
+			}
+			if err := d.Set("kafka_authentication_method", c.KafkaAuthenticationMethod); err != nil {
+				return err
+			}
+			if err := d.Set("ssl", c.Ssl); err != nil {
```
rikonen, line 107: Shouldn't ssl and kafka_authentication_method be user defined attributes that allow defining specific data source? It is possible to have both certificate authentication and SASL enabled for the same service as well as have both SSL enabled and non-SSL enabled connections available.

rikonen (changes requested):

aiven/datasource_service_component.go
ivan-savciuc: will add that

rikonen:
The change for Kafka authentication method and SSL attributes was not done. Either I don't understand how this is intended to be used or then at the moment if you specify a data source that matches specific Kafka service that has both certificate and SASL authentication enabled then the data source will randomly get attributes from one of those. If the service further has both SSL and non-SSL connectivity enabled the data source will randomly get attributes of the SSL or non-SSL connection.

ivan-savciuc:
> The change for Kafka authentication method and SSL attributes was not done. Either I don't understand how this is intended to be used or then at the moment if you specify a data source that matches specific Kafka service that has both certificate and SASL authentication enabled then the data source will randomly get attributes from one of those. If the service further has both SSL and non-SSL connectivity enabled the data source will randomly get attributes of the SSL or non-SSL connection.

Here is what I was thinking, let's say we have this list of components for a service:

```
"components": [{
			"component": "kafka",
			"host": "XXX.aivencloud.com",
			"kafka_authentication_method": "certificate",
			"port": 123,
			"route": "dynamic",
			"usage": "primary"
		}, {
			"component": "kafka",
			"host": "YYY.com",
			"kafka_authentication_method": "sasl",
			"port": 234,
			"route": "dynamic",
			"usage": "primary"
		}
...
}]
```

If a user does something like: 
```
data "aiven_service_component" "kafka" {
	project = aiven_kafka.bar.project
	service_name = aiven_kafka.bar.service_name
	component = "kafka"
	route = "dynamic"
}
```

A user will get the first entry from a list if he/she makes a more specific request by adding `kafka_authentication_method` = "sasl" datasource will get a second item. Of course, it can be a bit more convenient if we would make  `kafka_authentication_method` a required property and force the user to set all the time, but in this case, it will be less convenient for users trying to query a PG or any other then Kafka service type components. In this case, they would need to explicitly set `kafka_authentication_method` to an empty string all the time.

I think it is OK to get a "random" entire from a list of components if a query is not specific enough. If you do have a better idea or a strong opinion of how we should implement this, please let me know.

rikonen:
I was looking at the documentation, didn't notice the code was changed. The documentation should at least be updated to reflect this. The fact that the data source can change to point to completely different and potentially insecure connection type does sound like something that may end up causing problems.

Perhaps the most convenient but still safe behavior would be that if ssl isn't defined for datasource we only match entries that are missing ssl property or have it set to true and likewise if kafka_authentication_method isn't defined we only match entries with no value here or value certificate. That way the default case would work without having to define these parameters and nothing would break if additional connectivity options were later enabled.

ivan-savciuc:
@rikonen please check this out, if I understood it correctly. 

We leave `ssl` and `kafka_authentication_method` fields option, but follow more strict rules, if any of these options are not set, we filter out all components such that they do not contain any values for them.

aiven/datasource_service_component.go
```diff
@@ -0,0 +1,150 @@
...
+			"kafka_authentication_method": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Kafka authentication method. This is a value specific to the 'kafka' service component",
+			},
+			"ssl": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Description: "Whether the endpoint is encrypted or accepts plaintext. By default endpoints are " +
+					"always encrypted and this property is only included for service components they may " +
```
rikonen, line 44: `they may` => `that may`

README.md
```diff
@@ -1196,6 +1196,38 @@ peered with.
...
+`component` is a service component name.
+
+`route` is network access route.
+
+`host` is DNS name for connecting to the service component.
+
+`port` is port number for connecting to the service component.
+
+`kafka_authentication_method` is Kafka authentication method. This is a value specific 
+to the 'kafka' service components. If not set by the user only entries with empty with
```
rikonen, line 1221: `only entries with empty with` => `only entries with empty`

README.md
```diff
@@ -1196,6 +1196,38 @@ peered with.
...
+
+`port` is port number for connecting to the service component.
+
+`kafka_authentication_method` is Kafka authentication method. This is a value specific 
+to the 'kafka' service components. If not set by the user only entries with empty with 
+`kafka_authentication_method` will be selected.
+
+`ssl` whether the endpoint is encrypted or accepts plaintext. By default endpoints are
+always encrypted and this property is only included for service components they may
+disable encryption. If not set by the user only entries with empty with `ssl` will be
```
rikonen, line 1226: `only entries with empty with` => `only entries with empty`

README.md
```diff
@@ -1196,6 +1196,38 @@ peered with.
...
+
+`route` is network access route.
+
+`host` is DNS name for connecting to the service component.
+
+`port` is port number for connecting to the service component.
+
+`kafka_authentication_method` is Kafka authentication method. This is a value specific 
+to the 'kafka' service components. If not set by the user only entries with empty with 
+`kafka_authentication_method` will be selected.
```
rikonen, line 1222: This could mention the available options, `certificate` and `sasl`.

aiven/datasource_service_component.go
```diff
@@ -0,0 +1,150 @@
...
+			},
+			"route": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Network access route",
+			},
+			"kafka_authentication_method": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Kafka authentication method. This is a value specific to the 'kafka' service component",
```
rikonen, line 38: Available options could be mentioned here too.

aiven/datasource_service_component.go
```diff
@@ -0,0 +1,150 @@
...
+
+	for _, c := range service.Components {
+		if c.Component == componentName && c.Route == route && c.Usage == usage {
+			// check optional ssl search criteria, if not set by a user match entries without ssl
+			if ssl, ok := d.GetOk("ssl"); ok {
+				if c.Ssl != strconv.FormatBool(ssl.(bool)) {
+					continue
+				}
+			} else {
+				if c.Ssl != "" {
```
rikonen, line 89: I would accept both empty and true here, otherwise if we don't currently allow non-SSL connections for specific connection type but later add support for it then this would stop matching the same entry it used to match before the change.

rikonen (changes requested):

README.md
```diff
@@ -1196,6 +1196,39 @@ peered with.
...
+    route = "dynamic"
+}
+```
+
+`project` and `service_name` define the project and service the service component
+belongs to.
+
+`component` is a service component name.
+
+`route` is network access route.
```
rikonen, line 1214: Forgot to comment earlier but this should list different route options and what those actually mean. Something like

Route may be one of `dynamic`, `public`, and `private`. Usually you'll want to use `dynamic`, which for services that are not in private network identifies the regular public DNS name of the service and for services in private network the private DNS name. If the service is in private network but has also public access enabled the `public` route type can be used to get the public DNS name of the service. The `private` option should typically not be used.

README.md
```diff
@@ -1196,6 +1196,39 @@ peered with.
...
+    service_name = aiven_kafka.service1.service_name
+    component = "kafka"
+    route = "dynamic"
+}
+```
+
+`project` and `service_name` define the project and service the service component
+belongs to.
+
+`component` is a service component name.
```
rikonen, line 1212: This should mention something about the possible values. Like

`component` may match the name of the service (`cassandra`, `elasticsearch`, `grafana`, `influxdb`, `kafka`, `kafka_connect`, `mysql`, `pg` and `redis`), in which case the connection info of the service itself is returned. Some service types support additional service specific components like `kibana` for Elasticsearch, `kafka_connect`, `kafka_rest` and `schema_registry` for Kafka, and `pgbouncer` for PostgreSQL. Most service types also support `prometheus`.

README.md
```diff
@@ -1196,6 +1196,39 @@ peered with.
...
+`port` is port number for connecting to the service component.
+
+`kafka_authentication_method` is Kafka authentication method. This is a value specific 
+to the 'kafka' service components. And has the following available options: `certificate` 
+and `sasl`. If not set by the user only entries with empty `kafka_authentication_method` 
+will be selected.
+
+`ssl` whether the endpoint is encrypted or accepts plaintext. By default endpoints are
+always encrypted and this property is only included for service components they may
+disable encryption. If not set by the user only entries with empty `ssl` will be
```
rikonen, line 1227: empty `ssl` or `ssl` set to `true`.

README.md
```diff
@@ -1196,6 +1196,39 @@ peered with.
 
 `peer_vpc` defines the identifier or name of the remote VPC.
 
+### Datasource Service Component
```
rikonen, line 1199: This could start by saying something like

Service components can be defined to get the connection info for specific service. Services may support multiple different access routes (VPC peering and public access), have additional components or support multiple authenticatoin methods. Each of these may be represented by different DNS name or TCP port and the specific component to match can be selected by specifying appropriate filters as shown below.

README.md
```diff
@@ -1196,6 +1196,39 @@ peered with.
...
+```hcl-terraform
+data "aiven_service_component" "sc1" {
+    project = aiven_kafka.project1.project
+    service_name = aiven_kafka.service1.service_name
+    component = "kafka"
+    route = "dynamic"
```
rikonen, line 1205: This should also define `kafka_authentication_method = "certificate"` as otherwise with current logic this would not match to anything.

rikonen (commented):
The actual code looks good. Documentation could still be clarified to make this functionality more accessible, details in comments.

ivan-savciuc:
@rikonen thanks for your documentation comments, it helps since it is nowhere documented. What about  `usage` can it be something else than `primary`?

rikonen:
Usage can be one of `primary`, `replica` or `syncing`. `replica` is used by services that have separate master and standby roles for which it identifies the `replica` DNS name. `syncing` is used by limited set of services to expose nodes before they have finished restoring state but may already be partially available, for example a PostgreSQL node that is streaming WAL segments from backup or current master but hasn't yet fully caught up.

rikonen (approved):

PR #290, opened 2020-08-25
fix vpc user_peer_network_cidrs type conversion problem

ivan-savciuc:


aiven/resource_transit_gateway_vpc_attachement.go
```diff
@@ -81,7 +81,7 @@ func resourceTransitGatewayVPCAttachment() *schema.Resource {
 func resourceTransitGatewayVPCAttachmentUpdate(d *schema.ResourceData, m interface{}) error {
 	client := m.(*aiven.Client)
 
-	cidrs := d.Get("user_peer_network_cidrs").([]string)
+	cidrs := flattenToString(d.Get("user_peer_network_cidrs").([]interface{}))
```
rikonen, line 84: Fwiw this function name is quite confusing. I'd expect it to flatten an array into string instead of mapping individual array elements into strings.

rikonen (approved):

PR #293, opened 2020-08-28
add kafka schema subject compatibility level

ivan-savciuc:


rikonen:
With this change, how will schemas for which user doesn't define compatibility behave? Will updating Terraform state cause the global default compatibility level to be applied for all individual schemas and when you apply the state all schemas will end up having explicitly the current global default and changing the default will no longer affect the existing schemas even though they do not define the compatibility attribute in Terraform state, i.e. should be using default? (I would assume it ends up working like that, based on how the API is currently implemented might not be possible to make it work otherwise and we may need to change the API somehow.)

ivan-savciuc:
> With this change, how will schemas for which user doesn't define compatibility behave? Will updating Terraform state cause the global default compatibility level to be applied for all individual schemas and when you apply the state all schemas will end up having explicitly the current global default and changing the default will no longer affect the existing schemas even though they do not define the compatibility attribute in Terraform state, i.e. should be using default? (I would assume it ends up working like that, based on how the API is currently implemented might not be possible to make it work otherwise and we may need to change the API somehow.)

Indeed it will behave as you described, in this case, perhaps we can update individual subject compatibility levels only when it differs from the current global level? In this case, when a global compatibility level changes subjects that had a previous global default value will notify TF users about changes in plan, even though they did not touch it. Would that work?

rikonen:
Only updating levels when they differ from global default might not be good either as explicitly defining level is different from relying on the global default and if you have explicitly defined schema to have certain compatibility level you don't want it to change when the global default does. I think we should change the schema config API to somehow communicate whether the schema itself has some explicit config set or not.

ivan-savciuc:
Ok, then we convert this PR to draft until API changes are available?

rikonen:
I suppose. I can make the change. Would simply making `GET /project/<project>/service/<service_name>/kafka/schema/config/<subject_name>` return 404 when the schema doesn't have explicit config work from your point-of-view? (Cannot change the default behavior so will need to add some new parameter there to instruct the server to work like that instead of returning the global default.)

ivan-savciuc:
That would, thanks!

rikonen:
Internal pull request for this is now done. Ping me if I forget to update here when that gets deployed. The new parameter (unless changes requested) will be `global_default_fallback=false` (query parameter). With that the response will be 404 instead of global default for schema subjects lacking explicit config.

ivan-savciuc:
PR was updated with the latest Go-client that is using `global_default_fallback=false` parameter for getting Schema Subject configuration.

aiven/datasource_service_component.go
```diff
@@ -115,11 +115,11 @@ func datasourceServiceComponentRead(d *schema.ResourceData, m interface{}) error
 			// check optional ssl search criteria, if not set by a user match entries
 			// without ssl or ssl=true
 			if ssl, ok := d.GetOk("ssl"); ok {
-				if c.Ssl != strconv.FormatBool(ssl.(bool)) {
+				if *c.Ssl != ssl.(bool) {
```
rikonen, line 118: `c.Ssl` might be `nil` if someone specifies ssl attribute for service component that doesn't have it on server side. Skipping over would be better than nil pointer dereference.

rikonen (changes requested):

rikonen (approved):

PR #300, opened 2020-09-07
extend VPC peering connection creation 

ivan-savciuc:
extend VPC peering connection creation with new azure related optional fields


rikonen:
I'm guessing the ForceNew doesn't cause any problems if someone updates from older provider version that didn't yet support these settings to one that does? (I.e. not having those for some non-Azure VPC does not suddenly cause the VPC to be re-created.)

ivan-savciuc:
Since fields are optional, they won't force the creation of the new VPC peering connection after upgrade to the new provider.

rikonen (approved):

PR #305, opened 2020-09-10
add kafka connector read waiter

ivan-savciuc:


aiven/resource_kafka_connector.go
```diff
@@ -117,13 +123,21 @@ func flattenKafkaConnectorTasks(r *aiven.KafkaConnector) []map[string]interface{
 func resourceKafkaConnectorRead(d *schema.ResourceData, m interface{}) error {
 	project, serviceName, connectorName := splitResourceID3(d.Id())
 
-	res, err := m.(*aiven.Client).KafkaConnectors.List(project, serviceName)
+	// Wait until the peering connection has actually been built
```
rikonen, line 126: Comment is off

aiven/resource_kafka_connector.go
```diff
@@ -241,3 +255,38 @@ func resourceKafkaConnectorState(d *schema.ResourceData, m interface{}) ([]*sche
...
+	Client      *aiven.Client
+	Project     string
+	ServiceName string
+}
+
+// RefreshFunc will call the Aiven client and refresh it's state.
+func (w *KafkaConnectorWaiter) RefreshFunc() resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		list, err := w.Client.KafkaConnectors.List(w.Project, w.ServiceName)
+		if err != nil {
```
rikonen, line 270: Should this be more specific about which error is waited on? If the response is 404 then presumably the connector just doesn't exist.

ivan-savciuc: We won't be waiting when an error is 404, only in case of a different errors

rikonen (approved):

PR #326, opened 2020-09-30
docs improvement #1 

ivan-savciuc:
This is the first out of many upcoming PR's, changing documentation in a compatible way with Terraform Registry https://www.terraform.io/docs/registry/providers/docs.html


encima (approved):
All that beautiful markdown 😢 
Agreed that this is the best fix for now, I would say that we should stop the `config` auto generation indefinitely as it should come from the central API; once it is documented there then it can be pulled from the OpenAPI spec

PR #348, opened 2020-10-27
Add guide for documenting issues encountered during upgrades

chornsby:
This PR adds an initial guide document where we can record issues encountered during provider upgrades. I'm open for any and all restructuring / rewording / reorganising, just wanted to write something down for this case.


ivan-savciuc (commented):
Perhaps we can also mention that if user configuration options field is not set (gone from the TF file), that would also do the trick and users don't have to bother setting it to a ".

ivan-savciuc (approved):

PR #354, opened 2020-11-06
optimise kafka topic resource performance

ivan-savciuc:
The idea of PR is to optimise Kafka Topic performance. Since we are using a new Kafka Topic API endpoint to get Kafka Topic configuration in addition to list endpoint, it drastically increases the number of API calls. Kafka Topic endpoint may behave unstable when a user has 200+ topics and Terraform client performing multiple operations simultaneously. 

To remedy this, we reduce the number of API calls where possible, tune timings and improve error handling to reduce the number of API calls.

### DIFF
--- a/aiven/kafka_topic_availability.go
+++ b/aiven/kafka_topic_availability.go
@@ -122,7 +122,7 @@ func (w *KafkaTopicAvailabilityWaiter) Conf(timeout time.Duration) *resource.Sta
 		Target:     []string{"ACTIVE"},
 		Refresh:    w.RefreshFunc(),
 		Timeout:    timeout,
-		MinTimeout: 1 * time.Second,
+		MinTimeout: 10 * time.Second,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


rikonen:
Does the v2 topic listing endpoint not return the necessary info?

ivan-savciuc:
> Does the v2 topic listing endpoint not return the necessary info?

Nope, it has no `config` according to API docs https://api.aiven.io/doc/#operation/ServiceKafkaTopicList

rikonen:
I meant `POST /v2/project/<project>/service/<service_name>/topic`. That actually seems to be missing from documentation as it has been marked as experimental.

It takes JSON body in request, which must contain `"topic_names"` list identifying the topics whose details to retrieve and optionally takes `"config_names"` array identifying the names of the config options to return (default is to return all). Number of topics for single call should be kept somehow reasonable, say no more than 200.

That API will not work for some older services so the changes in this pull request may make sense anyway but could you look into implementing more optimal behavior with that API?

ivan-savciuc:
Nice, I will create a CH ticket for that

encima:
Is this pending the CH Topic/adding of endpoint to the public API?

rikonen (approved):

rikonen:
This change looks sensible regardless of future updates. I had a look at the v2 get topic configs in batch API and as far as I can tell it should be good to go. I'd make it public once we've demonstrated it works with the Terraform provider (i.e. when releasing new version of the provider that utilizes that API).

PR #364, opened 2020-11-13
Change VPC peering connection state handling

ivan-savciuc:
- We update the VPC peering connection code eliminating all TF SDK v2 deprecated function, updating them with the context. 
- Change creation function, such that only when VPC peering connection is in the `ACTIVE` state - TF resource is created. Also, the extended error messages aim to communicate a guide a user to achieve the `ACTIVE` state.

### DIFF
--- a/aiven/resource_vpc_peering_connection.go
+++ b/aiven/resource_vpc_peering_connection.go
@@ -2,8 +2,10 @@
 package aiven
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
 	"strings"
 	"time"
@@ -78,12 +80,11 @@ var aivenVPCPeeringConnectionSchema = map[string]*schema.Schema{
 
 func resourceVPCPeeringConnection() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceVPCPeeringConnectionCreate,
-		Read:   resourceVPCPeeringConnectionRead,
-		Delete: resourceVPCPeeringConnectionDelete,
-		Exists: resourceVPCPeeringConnectionExists,
+		CreateContext: resourceVPCPeeringConnectionCreate,
+		ReadContext:   resourceVPCPeeringConnectionRead,
+		DeleteContext: resourceVPCPeeringConnectionDelete,
 		Importer: &schema.ResourceImporter{
-			State: resourceVPCPeeringConnectionState,
+			StateContext: resourceVPCPeeringConnectionImport,
 		},
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(2 * time.Minute),
@@ -93,7 +94,7 @@ func resourceVPCPeeringConnection() *schema.Resource {
 	}
 }
 
-func resourceVPCPeeringConnectionCreate(d *schema.ResourceData, m interface{}) error {
+func resourceVPCPeeringConnectionCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	var (
 		pc     *aiven.VPCPeeringConnection
 		err    error
@@ -104,7 +105,7 @@ func resourceVPCPeeringConnectionCreate(d *schema.ResourceData, m interface{}) e
 	client := m.(*aiven.Client)
 	projectName, vpcID := splitResourceID2(d.Get("vpc_id").(string))
 	if projectName == "" || vpcID == "" {
-		return errors.New("incorrect VPC ID, expected structure <PROJECT_NAME>/<VPC_ID>")
+		return diag.Errorf("incorrect VPC ID, expected structure <PROJECT_NAME>/<VPC_ID>")
 	}
 
 	peerRegion := d.Get("peer_region").(string)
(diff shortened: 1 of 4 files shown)


aiven/resource_vpc_peering_connection.go
```diff
@@ -378,7 +467,6 @@ func (w *VPCPeeringBuildWaiter) Conf(timeout time.Duration) *resource.StateChang
 			"ACTIVE",
 			"REJECTED_BY_PEER",
 			"PENDING_PEER",
-			"INVALID_SPECIFICATION",
```
rikonen, line 381: Why was this removed? This is usually a terminal state and there's no point waiting more if the peering connection goes into this state.

aiven/resource_vpc_peering_connection.go
```diff
@@ -159,19 +160,65 @@ func resourceVPCPeeringConnectionCreate(d *schema.ResourceData, m interface{}) e
...
 
 	pc = res.(*aiven.VPCPeeringConnection)
+	if pc.State != "ACTIVE" {
+		switch pc.State {
+		case "PENDING_PEER":
+			return diag.Errorf("Aiven platform has created a connection to the specified "+
+				"peer successfully in the cloud, but the connection is not active until the user "+
+				"completes the setup in their cloud account. The steps needed in the user cloud "+
+				"account depend on the used cloud provider. Find more in the state info: %s",
+				stateInfoToString(pc.StateInfo))
```
rikonen, line 176: I wonder about this pending peer handling as this is the typical state the connection ends up in. Maybe check with @mwfrojdman what the expected correct flows would be. (Maybe it is this but I'm not sure.)

rikonen (changes requested):
Typo in the commit message / pull request title, handeling => handling

aiven/resource_vpc_peering_connection.go
ivan-savciuc: Indeed, added it back

aiven/resource_vpc_peering_connection.go
mwfrojdman: The happy path is
- create vpc peering connection through the aiven api (initial state `APPROVED`)
- wait for aiven to build the connection (-> state `PENDING_PEER`)
- Perform cloud provider specific actions in your own account to complete the setup
- Wait for the Aiven peering connection to enter state `ACTIVE`

On Azure some additional setup is needed before the steps described above

If the peering connection enters the state `INVALID_SPECIFICATION`, the peer vpc probably does doesn't exist, or some permissions are lacking depending on cloud provider. The error message may give some hints

ivan-savciuc: The idea behind this code is "block" the creation of ``` resource until it is in the `ACTIVE` state. The motivation was to notify a user whenever he/she need to perform some cloud provider-specific actions or change something in the configuration. 

Imaginable path: 
- When a user is creating `aiven_vpc_peering_connection` provider waits until resource reaches any state other than `APPROVED`.
- Let's say after `APPROVED` state we've got `PENDING_PEER`, in this case, creation of `aiven_vpc_peering_connection` will fail and a user will get a message with an invitation to perform some cloud provider-specific actions, also we print contents of the `state_info`. 
- A user performs required cloud provider-specific actions and runs `terraform apply` again.
- `aiven_vpc_peering_connection` is in the `ACTIVE` state, everyone is happy.

ivan-savciuc: @rikonen @mwfrojdman what would be desired behaviour for `PENDING_PEER` in Terraform Provider taking into consideration this thread?

rikonen:
I defer to @mwfrojdman as I'm not so familiar with the VPC flows.

encima (commented):
Tested running locally but would feel more comfortable with a VPC pro approving

prem-aiven-io:
@mwfrojdman - can you please review this?

aiven/resource_vpc_peering_connection.go
mwfrojdman: The AWS provider has an [accepter resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_peering_connection_accepter). I'm not sure how it works, but ideally the user could do something like:
1. Create the vpc peering connection (pcx) (like now)
2. When the pcx goes into PENDING_PEER on AWS, the user could use the AWS accepter resource to accept the pcx request in their account
3. The Aiven pcx resource (or is another resource needed? my TF skills are not that great) would depend on on the accepter resource and this resource would be sure to have state active

Pseudo example of how to use the accepter (not sure the `my_aiven_pcx` fields are very correct)
```
resource "aws_vpc_peering_connection_accepter" "my_accepter" {
  vpc_peering_connection_id = my_aiven_pcx.aws_vpc_peering_connection_id
  auto_accept               = true
}
```

rikonen:
Latest version looks otherwise reasonable but I was wondering how is the user supposed to get rid of the VPC connection if it ends up being in one of the failed states? The Terraform resource doesn't get created so you cannot delete it by removing that so one should go to web console and delete from there? I'm not sure how the existing connections work exactly but they might prevent creating a new one as well so if you first put in invalid specification and get an error about that then would you have to go to web UI to delete the invalid peering connection before you can actually update your Terraform config and create a working one? That wouldn't be terribly good user experience.

ivan-savciuc:
Added: in case of an error (any state other than `ACTIVE` or `PENDING_PEER`) peering connection will be deleted.

rikonen (approved):

PR #368, opened 2020-11-18
fix kafka topic creation typo

ivan-savciuc:
Fixes incorrect handling of an already exists error during creation of the topic.

### DIFF
--- a/aiven/kafka_topic_create.go
+++ b/aiven/kafka_topic_create.go
@@ -38,7 +38,7 @@ func (w *KafkaTopicCreateWaiter) RefreshFunc() resource.StateRefreshFunc {
 				return nil, "CREATING", nil
 			}
 
-			if ok && !aiven.IsAlreadyExists(aivenError) {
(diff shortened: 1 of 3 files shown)


encima (commented):
Is there a test for this? Good that we fixed it but adding test cases alongside fixes would be good also

ivan-savciuc:
@encima good point, I will try to add a test later today, it will require to run some API calls to create a topic before TF is allowed to do that. I'm not sure if we have a granularity of such level available in the TF Testing framework.

encima:
ah, that may be a bigger task then. We could use the OpenAPI spec to create a mock API layer to handle the calls (could be useful in the operator as well then). If you agree then I will merge this and create a ticket to add that functionality (which you can assign to me to do or help out with)?

ivan-savciuc:
I found a way to validate that logic using `PreConfig` function, and all looks to be straight forward. I will soon add tests.

ivan-savciuc:
Added tests

encima (approved):

PR #381, opened 2020-11-20
Add GCP tenant for Aiven services 

ivan-savciuc:
Adde/Exnteded `variant` user configuration option with `gcp` option.


rikonen (changes requested):
Is this some upcoming thing? I don't see this being supported on server side.

PR #384, opened 2020-11-23
add new project attributes

ivan-savciuc:


aiven/resource_project.go
```diff
@@ -60,6 +67,42 @@ var aivenProjectSchema = map[string]*schema.Schema{
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
 	},
+	"cloud": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Target cloud",
```
rikonen, line 73: `Default cloud for new services` would be more descriptive and perhaps the field name could be set to `default_cloud`.

aiven/resource_project.go
```diff
@@ -60,6 +67,42 @@ var aivenProjectSchema = map[string]*schema.Schema{
...
+		Description: "Estimated balance",
+	},
+	"payment_method": {
+		Type:        schema.TypeString,
+		Computed:    true,
+		Description: "Payment method",
+	},
+	"vat_id": {
+		Type:        schema.TypeString,
+		Computed:    true,
```
rikonen, line 103: This isn't computed but optional property the client can set.

rikonen (changes requested):

aiven/resource_project.go
ivan-savciuc: I will update go-client to include `vat_id` to create/update operations.

rikonen (approved):

PR #390, opened 2020-12-02
add project resource diff suppress logic for optional fields

ivan-savciuc:


aiven/resource_project.go
```diff
@@ -52,36 +57,41 @@ var aivenProjectSchema = map[string]*schema.Schema{
...
 	"available_credits": {
-		Type:        schema.TypeString,
-		Computed:    true,
-		Description: "Available credits",
-		Optional:    true,
+		Type:             schema.TypeString,
+		Computed:         true,
+		Description:      "Available credits",
+		Optional:         true,
+		DiffSuppressFunc: emptyObjectDiffSuppressFunc,
```
rikonen, line 94: This shouldn't be client settable in the first place

aiven/resource_project.go
```diff
@@ -12,21 +12,24 @@ import (
...
 	"billing_extra_text": {
-		Type:         schema.TypeString,
-		Description:  "Extra text to be included in all project invoices, e.g. purchase order or cost center number",
-		ValidateFunc: validation.StringLenBetween(0, 1000),
-		Optional:     true,
+		Type:             schema.TypeString,
+		Description:      "Extra text to be included in all project invoices, e.g. purchase order or cost center number",
+		ValidateFunc:     validation.StringLenBetween(0, 1000),
+		Optional:         true,
+		DiffSuppressFunc: emptyObjectDiffSuppressFunc,
```
rikonen, line 32: This sounds like something that should be possible to set to empty value. Possibly the same is true for some other fields as well. Won't this now prevent that?

ivan-savciuc: Yes, it is possible to set it to an empty string or ignore this field, for Terraform client it is all the same. The idea of adding this `DiffSuppressFunc` is to avoid situations when the user created a project using `copy_from_project`, which can populate some non zero value to this field and others. And we want to omit all these changes from a plan in case field is not set by the user to a particular value.

rikonen: Just to clarify: does this now work so that if the value is not present at all it will not be updated but if the value is present but empty it will be updated?

ivan-savciuc: Ah ok, now I get your point, in our case if a value is present, but empty TF client will ignore this change because `emptyObjectDiffSuppressFunc` will make it do so. We can have a different `DiffSuppressFunc` which will be a little bit smarter about that and will comply with the statement from above (`emptyObjectNoChangeDiffSuppressFunc`). 

Still, I believe in this case we would need to update go-client code because https://github.com/aiven/aiven-go-client/blob/master/project.go#L50 `UpdateProjectRequest ` has string pointers only for these fields: `billing_extra_text`, `billing_address` and `country_code` and the rest of optional fields will ignore by the go-client even if Terraform sets and empty string to the request.

WDYT, should we update a go-client or fields like `account_id`, `billing_currency`, `vat_id` should not have empty values?

rikonen: `billing_currency` cannot be empty. `vat_id` could be clearing it out is very rare. Probably same for `account_id`.

aiven/resource_project.go
```diff
@@ -178,14 +188,15 @@ func resourceProjectUpdate(d *schema.ResourceData, m interface{}) error {
...
+	billingExtraText := d.Get("billing_extra_text").(string)
 	billingAddress := d.Get("billing_address").(string)
 	countryCode := d.Get("country_code").(string)
 	project, err := client.Projects.Update(
 		d.Get("project").(string),
 		aiven.UpdateProjectRequest{
 			BillingAddress:   &billingAddress,
 			BillingEmails:    contactEmailListForAPI(d, "billing_emails", false),
-			BillingExtraText: optionalStringPointer(d, "billing_extra_text"),
+			BillingExtraText: &billingExtraText,
```
rikonen, line 199: Why this change? Doesn't this just cause the value to be always updated instead of omitting when it's not set on Terraform side?

aiven/resource_project.go
ivan-savciuc: Thanks, to clarify the situation completely, what about these fields:
- `card_id`
- `default_cloud`

All can be empty right?

And `copy_from_project` can only be set during the creation of resource?

rikonen: Yes to all

aiven/resource_project.go
ivan-savciuc: `optionalStringPointer` is converting `""` empty string to `nil`, but we need to keep it as empty string pointer. Anyways updated this logic by using a new helper function.

aiven/resource_project.go
```diff
@@ -160,12 +171,8 @@ func resourceProjectRead(d *schema.ResourceData, m interface{}) error {
 		return err
 	}
 
-	// Don't set card id unconditionally to prevent converting short card id format to long
-	currentCardID, err := getLongCardID(client, d.Get("card_id").(string))
-	if err != nil || currentCardID != project.Card.CardID {
-		if err := d.Set("card_id", project.Card.CardID); err != nil {
-			return err
-		}
+	if err := d.Set("card_id", project.Card.CardID); err != nil {
```
rikonen, line 174: Was there a reason for changing this logic? Nothing has changed related to card id handling and need for the special handling still remains afaict.

rikonen (changes requested):
Fails tests

aiven/provider.go
```diff
@@ -293,6 +305,20 @@ func emptyObjectDiffSuppressFunc(k, old, new string, _ *schema.ResourceData) boo
...
 }
 
+// emptyObjectNoChangeDiffSuppressFunc it suppresses a diff if a field is empty but have not
+// been set before to any value
+func emptyObjectNoChangeDiffSuppressFunc(k, old, new string, d *schema.ResourceData) bool {
+	if d.HasChange(k) {
+		return false
+	}
+
+	if new == "" && old != "" {
```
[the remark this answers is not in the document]
ivan-savciuc: Indeed we can optimise it.

aiven/resource_project.go
ivan-savciuc: Yep, this is something that I wanted to discuss here. It is a little bit misleading, comment and code itself. We claim we are trying to prevent converting short card id format to long, which is a byproduct of what we are actually doing in the code. We are in fact, only setting `card_id` if long card id has changed or in case of an error (I think we should not set anything in case of an error). I will try to change it a little bit...

aiven/resource_project.go
```diff
@@ -36,46 +42,51 @@ var aivenProjectSchema = map[string]*schema.Schema{
...
-		DiffSuppressFunc: createOnlyDiffSuppressFunc,
+		Description:      "Account ID",
+		DiffSuppressFunc: emptyObjectDiffSuppressFunc,
 	},
-	"country_code": {
+	"copy_from_project": {
 		Type:        schema.TypeString,
 		Optional:    true,
-		Description: "Billing country code of the project",
+		Description: "Copy properties from another project. Only has effect when a new project is created.",
```
rikonen, line 59: `createOnlyDiffSuppressFunc` was dropped from here. Is that intentional?

ivan-savciuc: ups, sorry, wasn't intentional, thx

rikonen (approved):

PR #393, opened 2020-12-03
add service user new password support

ivan-savciuc:


rikonen:
Should we rather use the existing password field?

ivan-savciuc:
> Should we rather use the existing password field?

Good idea, just did that!

aiven/resource_service_user.go
```diff
@@ -68,6 +68,8 @@ var aivenServiceUserSchema = map[string]*schema.Schema{
 		Type:        schema.TypeString,
 		Sensitive:   true,
 		Computed:    true,
+		Optional:    true,
+		ForceNew:    true,
```
rikonen, line 72: Why is `ForceNew` set to `true`? Code seems to update existing user so shouldn't be necessary. This would be problematic for Kafka users who have both certificate and password and this would cause the certificate to be changed as well.

rikonen (changes requested):

aiven/resource_service_user.go
ivan-savciuc: We do not have the `update` method for `service_user` resource, and all fields should be either `computed` or `ForceNew`. 

`create` operation does two API calls: `POST` to create a user and then `PUT` to update the password. Still, both of them are happening sort of in one transaction, and subsequent `read` operations should get a consistent certificate`s output.

rikonen: Ah, didn't read where the code actually was. Still, shouldn't we rather add the update method and make anything that cannot be updated have `ForceNew: true` instead of the one thing that actually does support update causing new user to be created?

ivan-savciuc: Sure, let's do that.

aiven/resource_service_user.go
```diff
@@ -120,11 +131,39 @@ func resourceServiceUserCreate(d *schema.ResourceData, m interface{}) error {
 		return err
 	}
 
+	if newPassword, ok := d.GetOk("password"); ok {
+		_, err := client.ServiceUsers.Update(projectName, serviceName, username,
```
rikonen, line 135: I was going to say this should just be passed in the create call but indeed create does not allow setting password, which is quite daft.

rikonen (approved):

PR #394, opened 2020-12-04
tests: add card_id to project resource

NARSimoes:
When we run this test `TestAccAivenProjectDataSource_basic` we have the following failure:

`--- FAIL: TestAccAivenProjectDataSource_basic (3.33s)
    datasource_project_test.go:14: Step 1/1 error: Check failed: Check 3/3 error: data.aiven_project.project: Attribute "card_id" is "", but "card_id" is not set in aiven_project.foo`

This fix the test but, I'm wondering if it's normal: the `card_id` is  `""`. Should this attribute (card_id) have some value?

### DIFF
--- a/aiven/resource_project_test.go
+++ b/aiven/resource_project_test.go
@@ -138,6 +138,7 @@ func testAccProjectResource(name string) string {
 			default_cloud = "aws-eu-west-2"
 			billing_currency = "EUR"
 			vat_id = "123"
+			card_id = ""
 		}
 
 		data "aiven_project" "project" {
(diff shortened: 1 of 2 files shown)


aiven/resource_project_test.go
```diff
@@ -138,6 +138,7 @@ func testAccProjectResource(name string) string {
 			default_cloud = "aws-eu-west-2"
 			billing_currency = "EUR"
 			vat_id = "123"
+			card_id = ""
```
ivan-savciuc, line 141: I reckon we should just remove `resource.TestCheckResourceAttrPair(datasourceName, "card_id", resourceName, "card_id"),` from the checks, since it is not really checking anything, perhaps we can add more meaningful check-ins instead validating other attributes.

ivan-savciuc (changes requested):

aiven/resource_project_test.go
NARSimoes: Removed that line from the checks. I agree with the meaningful checks and maybe we can do that in a separate PR.

ivan-savciuc (approved):

PR #397, opened 2020-12-08
replace deprecated functions after TF SDK v2 upgrade #1

ivan-savciuc:

PR #398, opened 2020-12-08
replace deprecated functions after TF SDK v2 upgrade #2

ivan-savciuc:


encima:
This change is for all calls under `datasource` and `resource` is that right? Tests are unaffected and continue to use the `fmt` module?

ivan-savciuc:
> This change is for all calls under `datasource` and `resource` is that right? Tests are unaffected and continue to use the `fmt` module?

Yep, this is only for `datasource` and `resource`, tests are unaffected.

encima (changes requested):
I would propose to close this as #397 is a related change which we would want in as a single entity in case we need to roll back or just for a clean log. What do you think?

ivan-savciuc:
Ok I will merge these two PRs into one.

encima:
> Ok I will merge these two PRs into one.

Already done #397, we can just close this

PR #399, opened 2020-12-09
Update the template with the right maximum values

ebeta-aiven:
https://app.clubhouse.io/aiven/story/25037/incorrect-maximum-value-for-some-kafka-properties


ivan-savciuc (dismissed):
Please also run `go generate` to update Golang cache files that we use in runtime and `go fmt ./...` to fix formating (go generate is often not compliment with fmt standards). Otherwise LGTM.

ivan-savciuc (approved):

PR #411, opened 2020-12-24
add support for new user configuration options

ivan-savciuc:


aiven/templates/service_user_config_schema.go
```diff
@@ -318,7 +329,7 @@ var (
           "properties": {
             "max_index_count": {
               "example": 3,
-              "maximum": 9223372036854775807,
+              "maximum": 9223372036854776000,
```
rikonen, line 332: I wonder why all these maximum values got changed. The API definition is unchanged and `9223372036854775807` is the correct value for all these.

ivan-savciuc: Hm... that is strange, they got changed automatically after running `generate_templates.sh`. And indeed after checking  API response `avn service types -v --json | grep 'maximum":'` correct values are 9`223372036854775807`.

ivan-savciuc: It looks like `jq` expression from  `generate_templates.sh` script is changing these values

rikonen (approved):

PR #415, opened 2021-01-08
add pg service upgrade check task

ivan-savciuc:


rikonen (changes requested):
Conflicts, otherwise looks good. I think the thing that comes up next is the fact that upgrade check will always return a failure if the service has pending maintenance updates and applying those is not possible via Terraform at the moment. But perhaps it is reasonable workflow to go and apply those from web UI first and then proceed with the service update via Terraform.

rikonen (approved):

PR #334, opened 2020-10-16
add support for new kafka topic features

ivan-savciuc:
- use new Kafka Topic config
- deprecate old fields
- change Kafka Topic cache, since list API method does not retrieve `config`
- update docs

PR #417, opened 2021-01-11
Load topic config on demand

juniorz:
`terraform plan` started to fail with `error waiting for Aiven Kafka topic to be ACTIVE: couldn't find resource (21 retries)`.

We have about Kafka 400 topics in a single project, and while reviewing the code we noticed a few issues introduced by #334:

1. Provider makes N+1 requests to the Aiven API while holding the semaphore.
2. Provider is not resilient to failures on the Aiven API and abandons the remaining unprocessed topics from the list on the first error (which will eventually cascade and lead to other N+1 requests).

This PR addresses both issues by:

1. Using `KafkaTopicAvailabilityWaiter.refresh()` to only refresh the list of known topics.
2. Returning a `resource.StateRefreshFunc` that only fetches the config for the topic that was requested.

This way we avoid `N-n` requests to the Aiven API, where `N` is the total topics in the project and `n` is how many topics present in this Terraform configuration.

### DIFF
--- a/aiven/kafka_topic_availability.go
+++ b/aiven/kafka_topic_availability.go
@@ -2,13 +2,13 @@ package aiven
 
 import (
 	"fmt"
-	"github.com/aiven/terraform-provider-aiven/pkg/cache"
-	"golang.org/x/sync/semaphore"
 	"log"
 	"time"
 
 	"github.com/aiven/aiven-go-client"
+	"github.com/aiven/terraform-provider-aiven/pkg/cache"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"golang.org/x/sync/semaphore"
 )
 
 // KafkaTopicAvailabilityWaiter is used to refresh the Aiven Kafka Topic endpoints when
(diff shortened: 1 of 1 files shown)


juniorz:
cc/ @ivan-savciuc @rikonen

ivan-savciuc:
Hi @juniorz, thanks for this PR and the description of a problem. 

As you can see Aiven API Kafka Topic v1 endpoints require multiple calls get a Topic with the config, we have a list endpoint which retrieves as many topics as you have for a service (400+ in your case). We have to iterate over each of them individually to get a config by calling a separate endpoint.   

You propose to move the code which sets the Topic config to the `RefreshFunc`, which will definitely reduce semaphore holding time and in your case (since not all topics are in Terraform) improve overall performance. 

The downside of this solution is that we no longer have topic config in the local in-memory cache and of course for the customers that do have all topics in the TF state no performance gain will be achieved. But the main problem is that we no longer have it in the cache. 

We've been thinking some time ago what can be done to make it faster, one idea was to fan-out the setting of the Topic config using wait groups, and it will stress Aiven API since multiple requests will be running simultaneously. I'm talking about something like that:
```go
var wg sync.WaitGroup

....

// we can limit the number of running goroutines

wg.Add(1)
go func() error {
	defer wg.Done()

	// when topic from a topics list is ACTIVE but has an empty config
	if topic.State == "ACTIVE" && topic.Config.CleanupPolicy.Value == "" {
		topic, err = w.Client.KafkaTopics.Get(w.Project, w.ServiceName, item.TopicName)
		if err != nil {
			return err
		}
	}
}()
wg.Wait()                
```

But it will further stress our API (the whole idea of having cache is not to stress it) and will make code more complicated. 

The ultimate solution for this issue is to use API v2, which retrieves back only topics you need, supports pagination and includes a Topic config straight away. We have an internal ticket for that and will be prioritised.

To conclude this, I would recommend trying fan-out from above or other ways to have Topic config in the in-memory cache. Or we replace this with v2 API endpoints which will no longer have `N+n` performance issues.

juniorz:
Hi @ivan-savciuc. The topic config is also added to cache on demand, right after the config is obtained:

https://github.com/aiven/terraform-provider-aiven/pull/417/files#diff-3bdbe561e3cd47eb3b86328c230f546a90677851af94a284944f441e15b4f0b4R79

So the cache will have a list of all topics in the project but will only have the config for those who terraform cares about. Why is not this enough for terraform?

juniorz:
I want to highlight that the focus of this PR is not merely to improve performance to reduce build time. Is to fix a performance issue that makes the provider unusable for projects with a large number of topics.

The build time will still linearly increase with the number of topics managed by terraform, which is not worse than the current behavior or linearly increasing with the number of topics in the project.

ivan-savciuc:
> Hi @ivan-savciuc. The topic config is also added to cache on demand, right after the config is obtained:
> 
> https://github.com/aiven/terraform-provider-aiven/pull/417/files#diff-3bdbe561e3cd47eb3b86328c230f546a90677851af94a284944f441e15b4f0b4R79
> 
> So the cache will have a list of all topics in the project but will only have the config for those who terraform cares about. Why is not this enough for terraform?

Ah indeed, initially I missed this line, but it looks like cache can be overwritten by a `w.refresh()` function if we leave it there. An example lets us have two Kafka Topics A and B, both created simultaneously, but Topic A was created first, and B is a bit lagging. In this case, the first run of `RefreshFunc` will create a cache for Topic A with the config. While waiting for Topic B to be available `w.refresh()` will be continuously triggered, it will update the cache with the Topic A but without config.

ivan-savciuc:
While thinking about a possible solution for this issue, I got this idea. What if we completely reduce the `refresh` function to refresh cache only for an individual topic on-demand? The initial idea to pre-warm cache on the first run does not work here since we have to fire a separate API call to get a Topic config, perhaps before we use v2 endpoint we can do with something like:

```go
func (w *KafkaTopicAvailabilityWaiter) refresh() error {
	if !kafkaTopicAvailabilitySem.TryAcquire(1) {
		log.Printf("[TRACE] Kafka Topic Availability cache refresh already in progress ...")
		return nil
	}
	defer kafkaTopicAvailabilitySem.Release(1)

	topic, err := w.Client.KafkaTopics.Get(w.Project, w.ServiceName, w.TopicName)
	if err != nil {
		return err
	}

	log.Printf("[TRACE] got a topic `%s` from aiven API with the status `%s`", topic.TopicName, topic.State)
	cache.GetTopicCache().StoreByProjectAndServiceName(w.Project, w.ServiceName, []*aiven.KafkaTopic{topic})
	return nil
}
```

@juniorz sorry for hijacking with some new ideas here, just trying to find the best solution to this issue. Would you be open to trying something like that?

juniorz:
@ivan-savciuc no problem! I have totally overlooked this aspect!

I agree with the approach you suggest. I was just not sure if there was any other part of the provider that relied on using the cache for existence checks, and decided to keep it there.

juniorz:
@ivan-savciuc I have validated your proposed changes in my environment, and it fixes the timeout as expected. I amended this PR with the changes for your convenience. Let me know about the plans to release this fix, as it is blocking our development at this moment.

Thank you!

ivan-savciuc (approved):

ivan-savciuc:
@juniorz thank you, we will release it ASAP.

rikonen (approved):

PR #423, opened 2021-01-20
add support for renaming projects

ivan-savciuc:


rikonen (approved):
Worth noting that renaming project is only possible if the project currently has no powered on services.

PR #424, opened 2021-01-20
kafka topic availability improvements

ivan-savciuc:
Some customers with 450+ topics per Kafka service report `error: error waiting for Aiven Kafka topic to be ACTIVE: couldn't find resource (21 retries)`, even after upgrading to `2.1.4`, after analyzing log files I found that many TF goroutines are waiting for the semaphore to unlock the refresh cache function, lots of messages like `Kafka Topic Availability cache refresh already in progress ...`.

Also, I can reproduce that on my box with 500 topics all present in the state file. Terraform is waiting for all of them to be created and accessible. Sometimes, it takes quite some time before TF waiters internal limit accedes. Also, we only allow one request to API at the time, which makes things drastically slower.

So I have added the following improvements:
- Let semaphore run 5 goroutines to refresh the cache simultaneously.
- Increase default waiting time (though the customer might easily update it)
- Increase interval for querying Kafka Topic availability 
- And increase the number of attempts to refresh Topic availability.

In my tests was able to create a Kafka service with 750 Topics per service without any issues, if we go beyond this (talking about 1k+ topics) waiting time is so significant that we even this version is failing with `error: error waiting for Aiven Kafka topic to be ACTIVE: couldn't find resource (51 retries)`, but I think to support such massive amount of topics we would need to use `/v2/project/<project>/service/<service_name>/topic` endpoint and change waiter logic completely.

### DIFF
--- a/aiven/kafka_topic_availability.go
+++ b/aiven/kafka_topic_availability.go
@@ -71,7 +71,7 @@ func (w *KafkaTopicAvailabilityWaiter) RefreshFunc() resource.StateRefreshFunc {
 }
 
 func (w *KafkaTopicAvailabilityWaiter) refresh() error {
-	if !kafkaTopicAvailabilitySem.TryAcquire(1) {
+	if !kafkaTopicAvailabilitySem.TryAcquire(5) {
 		log.Printf("[TRACE] Kafka Topic Availability cache refresh already in progress ...")
 		return nil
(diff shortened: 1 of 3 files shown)


rikonen:
Description telling why these changes were done would be nice.

rikonen (approved):

PR #430, opened 2021-01-27
kafka topic v2

ivan-savciuc:
We are using v2 Kafka Topic API endpoint as a default way of getting Kafka Topics, and if it is not available for a particular Kafka service, we do have a fallback of v1 Endpoint.

Also logic is a bit different, now the first TF goroutine from the batch will warm up the cache by adding topic names to cache queue by running v1 API endpoint get all Kafka topics for a service. All topic names from the cache queue will be bundled in batches up to 100 topics at the time, and v2 API endpoint will be queried until we get data for all topics in the cache.  

Please note that v1 fallback logic was not tested, since it is hard to find any old Kafka nodes that do not support  v2.

### DIFF
--- a/aiven/kafka_topic_availability.go
+++ b/aiven/kafka_topic_availability.go
@@ -3,6 +3,7 @@ package aiven
 import (
 	"fmt"
 	"log"
+	"sync"
 	"time"
 
 	"github.com/aiven/aiven-go-client"
@@ -21,6 +22,7 @@ type KafkaTopicAvailabilityWaiter struct {
 }
 
 var kafkaTopicAvailabilitySem = semaphore.NewWeighted(1)
+var warmingUpCacheOne sync.Once
 
 // RefreshFunc will call the Aiven client and refresh it's state.
 func (w *KafkaTopicAvailabilityWaiter) RefreshFunc() resource.StateRefreshFunc {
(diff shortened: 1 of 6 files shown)


aiven/kafka_topic_availability.go
```diff
@@ -71,19 +73,57 @@ func (w *KafkaTopicAvailabilityWaiter) RefreshFunc() resource.StateRefreshFunc {
...
+
+	for {
+		queue := c.GetQueue(w.Project, w.ServiceName)
+
+		if len(queue) == 0 {
+			break
+		}
+
+		log.Printf("[DEBUG] Kafka Topic queue: %+v", queue)
+		v2Topics, err := w.Client.KafkaTopics.V2List(w.Project, w.ServiceName, queue)
```
rikonen, line 107: Probably makes sense to do in batches with at most, say, 200 items fetched in one go. There will likely be problems if you try getting 10k or however many topics some environment might have in one go.

rikonen: Sorry, didn't read far enough. Looks like it already does this.

rikonen:
Did you have any numbers how this performs?

ivan-savciuc:
> Did you have any numbers how this performs?

Not really, I have tested it multiple times usually with 450-550 topics for a service, and it only takes some time (up to 10-20 secconds), in the beginning, to warm up the cache and the rest of the TF plan is executed almost instantly. I will run some tests now with the timer and will provide result here.

ivan-savciuc:
**Results for Kafka service with 550 topics**

This version:
```
terraform refresh  
real    0m24.574s
user    0m7.939s
sys     0m2.232s

```

The current master version (with errors, impossible to refresh some of the topics):
```
terraform refresh  
real    5m6.770s
user    0m8.739s
sys     0m3.169s

```

rikonen:
OK, sounds good. Just one more question: previously we did not fetch all topics upfront. With these changes we do even if the cluster does not support the v2 API. Is that potentially going to cause some problems?

ivan-savciuc:
In the version before Kafka Topic Config implementation, we had a stable code fetching all the available Kafka Topics and working mainly with the local in-memory cache. After we added Topic Config, we had to do N (number of topics in a service) requests to API to get all necessary information. All problems started to appear for the customers with massive Kafka services. And in the last 3 iterations, we were giggling with these N requests that we have to perform such that it causes fewer problems. Still, the truth is that some customers are having problems with all these 3 recent versions, some have improvements with a particular version, and for some things get worth with the new updates.

So what I'm trying to say, for customers with old Kafka nodes (when v2 not supported) problems won't disappear. Hopefully, we can run some maintenance upgrade to existing old Kafka clusters or do that by customer request. For thought who will be stuck with v1 API, things will be working in the same way with version `v2.1.3`, which is in my opinion, the less problematic implementation of Kafka Topic N requests.

rikonen (approved):

PR #446, opened 2021-02-25
isolating Azure fields settings for VPC peering connection

ivan-savciuc:
We use a VPC peering connection codebase for Azure, AWS, GCP, and AWS Transit Gateway Attachment. Depending on the situation, we silence errors when Terraform attempts to set a value to a field that does not exist. It all works fine when a client setting TF_SCHEMA_PANIC_ON_ERROR is turned off, but we see that it is not the case for some customers. When panic-on-error is on, Terraform SDK throws a panic on the attempt to set a none existing VPC peering connection field and interrupt the plan's execution.

Ideally, we would like to have a separate resource for each type of VPC peering connection, but this patch should resolve issues for customers with TF_SCHEMA_PANIC_ON_ERROR=on.


rikonen (approved):
I guess this change could be problematic if user_peer_network_cidrs goes from non-empty list to empty list in which case state doesn't get updated but maybe that's better bug to have than panic.

PR #464, opened 2021-03-19
remove kafka topic 404 error silencing

ivan-savciuc:


rikonen:
Won't this cause errors for topic creation?

ivan-savciuc:
> Won't this cause errors for topic creation?



> Won't this cause errors for topic creation?

Nope. I have tested this. Probably because after creation we do not read newly created topics.

rikonen:
Yes looks like in b50537c1 the code was changed so that it does not read topic after it has been created. I think that change combined with this could be problematic as there is good chance the topic will not be returned by the API right after creation. Could we update resourceKafkaTopicCreate so that it waits a moment for the API to start returning topic before considering creation to be complete?

ivan-savciuc:
`ContinuousTargetOccurence: 3` was added to the Kafka Topic create waiter, so 3 times API should send back 409 "Topic ... already exists" before creation can be considered successful.

rikonen:
That does not necessarily do quite the same thing as the topic listing API has server side caching (which should be invalidated upon creation of new topic but might still play some role here) and I don't think there's any guarantee that getting error for topic creation guarantees the topic can be listed; the broker handling the request might have recorded the request to create the topic but it may not yet have been propagated to other brokers. Just doing `GET /v1/project/<project>/service/<service>/topic/<topic>` and treating the operation as success when that returns 200 would probably be the way to go.

ivan-savciuc:
Ok, then I'll add `GET /v1/project/<project>/service/<service>/topic/<topic>` HTTP status checking to the create waiter.

ivan-savciuc:
@rikonen: a quick question, do you think we need to wait until the topic state is ACTIVE in the creation waiter, or we check that response is OK 200 regardless of the topic state?

ivan-savciuc:
I've added `GET /v1/project/<project>/service/<service>/topic/<topic>` checking logic, and it takes almost forever if your TF plan has hundreds of topics. I do not think we can leave it like that, and perhaps we can find another means to validate whether the topic was successfully created or change an error handling logic for the V2 endpoint to send 404 only when the topic does not exists.

rikonen:
I can check what's the status of this API being available in currently deployed services. It is possibly already available for all services. But also what is actually the slow part here? Just creating a topic, then waiting for x seconds and checking that this individual topic is found does not sound especially problematic in itself.

rikonen:
Actually what's the problem with 404 responses? The "service too old" response is 409.

rikonen:
There are still some services that lack the new topic APIs. All of those should get updated at latest on 2021-04-23

ivan-savciuc:
I think this is OK that not all services have a new topic API.

To recap everything, we silence the 404 error in availability waiter because of the flaky API that could report 404 Not Found even after the topic creation. The side effect is that we wait forever for a topic that was deleted from a Kafka service outside of Terraform; instead of waiting, we would like to report an error. So this is where it all started.

As you suggested, we added `GET /v1/project/<project>/service/<service>/topic/<topic>` 200 OK validation to the creation waiter, and what is happening is after creating a topic `GET /v1/project/<project>/service/<service>/topic/<topic>`  retrieving error: 
```
{"errors":[{"message":"Failed to fetch Kafka topic: topic-s1-40","status":501}
```
For the first couple of newly created topics, it can be just one or two attempts before we get `200 OK`, but further we create new topics, it takes more and more attempts. We have an only a 1-minute time limit for a topic to be created, and in some of my tests, even with 100 topics in the plan, I'm getting ```timeout while waiting for the state to become 'CREATED' (last state: 'CONFIGURING', timeout: 1m0s)```

rikonen (approved):

PR #465, opened 2021-03-22
change kafka topic resource read logic for deprecated fields

ivan-savciuc:


aiven/resource_kafka_topic.go
```diff
@@ -375,7 +375,7 @@ func resourceKafkaTopicRead(ctx context.Context, d *schema.ResourceData, m inter
 	if _, ok := d.GetOk("retention_hours"); topic.RetentionHours != nil && ok {
 		// it could be -1, which means infinite retention
 		if *topic.RetentionHours >= -1 {
-			if err := d.Set("retention_hours", *topic.RetentionHours); err != nil {
+			if err := d.Set("retention_hours", topic.Config.RetentionMs.Value); err != nil {
```
rikonen, line 378: Setting RetentionMs to retention_hours doesn't sound right and also the if condition above checks different variable

rikonen (changes requested):
Description in commit message would be nice. It's not obvious why these are changed

aiven/resource_kafka_topic.go
ivan-savciuc: New V2 Kafka API does not have these fields in the response: `minimum_in_sync_replicas`, `retention_bytes` and `retention_hours`, but has corresponding fields the Config section. Therefore we should use Config fields from the response to populate these deprecated fields for the Terraform resource.

rikonen (approved):

PR #474, opened 2021-04-01
add support for kafka topic tags

ivan-savciuc:


aiven/resource_kafka_topic.go
```diff
@@ -394,9 +431,25 @@ func resourceKafkaTopicRead(ctx context.Context, d *schema.ResourceData, m inter
...
 	}
 
+	if err := d.Set("tag", flattenKafkaTopicTags(topic.Tags)); err != nil {
+		return diag.Errorf("error setting Kafka Topic Tags for resource %s: %s", d.Id(), err)
+	}
+
 	return nil
 }
 
+func flattenKafkaTopicTags(list []aiven.KafkaTopicTag) []map[string]interface{} {
```
rikonen, line 441: "flatten" is confusing name for this as this converts a list of one type to a list of other type instead of doing some flattening. Something like `convertTagsToAPIFormat` would be more accurate.

ivan-savciuc: I agree. It sounds a little bit misleading since, as you mentioned, we convert one type of list no another. I took these naming ideas from [Terraform Tutorial](https://learn.hashicorp.com/tutorials/terraform/provider-complex-read?in=terraform/providers#implement-complex-read) and checking other Terraform providers. The idea is that we are "flattening" API response to a map that represents the TF schema.

If you as still thinking we should change that, I propose we do that for all of such functions in the codebase for consistency.

rikonen (approved):

PR #487, opened 2021-04-27
fix project technical_emails and billing_emails schema

ivan-savciuc:


rikonen:
Could you please add the reason why there was a problem with this into the commit message

ivan-savciuc:
> Could you please add the reason why there was a problem with this into the commit message

Added additional comment:
If DiffSuppressFunc is true, a default diff for the type of the schema will be ignored. And for the TypeSet diff will append elements to the list.

ivan-savciuc:
Please let me know if that sounds reasonable, if not I will add more information to the commit message.

rikonen (approved):

PR #489, opened 2021-04-30
add import id format to the docs for where it is missing

ivan-savciuc:


rikonen (approved):
It could also be helpful to give an example of what a real-life id would look like, the format description may not be completely evident as it is (my bad as I added those initially).

PR #491, opened 2021-05-06
resend account team member invitation if deleted or expired

ivan-savciuc:


rikonen:
Code seems fine but commit message is missing description of why this change is done

ivan-savciuc:
Added extra comment to the commit message:  If an invitation was sent but not accepted by the user for some times, it expires, and on the next terraform apply, there is an error saying that the invitation that is already in the state is not found. To avoid this, we are sending an invitation once more if it was deleted or expired (we cannot tell which is the case) but only when the invitation was accepted before.

rikonen (approved):

PR #498, opened 2021-05-17
Fix failing acceptance tests

jonatasbaldin:
Hey there!

I was getting errors while running the acceptance tests, so I fixed some:

1. [This error](https://github.com/hashicorp/terraform-plugin-sdk/issues/697) was happening, so I updated the `terraform-plugin-sdk` library.

2. The `aiven_connection_pool` resource, when used with the argument `database_name`, should use `depends_on` due to race condition errors (pool being created before the database). I fixed the docs. 

3. `TestAccAivenProject_accounts`, `TestAccAivenProjectUser_basic`, `TestAccAivenProjectVPC_basic`: these tests were failing with:
```
    testing_new.go:63: Error running post-test destroy, there may be dangling resources: 403: {"errors":[{"message":"Not a project member","status":403}],"message":"Not a project member"} - 
``` 
So I changed the `CheckDestroy` method to accept `403` as well.

4. Some M3DB tests were failing where the `m3db_version` and `m3aggregator_version` were expecting a string, not a float.

5. I also updated the docs to reflect the changes where necessary.

Let me know what you think!


ivan-savciuc (approved):
LGTM

PR #511, opened 2021-05-18
Add Prometheus integration example

hsorellana:
Add example of usage for a Prometheus integration and update available `endpoint_type` values in `aiven_service_integration_endpoint` resource


ivan-savciuc (approved):
Thanks, @hsorellana for contributing!

PR #513, opened 2021-05-19
Fix comment on resource usage

hsorellana:


ivan-savciuc (changes requested):
@hsorellana, thanks for contributing. 

The idea here is that we have two options on how to work with the `aiven_service_integration` resource. 

The first one is to use service names:
`source_service_name` and `destination_service_name` both accept service names that can be passed like `aiven_kafka.XXX.service_name` or `aiven_pg.YYY.service_name`. 

The second option is to use endpoint IDs, which can be done by referencing the service integration ID. something like `source_endpoint_id = aiven_service_integration_endpoint.XXX.id`.

And proposed changes suggest the opposite from above and may lead to confusion. Feel free to extend comments with the logic from above.

ivan-savciuc:
Closing this PR for inactivity, please reopen if necessary.

PR #514, opened 2021-05-20
Add format for docs to improve readability

hsorellana:


encima (approved):
Thanks for contributing! Agreed that it is clearer and easier to grep quickly. Docs change will not need tests to run but doing it to be safe. Approval from me, what do you think @ivan-savciuc ?

ivan-savciuc:
These changes indeed improve visibility. Thanks, @hsorellana for contributing. Though I have two issues with this:

- Consistency reasons, all our documentation is generated in one way, and if we update just this one for Elastic Search, it does not sound right to me. If we decide to reformat documentation, it should happen for all files. 

- User configuration options are hierarchically formatted to reflect the underlying resource structure, and flattening may lead to confusion since it is not straightforward, for example, that `elasticsearch` is embedded into `elasticsearch_user_config`. And once again, if we would decide to change that, we should do that for all files. 

So, I would rather close this PR, sorry about that, and I'm open to changes regarding formatting, but we should do that for all files.

PR #515, opened 2021-05-25
remove default values for user configuration options

ivan-savciuc:
It will prevent services from REBUILDING when updated to the new TF provider version, which has some new default values to be propagated.

### DIFF
--- a/aiven/user_config.go
+++ b/aiven/user_config.go
@@ -44,13 +44,11 @@ func generateTerraformUserConfigSchema(key string, definition map[string]interfa
 		diffFunction = emptyObjectDiffSuppressFunc
 	}
 
-	defaultValue := getAivenSchemaDefaultValue(definition)
 	title := definition["title"].(string)
 
 	switch valueType {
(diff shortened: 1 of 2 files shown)


rikonen:
The original intent of the change was to prevent differences between local state and remote state when the fields with defaults are not set on Terraform side. We should in general not be introducing new user config options that have a default value so I wonder which values have caused this to be a problem. With this change does apply show a change when you run it after creating a service whose user config has some defaults but those are not set in Terraform configuration?

ivan-savciuc:
@rikonen, I do not know which of the new user configuration options were recently added with default values that revived this issue. With this change, a user won't see anything in the plan if a user configuration option has a default value set on the Aiven side. We have some optional user config options that do not have a default value set in the schema but do receive some values on the Aiven side, and a user won't see that in the plan as a change(like `kafka_version`, `group_max_session_timeout_ms` or `log_retention_bytes `).

rikonen (approved):

PR #519, opened 2021-06-04
add database delete waiter

ivan-savciuc:
Sometimes deletion of database results in a 500 error

### DIFF
--- a/aiven/resource_database.go
+++ b/aiven/resource_database.go
@@ -6,7 +6,9 @@ import (
 	"context"
 	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


encima:
+1 for adding. I think adding a test makes sense and what is the expected behaviour on timeout? Do we recommend to run again or that the call timed out but DB will be deleted?

ivan-savciuc:
> +1 for adding. I think adding a test makes sense and what is the expected behaviour on timeout? Do we recommend to run again or that the call timed out but DB will be deleted?

I was looking into how we can test that, and the terraform testing framework that we are using does not allow intercepting the creation and deletion of a test plan. So, therefore, unfortunately, there is no good way of testing this particular use-case other than creating a database and deleting it via the web console and then running terraform destroy. The ideas of this waiter is to keep deleting a database until we either get a 200 OK response code from Aiven or 404 Not found.

encima (approved):
Sucks that TF does not have decent testing to handle deletes but that is a limitation. Moving to approve without a test when it is not possible

PR #524, opened 2021-06-24
add service integration endpoint_id validation

ivan-savciuc:


encima (approved):
Nice, tests worked locally (thanks for the ARM builds btw!) and regex tested outside for other use cases.

PR #525, opened 2021-06-30
add use_source_project_billing_group support for project resource

ivan-savciuc:


aiven/resource_project.go
```diff
@@ -190,14 +197,15 @@ func resourceProjectCreate(_ context.Context, d *schema.ResourceData, m interfac
 			AccountId:        optionalStringPointer(d, "account_id"),
 			BillingCurrency:  billingCurrency,
 			VatID:            vatID,
+			//UseSourceProjectBillingGroup: d.Get("use_source_project_billing_group").(bool),
```
rikonen, line 200: Why is this commented out?

rikonen (changes requested):

aiven/resource_project.go
ivan-savciuc: It shouldn't, and it is my leftover from some tests.

rikonen (approved):

PR #530, opened 2021-07-13
Small cleaning VPC peering log error

malonsocasas:
Had a confusing error message when trying to import a VPC peering in the terraform state

This is just a one line fix of this error message


ivan-savciuc (approved):
@malonsocasas thanks for contributing!

PR #534, opened 2021-07-23
change service integration behaviour

ivan-savciuc:
In some cases, service creation may trigger the creation of service integration. For example, PG read replica service integration does that. PG replica service is created with service integration user configuration options that trigger the creation of service integration. To handle such cases correctly, we update the creation logic for the service_integration resource; it should work correctly when a service integration already exists.

### DIFF
--- a/aiven/resource_service.go
+++ b/aiven/resource_service.go
@@ -135,10 +135,9 @@ func serviceCommonSchema() map[string]*schema.Schema {
 			Description: "Service state",
 		},
 		"service_integrations": {
-			Type:             schema.TypeList,
-			Optional:         true,
-			Description:      "Service integrations to specify when creating a service. Not applied after initial service creation",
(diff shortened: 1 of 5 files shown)


encima (dismissed):
Only tested by running tests existing. Check for adding integrations to already running services makes sense but would have liked to test. Approving now as I am holding up the release. Will merge by EOD unless more stringent checks are needed and then I will make time on Monday

ivan-savciuc:
Updated this branch with the latest master changer

mhoffm-aiven (changes requested):
purely cosmetical, but could we separate stdlib imports from external imports?

ivan-savciuc:
> purely cosmetical, but could we separate stdlib imports from external imports?

sure, will do, I'll try to configure my IDE so it can do that automatically

mhoffm-aiven (approved):

PR #537, opened 2021-08-05
DRAFT: use a waiter when deleting vpc peering connections

mhoffm-aiven:
`vpc_peering_connection`: wait for deletion

Previously `vpc_peering_connections` would be deleted asynchronously. There is a race condition when deleting and recreating the resource, where recreation fails because the previous deletion has not finished yet. This PR fixes the race by waiting for the `vpc_peering_connection` to be deleted synchronously.

### DIFF
--- a/aiven/resource_vpc_peering_connection.go
+++ b/aiven/resource_vpc_peering_connection.go
@@ -5,12 +5,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"log"
 	"strings"
 	"time"
 
 	"github.com/aiven/aiven-go-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -88,6 +87,7 @@ func resourceVPCPeeringConnection() *schema.Resource {
 		},
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(2 * time.Minute),
+			Delete: schema.DefaultTimeout(2 * time.Minute),
 		},
 
 		Schema: aivenVPCPeeringConnectionSchema,
@@ -108,8 +108,9 @@ func resourceVPCPeeringConnectionCreate(ctx context.Context, d *schema.ResourceD
 		return diag.Errorf("incorrect VPC ID, expected structure <PROJECT_NAME>/<VPC_ID>")
 	}
 
+	peerCloudAccount := d.Get("peer_cloud_account").(string)
+	peerVPC := d.Get("peer_vpc").(string)
 	peerRegion := d.Get("peer_region").(string)
-
 	if peerRegion != "" {
 		region = &peerRegion
 	}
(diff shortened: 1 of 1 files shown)


aiven/resource_vpc_peering_connection.go
```diff
@@ -286,33 +287,31 @@ func resourceVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData,
...
-			err := client.VPCPeeringConnections.DeleteVPCPeeringWithResourceGroup(
-				projectName, vpcID, peerCloudAccount, peerVPC, peerResourceGroup.(string), peerRegion)
-			if err != nil {
-				return diag.Errorf("Error deleting VPC peering connection with resource group: %s", err)
-			}
-			return nil
-		} else {
-			return diag.Errorf("cannot delete an Azure VPC peering connection without `peer_resource_group`")
-		}
+	var peerResourceGroupPtr *string
```
ivan-savciuc, line 295: Hi, VPC peering connection may be in one of these states:
- ACTIVE
- REJECTED_BY_PEER 
- PENDING_PEER
- INVALID_SPECIFICATION
- DELETING
- DELETED
- DELETED_BY_PEER
- APPROVED

When we initiate deletion, it should switch to the **DELETING** state, and after a while, should change to **DELETED**, and then be gone from the system.

Let's rework the deletion waiter slightly. First of all, I would recommend keeping the existing deletion logic, and in the deletion waiter, we can query the VPC peering connection. If we get back the `404 NotFound` error, we consider the status to be **DELETED**. Otherwise, we take what is coming from the API.

Something like that:
```golang
func (w *VPCPeeringConnectionDeleteWaiter) RefreshFunc() resource.StateRefreshFunc {
	return func() (interface{}, string, error) {
		pc, err := w.Client.VPCPeeringConnections.GetVPCPeering(w.Project, w.VPCID, w.PeerCloudAccount, w.PeerVPC, w.PeerRegion)

		if err != nil {
			if aiven.IsNotFound(err) {
				return nil, "DELETED", nil
			}
			return nil, "", err
		}

		return pc, pc.State, nil
	}
}
```

Moreover, I was checking whether we can get away with checking `peer_resource_group` and determining whether it is Azure or not. I think we should still use `isAzureVPCPeeringConnection` to determine that because the user by accident may set this optional field but still VPC will be created in AWS or GCP cloud, so we should be sure that we are deleting an Azure VPC.

ivan-savciuc (changes requested):

aiven/resource_vpc_peering_connection.go
```diff
@@ -286,33 +287,90 @@ func resourceVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData,
...
+			"APPROVED",
+			"APPROVED_PEER_REQUESTED",
+			"DELETING",
+			"INVALID_SPECIFICATION",
+			"PENDING_PEER",
+			"REJECTED_BY_PEER",
+		},
+		Target: []string{
+			"DELETED",
+			"DELETED_BY_PEER",
```
ivan-savciuc, line 338: I think `DELETED_BY_PEER` should be in the Pending states

ivan-savciuc: `DELETED_BY_PEER` means: The customer deleted the peering connection through their own cloud account. That is, the customer deleted the peering cloud resource in their account. Executor detects this when regularly rebuilding existing peering connections in the `PENDING_PEER` and `ACTIVE` states, and noticed that the connection cloud resource is in a deleted state or is missing and we infer the customer must have done something about it. There are no transitions from this state

aiven/resource_vpc_peering_connection.go
```diff
@@ -286,33 +287,90 @@ func resourceVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData,
...
+		projectName,
+		vpcID,
+		peerCloudAccount,
+		peerVPC,
+		peerRegion,
+	); err != nil && !aiven.IsNotFound(err) {
 		return diag.Errorf("Error deleting VPC peering connection: %s", err)
 	}
 
+	if _, err := (&resource.StateChangeConf{
```
ivan-savciuc, line 326: Just a minor thing: It is hard to read this huge `IF` statement. If we can split it, that would be cool. 

I mean something like:
```go
conf := &resource.StateChangeConf{...}
conf.WaitForStateContext ...
```

aiven/resource_vpc_peering_connection.go
```diff
@@ -286,33 +287,90 @@ func resourceVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData,
...
+			"REJECTED_BY_PEER",
+		},
+		Target: []string{
+			"DELETED",
+			"DELETED_BY_PEER",
+		},
+		Refresh: func() (interface{}, string, error) {
+			var (
+				pc       *aiven.VPCPeeringConnection
+				innerErr error
```
ivan-savciuc, line 343: Perhaps we can use `err` here as a variable name. It is easier to read when `err` is used everywhere where possible.  Perhaps something to consider?

ivan-savciuc (changes requested):
Good job, it looks great. I found one thing with the state that should be changed, also added some code styling suggestions.

ivan-savciuc:
Also, please squash git commits into 1-2 so that we can merge them.

aiven/resource_vpc_peering_connection.go
mhoffm-aiven: done

aiven/resource_vpc_peering_connection.go
mhoffm-aiven: done

aiven/resource_vpc_peering_connection.go
mhoffm-aiven: done

aiven/resource_vpc_peering_connection.go
```diff
@@ -132,41 +133,57 @@ func resourceVPCPeeringConnectionCreate(ctx context.Context, d *schema.ResourceD
...
 			PeerRegion:           region,
 			UserPeerNetworkCIDRs: cidrs,
 			PeerAzureAppId:       peerAzureAppID,
 			PeerAzureTenantId:    peerAzureTenantID,
 			PeerResourceGroup:    peerResourceGroup,
 		},
-	)
-	if err != nil {
-		return diag.Errorf("Error creating VPC peering connection: %s", err)
+	); err != nil && !aiven.IsAlreadyExists(err) {
```
ivan-savciuc, line 148: `!aiven.IsAlreadyExist` should not be here. When we attempt to create a resource, an expectation is that it does not exist upfront. The reason is that preexisting resources might have a different configuration from the one we want to create, but both might have the same ID.

aiven/resource_vpc_peering_connection.go
```diff
@@ -286,33 +303,90 @@ func resourceVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData,
...
+			"INVALID_SPECIFICATION",
+			"PENDING_PEER",
+			"REJECTED_BY_PEER",
+			"DELETED_BY_PEER",
+		},
+		Target: []string{
+			"DELETED",
+		},
+		Refresh: func() (interface{}, string, error) {
+			var (
```
ivan-savciuc, line 357: We can reduce the number of lines of code by removing `var ( ... )` with a single line `var pc *aiven.VPCPeeringConnection`, since we only have one variable.

ivan-savciuc (changes requested):

aiven/resource_vpc_peering_connection.go
mhoffm-aiven: done

aiven/resource_vpc_peering_connection.go
mhoffm-aiven: done

ivan-savciuc (approved):
Good job @mhoffm-aiven

PR #541, opened 2021-08-09
linters + change imports style

ivan-savciuc:
This PR updates golang imports appearance and updated GitHub workflow to validate that, also includes other checks as well

### DIFF
--- a/aiven/provider.go
+++ b/aiven/provider.go
@@ -5,13 +5,14 @@ package aiven
 import (
 	"context"
 	"fmt"
-	"github.com/aiven/terraform-provider-aiven/pkg/cache"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
 	"net/url"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/aiven/terraform-provider-aiven/pkg/cache"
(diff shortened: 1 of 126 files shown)


aiven/provider.go
```diff
@@ -5,14 +5,15 @@ package aiven
...
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"log"
 	"net/url"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/aiven/terraform-provider-aiven/pkg/cache"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+
```
mhoffm-aiven, line 16: this whitespace is weird

mhoffm-aiven (changes requested):
There are some files left where the external imports are not grouped properly ( see comment ). All in all loving it though.

aiven/provider.go
ivan-savciuc: I did `goimports  -w .` and as you can see `golangci-clint -E goimports` is also has no complaints. How different should this be from what we have?

mhoffm-aiven: i think goimports does not care about grouping of the imports; i think we should group all external imports together ( its only 3 files that need this post processing )

mhoffm-aiven (approved):
looks good to me

PR #545, opened 2021-08-16
Remove old version reference

derrix060:
I could just update the version reference, but it would be obsolete soon.

Removing the version here would force people to check the latest version :)


ivan-savciuc (approved):

ivan-savciuc:
@derrix060 thanks for contributing

PR #549, opened 2021-08-18
remove specific version from acceptance tests

ivan-savciuc:
It is not a good idea to use a specific service version for acceptance testing since Aiven services always evolve, and certain versions get old and eventually reach EOL, which will break tests.


mhoffm-aiven (approved):
LGTM, also works when running on my machine.

PR #555, opened 2021-09-01
RFC: split elasticsearch resource ACLs

mhoffm-aiven:
This PR aims to split `elasticsearch_acl` into `elasticsearch_acl_config` to configure acls  and `elasticsearch_acl_set` to create the acl rules. This way we can add acls in different resources and attempt to make our terraform configuration more modularizable.

### DIFF
--- a/aiven/datasource_elasticsearch_acl.go
+++ b/aiven/datasource_elasticsearch_acl.go
@@ -11,9 +11,9 @@ import (
 
 func datasourceElasticsearchACL() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: datasourceElasticsearchACLRead,
-		Schema: resourceSchemaAsDatasourceSchema(aivenElasticsearchACLSchema,
-			"project", "service_name"),
+		DeprecationMessage: "ElasticsearchACL is deprecated please use ElasticsearchACLConfig and ElasticsearchACLRule",
+		ReadContext:        datasourceElasticsearchACLRead,
+		Schema:             resourceSchemaAsDatasourceSchema(aivenElasticsearchACLSchema, "project", "service_name"),
 	}
 }
 
@@ -34,6 +34,5 @@ func datasourceElasticsearchACLRead(ctx context.Context, d *schema.ResourceData,
 		return resourceElasticsearchACLRead(ctx, d, m)
 	}
 
-	return diag.Errorf("elasticsearch acl %s/%s not found",
-		projectName, serviceName)
+	return diag.Errorf("elasticsearch acl %s/%s not found", projectName, serviceName)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ parent: README
 nav_order: 1
 ---# Changelog
 
+## [2.1.20] - not yet released
+- split `aiven_elasticsearch_acl` into `aiven_elasticsearch_acl_config` and `aiven_elasticsearch_acl_rule`
+- deprecated `aiven_elasticsearch_acl`
+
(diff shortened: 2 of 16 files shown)


aiven/datasource_elasticsearch_acl.go
```diff
@@ -11,9 +11,9 @@ import (
 
 func datasourceElasticsearchACL() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: datasourceElasticsearchACLRead,
-		Schema: resourceSchemaAsDatasourceSchema(aivenElasticsearchACLSchema,
-			"project", "service_name"),
+		DeprecationMessage: "ElasticsearchACL is deprectated as it is not easily modularizable, please use ElasticsearchACLConfig and ElasticsearchACLSet",
```
ivan-savciuc, line 14: Perhaps we don't have to provide such a long deprecation message, for my money `ElasticsearchACL is deprecated please use ElasticsearchACLConfig and ElasticsearchACLSet` should do. And btw there is a small typo in word `deprectated`

aiven/internal/mutexkv/mutexkv.go
```diff
@@ -0,0 +1,51 @@
+package mutexkv
```
ivan-savciuc, line 1: we already gave a `pkg` package, perhaps it makes sense to move this file over there

ivan-savciuc: Also, I was thinking that this whole concurrency thing can be simplified with something like this:
```go
var esACLSem = semaphore.NewWeighted(1)

func resourceElasticsearchACLUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
	if !esACLSem.TryAcquire(1) {
		log.Printf("[TRACE] ES ACL update is busy by another goroutine ...")
		time.Sleep(2 * time.Second)
		return resourceElasticsearchACLUpdate(ctx, d, m)
	}
	defer kafkaTopicAvailabilitySem.Release(1)
...
}
```

ivan-savciuc (commented):
I did the first readthrough and some testing. I will do more soon.

aiven/internal/mutexkv/mutexkv.go
mhoffm-aiven: Is that simpler though? I just copied straight from the terraform sdk doc's I.have to admit

aiven/datasource_elasticsearch_acl.go
mhoffm-aiven: Thanks! I will update the message and fix the typo

aiven/internal/mutexkv/mutexkv.go
mhoffm-aiven: Internal package seems appropriate because we do not want to export it. But I'm not passionate about it should I move it to pkg/mutexkv?

aiven/datasource_elasticsearch_acl.go
mhoffm-aiven: done

ivan-savciuc (commented):
I think we can do some code unification here. For example, we have 6 places where we run the same Get+Update logic, and I think we can create one function that takes what a user wants to append to delete to the ACL's and this function will make the diff and update ACL's. WDYT?

mhoffm-aiven:
> I think we can do some code unification here. For example, we have 6 places where we run the same Get+Update logic, and I think we can create one function that takes what a user wants to append to delete to the ACL's and this function will make the diff and update ACL's. WDYT?

Sounds good. I'll implement it soon.

ivan-savciuc (approved):

PR #569, opened 2021-09-21
Convert connection pool username field to optional

gandelman-a:
The API allows pools to be created with no pool-specific username specified.
This switches the field on the aiven_connection_pool resource similarly
and allows to reuse the incoming user by not specifying a username,
which matches behavior of the UI.

Depends on changes top aiven-go-client via https://github.com/aiven/aiven-go-client/pull/134

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/aiven/terraform-provider-aiven
 go 1.16
 
 require (
-	github.com/aiven/aiven-go-client v1.6.1
+	github.com/aiven/aiven-go-client v1.6.2-0.20210922162241-f938fd9c827e
 	github.com/aws/aws-sdk-go v1.30.12 // indirect
 	github.com/google/go-cmp v0.5.6
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.7.0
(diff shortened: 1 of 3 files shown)


cyx (commented):
👌

go.mod
```diff
@@ -12,3 +12,5 @@ require (
 	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
 )
+
```
ivan-savciuc, line 15: please use go-client@f938fd9c827ecb47f38d135b7ff84ccfaecc07f3 version, we will replace that later with the release

ivan-savciuc (changes requested):
Need to update go-client otherwise LGTM, thanks @gandelman-a

go.mod
```diff
@@ -12,3 +12,5 @@ require (
...
+replace github.com/aiven/aiven-go-client => ../aiven-go-client
```
gandelman-a, line 16: Whoops

ivan-savciuc (approved):
Thanks!

PR #572, opened 2021-09-22
Fix invalid argument name in opensearch example

Tapppi:
The `opensearch_dashboards_request_timeout` in the opensearch docs example is an invalid argument name and presumably should be `opensearch_request_timeout`


ivan-savciuc (approved):

ivan-savciuc:
@Tapppi thanks for contributing

PR #579, opened 2021-10-05
flink: acceptance test for kafka to pg; wait more in service integrat…

mhoffm-aiven:
…ion creation; fix target states in flink deletion waiter


ivan-savciuc (approved):
LGTM

PR #581, opened 2021-10-06
add azure privatelink docs

ivan-savciuc:


mhoffm-aiven (dismissed):
Should we link to the documentation here too maybe? what do you think?

ivan-savciuc:
@mhoffm-aiven I do not think so; the URL may change, probably we will have a separate blogpost soon regarding the Azure private link for Terraform. Moreover, the Aiven help article is meant for the `avn` client, not for Terraform, and it may be confusing.

mhoffm-aiven (approved):
lgtm

PR #599, opened 2021-10-25
add hybrid opensearch service support

ivan-savciuc:
After upgrading an existing Elasticsearch service to Opensearch, we create a "hybrid" service, and this PR adds support of such services to the `aiven_opensearch` resource.

### DIFF
--- a/aiven/resource_opensearch.go
+++ b/aiven/resource_opensearch.go
@@ -1,8 +1,12 @@
 package aiven
 
 import (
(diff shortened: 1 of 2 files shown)


ivan-savciuc:
@rikonen tests are mainly failing because `Operation would exceed service node quota`, and this is not related to these changes. We will try to change the node quota for this project.

rikonen (approved):

PR #606, opened 2021-11-03
fix ip_filter issue when a TF manifest has IP filters sorted in a dif…

ivan-savciuc:
…ferent way than on Aiven side


mhoffm-aiven:
Acceptance Test Failure seems to be related to flakiness in the database resource test

PR #616, opened 2021-11-09
replace reference to `tables` with `table_id` in flinks docs/examples

encima:


examples/flink/main.tf
```diff
@@ -99,13 +99,13 @@ resource "aiven_flink_job" "flink_job" {
   project = aiven_flink.flink.project
   service_name = aiven_flink.flink.service_name
   job_name = "my_job"
-  tables = [
+  table_id = [
     aiven_flink_table.source.table_name,
     aiven_flink_table.sink.table_name
```
mhoffm-aiven, line 104: should be id here too

mhoffm-aiven (changes requested):

encima:
One check is that changelog is updated but the changelog is for releases, should this not be included for every PR and instead be a requirement for releases?

mhoffm-aiven (approved):
lgtm

PR #617, opened 2021-11-09
resource_service_integration: only allow read_replica as preexisting resource

mhoffm-aiven:
resource_service_integration: simplify handling of preexisting service integrations

currently it is allowed to import arbitary service integrations which is non canonical behaviour. preexisting resources should be imported as opt-in. This PR restricts the allowed preexisting integrations to be of `read_replica` type since that is the only type that can be created alongside a service by terraform itself.

### DIFF
--- a/aiven/resource_service_integration.go
+++ b/aiven/resource_service_integration.go
@@ -11,8 +11,6 @@ import (
 
 	"github.com/aiven/aiven-go-client"
 	"github.com/aiven/terraform-provider-aiven/aiven/templates"
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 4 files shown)


aiven/resource_service_integration.go
```diff
@@ -279,9 +227,8 @@ func resourceServiceIntegrationRead(_ context.Context, d *schema.ResourceData, m
 		return nil
 	}
```
dmitrii-vasilev, line 228: It seems the code above is more indented

dmitrii-vasilev (dismissed):
Looks good. I'll check the test later.

aiven/resource_service_integration.go
dmitrii-vasilev: Looks like it's something on GitHub side

heikju: Indentation looking weird could be a tabs/spaces issue.

rikonen (approved):
Looks good to me. Such changes would be nicer to review if indentation change was in it's own commit, the general restructuring in it's own and the actual beef in third one.

aiven/resource_service_integration.go
mhoffm-aiven: i double checked; indentation is fine here; seems to be a github thing

PR #621, opened 2021-11-13
use semgrep directory configuration

MichaHoffmann:
This PR aims to structure our custom semgrep rules better by splitting the existing rules into multiple files and moving them into the default semgrep config path


ivan-savciuc (approved):
LGTM

mhoffm-aiven:
Pretty Stale now, we can reopen it if we decide to go for directory semgrep configuration, at the moment the rules we have are sufficient and it is not needed to split the file.

PR #623, opened 2021-11-15
sync flink resources

mhoffm-aiven:
This PR aims to synchronize the flink resources ( table/job ) with the upstream API.

### DIFF
--- a/aiven/resource_flink_table.go
+++ b/aiven/resource_flink_table.go
@@ -8,6 +8,7 @@ import (
 	"github.com/aiven/aiven-go-client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 var aivenFlinkTableSchema = map[string]*schema.Schema{
@@ -20,6 +21,12 @@ var aivenFlinkTableSchema = map[string]*schema.Schema{
 		ForceNew:    true,
 		Description: complex("Specifies the name of the table.").forceNew().build(),
 	},
+	"schema_sql": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: complex("The SQL statement to create the table.").forceNew().build(),
+	},
 	"integration_id": {
 		Type:        schema.TypeString,
 		Required:    true,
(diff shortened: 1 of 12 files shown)


aiven/provider.go
```diff
@@ -398,6 +398,18 @@ func validateDurationString(v interface{}, k string) (ws []string, errors []erro
 	return
 }
 
+// validateStringEnum returns a schema.ValidateFunc that ensures that a string one of a few possible values
```
ivan-savciuc, line 401: Can you please check if we can reuse `validation.StringInSlice`?

mhoffm-aiven: good catch we can!

aiven/resource_flink_table.go
```diff
@@ -8,6 +8,30 @@ import (
...
+	startupModeLargestOffset  = "largest-offset"
+	startupModeGroupOffsets   = "group-offsets"
+	startupModeTimestamp      = "timestamp"
+)
+
+var (
+	aivenFlinkTableConnectorTypes    = []string{connectorTypeKafka, connectorTypeUpsertKafka}
+	aivenFlinkTableKafkaStartupModes = []string{startupModeEarliestOffset, startupModeLargestOffset, startupModeGroupOffsets, startupModeTimestamp}
+	aivenFlinkTableKafkaValueFormats = []string{formatAvro, formatAvroConfluent, formatDebeziumAvroConfluent, formatDebeziumJson, formatJson}
+	aivenFlinkTableKafkaKeyFormats   = aivenFlinkTableKafkaValueFormats
```
ivan-savciuc, line 34: Why do we need two variables with the same values?

aiven/resource_flink_table.go
```diff
@@ -8,6 +8,30 @@ import (
...
+	formatJson                  = "json"
+
+	startupModeEarliestOffset = "earliest-offset"
+	startupModeLargestOffset  = "largest-offset"
+	startupModeGroupOffsets   = "group-offsets"
+	startupModeTimestamp      = "timestamp"
+)
+
+var (
+	aivenFlinkTableConnectorTypes    = []string{connectorTypeKafka, connectorTypeUpsertKafka}
```
ivan-savciuc, line 31: The problem with the package level variables is that someone can accidentally overwrite it's value. Can use use functions instead? 

Something like:
```golang
func getFlinkTableConnectorTypes() []string {
    return []string{connectorTypeKafka, connectorTypeUpsertKafka}
}
```

aiven/resource_flink_test.go
```diff
@@ -13,12 +13,17 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-func TestAccAiven_flinkBasic(t *testing.T) {
-	resourceName := "aiven_flink.bar"
-	serviceName := fmt.Sprintf("test-acc-flink-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+func TestAccAiven_flink(t *testing.T) {
+	t.Parallel()
```
ivan-savciuc, line 17: Looks like we do not need that since we use `resource.ParallelTest`

ivan-savciuc (dismissed):

aiven/resource_flink_test.go
mhoffm-aiven: I tested and it does not run in parallel without that! I think `resource.ParallelTest(tt)` only registers the test suit as parallel on `tt`.

aiven/resource_flink_table.go
mhoffm-aiven: they have different semantic and only coincidentally the same values i think, i'm for separating the names

aiven/resource_flink_table.go
mhoffm-aiven: :heavy_check_mark:

aiven/resource_flink_table.go
mhoffm-aiven: :heavy_check_mark:

ivan-savciuc (approved):
LGTM

PR #633, opened 2021-12-13
service_change.go: add more logging to service waiter

mhoffm-aiven:
This PR adds a little more logging to the service waiter to help debug why some service time out without apparent reason.


ivan-savciuc (approved):
LGTM

PR #634, opened 2021-12-13
actions: enable long acc tests in nightly test runs

mhoffm-aiven:
This PR enables long acc test for kafka topics in nightly runs where we do not care about the time too much.


ivan-savciuc (approved):
LGTM

PR #641, opened 2021-12-24
saml auth provider: lookup URLs correctly

mhuxtable:
The auth provider was returning the SAMLCertificate for the saml_idp_url
and saml_entity_id attributes, which prevents use of this resource in
the provider.


ivan-savciuc (approved):
LGTM

PR #667, opened 2022-01-22
provider: add provider version to user agent

MichaHoffmann:
This PR adds the provider version to the user agent for insights and better deprecation policies.


ivan-savciuc (approved):
LGTM

PR #679, opened 2022-02-08
Implement importer for static_ip

gandelman-a:
If static IPs are created and associated to a service, but the TF run is terminated ungracefully such that changes are not persisted to state, terraform becomes wedged and unable to subsequently run, or even generate a plan, against said infra. Subsequent runs will see the previously associated IP, but since its not in state will consider it needing to be dissociated, and fail on https://github.com/aiven/terraform-provider-aiven/blob/master/aiven/internal/service/custom_diff.go#L135

The only way out is to import the associated IPs.  This implements the missing importer on static ip resource.


ivan-savciuc (approved):
@gandelman-a thanks for contributing!

PR #681, opened 2022-02-09
readme: remove sudo from podman

mhoffm-aiven:
we should not tell people to use `sudo podman`, this PR fixes that line

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ For information about writing acceptance tests, see the main [Terraform contribu
 In case you'd like to avoid Go-dependency on your local machine or CI, you can run the following build command:
 
 ```
-sudo podman run -v .:/terraform-provider-aiven:z --workdir /terraform-provider-aiven golang:latest make <COMMAND>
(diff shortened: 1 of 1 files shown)


ivan-savciuc (approved):
LGTM

ivan-savciuc:
Did some testing on my box:
```
podman run -v .:/terraform-provider-aiven:z --workdir /terraform-provider-aiven golang:latest make lint 
✔ docker.io/library/golang:latest
Trying to pull docker.io/library/golang:latest...
Getting image source signatures
Copying blob 9297634c9537 done  
Copying blob 461bb1d8c517 done  
Copying blob e6d3e61f7a50 done  
Copying blob 6a0b51344be5 done  
Copying blob 412caad352a3 done  
Copying blob 0c6b8ff8c37e done  
Copying blob 856234c9f319 done  
Copying config 80d9a75ccb done  
Writing manifest to image destination
Storing signatures
Error: lsetxattr /home/ivan.savciuc/terraform-provider-aiven/hack/tools/bin/golangci-lint: operation not permitted
``` 

it does not work without `sudo` out of the box, we should extend docs with the instructions to make it work without using `sudo`.

PR #684, opened 2022-02-09
flink_table: fix possible startup modes for kafka

### LINKED ISSUE #682: Mismatched parameter name on Flink table between Terraform resource and API
In the `aiven_flink_table` resource, the option `kafka_startup_mode` has a possible value of `largest-offset` (https://github.com/aiven/terraform-provider-aiven/blob/57fe89d39a012d897eaadd632fa2e9ffb5521ab2/docs/resources/flink_table.md#L55 and https://github.com/aiven/terraform-provider-aiven/blob/da0dff84082d9f78374f549fda7211ccc98ef8b4/aiven/resource_flink_table.go#L222), but this is rejected by the API as it should be `latest-offset` instead.

mhoffm-aiven:
Fixes #682


ivan-savciuc (approved):
LGTM

PR #690, opened 2022-02-15
remove project resource deprecated fields

ivan-savciuc:


mhoffm-aiven (dismissed):
LGTM, but we should wait for release v3 with the breaking change

PR #699, opened 2022-03-04
Add missing import statement documentation

dangquangdon:
- Add `import.sh` files for the doc generator to update the documentation about `import`
- Fix some wrong examples


examples/resources/aiven_clickhouse_grant/import.sh
```diff
@@ -0,0 +1 @@
+# TOO
```
mhoffm-aiven, line 1: Better drop the file i'd say, we can add it later

mhoffm-aiven (changes requested):
LGTM thanks for the effort! Only recommendation from me would be dropping the `#TODO` imports

ivan-savciuc (approved):

PR #710, opened 2022-03-25
split aiven_database and deprecate it

mlatief:
This will introduce new service-specific database resources and data
sources, instead of the current generic `aiven_database` resource and
data source:
- `aiven_pg_database` for Aiven PostgreSQL databases.
- `aiven_mysql_database` for Aiven MySQL databases.
- `aiven_influxdb_database` for Aiven InfluxDB databases.

The old `aiven_database` will have a DeprecationMessage added, and the
description is updated to include a deprecation note.


internal/service/influxdb/resource_influxdb_database.go
```diff
@@ -0,0 +1,172 @@
...
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Delete: schema.DefaultTimeout(2 * time.Minute),
+		},
+
+		Schema: aivenInfluxDBDatabaseSchema,
+	}
+}
+
+func resourceDatabaseCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
```
dangquangdon, line 51: Perhaps this can be named something like `resourceInfluxDBCreate`, to make it more specific. And the same for other functions (Read, Delete, Update....) and similarly for PG and MySQL

internal/service/influxdb/resource_influxdb_database_test.go
```diff
@@ -0,0 +1,184 @@
...
+				PreventPostDestroyRefresh: true,
+				ExpectNonEmptyPlan:        true,
+				PlanOnly:                  true,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "project", os.Getenv("AIVEN_PROJECT_NAME")),
+					resource.TestCheckResourceAttr(resourceName, "service_name", fmt.Sprintf("test-acc-sr-%s", rName2)),
+					resource.TestCheckResourceAttr(resourceName, "database_name", fmt.Sprintf("test-acc-db-%s", rName2)),
+					resource.TestCheckResourceAttr(resourceName, "termination_protection", "true"),
+				),
+			},
```
dangquangdon, line 59: Can also have a test step for the `import` statement, since we support it as well

dangquangdon (changes requested):
Only scanning quickly through the code, but I would say it's easier if this one big commit could be split into multiple smaller ones, for example one for deprecated the old `database`, and separate ones for influxdb, pg, mysql. So that it is easier to be reviewed and compared.

And perhaps, we can utilize the old `database` code, and make it less duplicated. Given that the scope of this ticket is only to split, we can later have another ticket for the refactoring. Because as far as I see, we have a lot of similarities in the logic of these splited resources.

mlatief:
> Only scanning quickly through the code, but I would say it's easier if this one big commit could be split into multiple smaller ones, for example one for deprecated the old `database`, and separate ones for influxdb, pg, mysql. So that it is easier to be reviewed and compared.

Unfortunately that seems inevitable, as it won't make so much sense deprecating the generic resource without providing the alternatives.

> And perhaps, we can utilize the old `database` code, and make it less duplicated. Given that the scope of this ticket is only to split, we can later have another ticket for the refactoring. Because as far as I see, we have a lot of similarities in the logic of these splited resources.

We discussed this with @ivan-savciuc earlier, and came to conclusion that we might have to compromise on this redundancy, at least for now.

internal/service/pg/resource_pg_database_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { acc.TestAccPreCheck(t) },
+		ProviderFactories: acc.TestAccProviderFactories,
+		CheckDestroy:      testAccCheckAivenPGDatabaseResourceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccPGDatabaseResource(projectName, rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAivenPGDatabaseAttributes("data.aiven_pg_database.database"),
```
ivan-savciuc, line 44: Can we also validate here the `lc_ctype` and `lc_collate` fields? These should be set based on the Config.

internal/service/pg/resource_pg_database_test.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+
+		if a["service_name"] == "" {
+			return fmt.Errorf("expected to get a service_name from Aiven")
+		}
+
+		if a["database_name"] == "" {
+			return fmt.Errorf("expected to get a database_name from Aiven")
+		}
+
```
ivan-savciuc, line 234: Also here I think we can benefit from check the `lc_collate` and `lc_ctype`

ivan-savciuc (commented):
LGTM, just few minor things

ivan-savciuc (approved):
LGTM :+1:

PR #712, opened 2022-03-31
v3: fix account team member deletion

ivan-savciuc:


mlatief (approved):
LGTM. A minor comment, just wondering if it would make sense adding the `UserId` to the schema at some point.

PR #713, opened 2022-03-31
v2: fix account team member deletion

ivan-savciuc:


aiven/resource_account_team_member.go
```diff
@@ -190,7 +190,7 @@ func resourceAccountTeamMemberDelete(_ context.Context, d *schema.ResourceData,
 	// delete account team member
 	for _, m := range r.Members {
 		if m.UserEmail == userEmail {
-			err = client.AccountTeamMembers.Delete(schemautil.SplitResourceID3(d.Id()))
+			err = client.AccountTeamMembers.Delete(accountId, teamId, m.UserId)
```
dangquangdon, line 193: Maybe we can make a test for this too? I wonder why `testAccCheckAivenAccountTeamMemberResourceDestroy` didn't catch this before :thinking:

ivan-savciuc: The deletion of an account team member has two deletion operations:
1. Deletion of a user that was invited 
2. Deletion of a user who was invited and accepted an invitation and registered on the Aiven platform. 

And the problem was with the second type of user, who accepted the invitation and did the full registration. It is going to be problematic to test this use case automatically, this is why we haven't seen this bug during our test runs.

dangquangdon: Hmm, as far as I remember, `d.Id()` is `accountId/teamId/userEmail` or something like that right (the same ID pattern used for the `import`) ? So if the behavior is different when user were using `userEmail` and `UserId`, we probably should document it somewhere so that if it happen again, users can manage to fix it by themselves, I think

dangquangdon: ah never mind, with this fix, user will only have one option to use `UserId` :joy: , silly me.

PR #722, opened 2022-04-08
split vpc peering connection

ivan-savciuc:


internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -0,0 +1,337 @@
...
+
+func resourceAWSVPCPeeringConnectionCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	var (
+		pc     *aiven.VPCPeeringConnection
+		err    error
+		region *string
+	)
+
+	client := m.(*aiven.Client)
+	projectName, vpcID := schemautil.SplitResourceID2(d.Get("vpc_id").(string))
```
dmitrii-vasilev, line 89: Wondering if it's possible to use a special combined type for `vpc_id` to get `projectName, vpcID` pair and check errors right at the schema level?

internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -0,0 +1,337 @@
...
+			diags = append(diags, diag.Errorf("A user deleted the peering cloud resource in their account. "+
+				"There are no transitions from this state")...)
+		case "REJECTED_BY_PEER":
+			diags = append(diags, diag.Errorf("AWS VPC peering connection request was rejected, state info: %s",
+				stateInfoToString(pc.StateInfo))...)
+		case "INVALID_SPECIFICATION":
+			diags = append(diags, diag.Errorf("VPC peering connection cannot be created, more in the state info: %s",
+				stateInfoToString(pc.StateInfo))...)
+		default:
+			return diag.Errorf("Unknown VPC peering connection cannot state: %s", pc.State)
```
dmitrii-vasilev, line 174: Probably should be `Unknown VPC peering connection state` instead?

internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -0,0 +1,337 @@
...
+		vpcID,
+		peerCloudAccount,
+		peerVPC,
+		peerRegion,
+	)
+	if err != nil && !aiven.IsNotFound(err) {
+		return diag.Errorf("Error deleting VPC peering connection: %s", err)
+	}
+
+	stateChangeConf := &resource.StateChangeConf{
```
dmitrii-vasilev, line 217: Looks like this state change configuration is almost the same as in `resourceAWSVPCPeeringConnectionCreate()` except `Pending` and `Target` states. Can it be extracted to a function?

internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func resourceAWSVPCPeeringConnectionImport(ctx context.Context, d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	if len(strings.Split(d.Id(), "/")) != 3 {
+		return nil, fmt.Errorf("invalid identifier %v, expected <project_name>/<vpc_id>/<aws_account_id>", d.Id())
+	}
+
+	dig := resourceAWSVPCPeeringConnectionRead(ctx, d, m)
+	if dig.HasError() {
+		return nil, errors.New("cannot get aws vpc peering connection")
```
dmitrii-vasilev, line 261: Probably better to use `AWS VPC`

internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -0,0 +1,337 @@
...
+) diag.Diagnostics {
+	// Warning or errors can be collected in a slice type
+	var diags diag.Diagnostics
+
+	if err := d.Set("vpc_id", schemautil.BuildResourceID(project, vpcID)); err != nil {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  fmt.Sprintf("Unable to set vpc_id field: %s", err),
+			Detail:   fmt.Sprintf("Unable to set vpc_id field for AWS VPC peering connection: %s", err),
+		})
```
dmitrii-vasilev, line 281: Probably at least `diag.Diagnostic` can be created inside a function with `field_name` and `err` arguments.

internal/service/vpc/resource_azure_vpc_peering_connection.go
```diff
@@ -0,0 +1,363 @@
...
+		peerResourceGroup = v.(string)
+	}
+
+	if _, err = client.VPCPeeringConnections.Create(
+		projectName,
+		vpcID,
+		aiven.CreateVPCPeeringConnectionRequest{
+			PeerCloudAccount:  azureSubscriptionId,
+			PeerVPC:           vnetName,
+			PeerRegion:        region,
```
dmitrii-vasilev, line 126: It seems `region` is always `nil`. Is it expected?

internal/service/vpc/resource_azure_vpc_peering_connection.go
```diff
@@ -0,0 +1,363 @@
...
+			PeerAzureAppId:    peerAzureAppID,
+			PeerAzureTenantId: peerAzureTenantID,
+			PeerResourceGroup: peerResourceGroup,
+		},
+	); err != nil {
+		return diag.Errorf("Error waiting for VPC peering connection creation: %s", err)
+	}
+
+	stateChangeConf := &resource.StateChangeConf{
```
dmitrii-vasilev, line 135: The state change configuration probably can be de-duplicated between all clouds.

internal/service/vpc/resource_azure_vpc_peering_connection.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	res, err := stateChangeConf.WaitForStateContext(ctx)
+	if err != nil {
+		return diag.Errorf("Error creating VPC peering connection: %s", err)
+	}
+
+	diags := diag.Diagnostics{}
+	pc = res.(*aiven.VPCPeeringConnection)
+	if pc.State != "ACTIVE" {
```
dmitrii-vasilev, line 171: Probably can be de-duplicated between all clouds

dmitrii-vasilev (changes requested):
Looks good in general. I've added some questions and comments about code blocks which probably can be de-duplicated by extracting to functions.

internal/service/vpc/resource_aws_vpc_peering_connection.go
ivan-savciuc: I agree, though Go convention is to use only lowercase letters when constructing errors

internal/service/vpc/resource_azure_vpc_peering_connection.go
ivan-savciuc: :+1:

internal/service/vpc/resource_aws_vpc_peering_connection.go
ivan-savciuc: indeed, there is no value in what we are doing since it can be generated :+1:

internal/service/vpc/resource_azure_vpc_peering_connection.go
ivan-savciuc: In the current shape yes it could be a good candidate for de-duplication, however, we have a problem (not related to this PR) with this part of the code that will be fixed in a separate PR. We are trying to get Azure, GCP and AWS VPC peering connection using the same set of fields which is not enough and it only works properly for the AWS now and in some cases for GCP. So this logic will be updated soon and we will have a different state change waiter and get functions for each cloud type.

internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -0,0 +1,337 @@
...
+	return &schema.Resource{
+		Description:   "The AWS VPC Peering Connection resource allows the creation and management of Aiven AWS VPC Peering Connections.",
+		CreateContext: resourceAWSVPCPeeringConnectionCreate,
+		ReadContext:   resourceAWSVPCPeeringConnectionRead,
+		DeleteContext: resourceAWSVPCPeeringConnectionDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceAWSVPCPeeringConnectionImport,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(2 * time.Minute),
```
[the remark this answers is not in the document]
ivan-savciuc: this is a default value and a customer can set any timeout they like, and during my tests - yes 2 minutes is enough

internal/service/vpc/resource_aws_vpc_peering_connection.go
ivan-savciuc: :+1:

internal/service/vpc/datasource_gcp_vpc_peering_connection.go
```diff
@@ -0,0 +1,43 @@
...
+	gcpProjectId := d.Get("gcp_project_id").(string)
+	peerVPC := d.Get("peer_vpc").(string)
+
+	vpc, err := client.VPCs.Get(projectName, vpcID)
+	if err != nil {
+		return diag.Errorf("Error deleting VPC peering connection: %s", err)
+	}
+
+	for _, peer := range vpc.PeeringConnections {
+		if peer.PeerCloudAccount == gcpProjectId && peer.PeerVPC == peerVPC {
```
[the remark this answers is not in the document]
ivan-savciuc: each cloud data source is a bit different, id constructed from a different set of elements

dmitrii-vasilev (approved):
LGTM

Issue #720, opened 2022-04-07
`billing_emails` not stored in state for projects when set through `billing_group` so subsequent changes remove the emails


encima:
Setting `billing_emails` in a `billing_group` does not store them for projects (but returned in response) so subsequent runs will show the projects as being modified by removing the `billing_emails`

```
resource "aiven_account" "tst-account" {
  name = "aivener-test"  
}

resource "aiven_billing_group" "tst-bg" {
    name             = "test-bg"
    billing_currency = "USD"
    vat_id           = "123ABC"
    account_id = aiven_account.tst-account.account_id
    billing_emails = [ "encima@aiven.fi", "encima+1@aiven.fi" ]
    depends_on = [
      aiven_account.tst-account
    ]
}

## Do not use this project, keep it as a template
resource "aiven_project" "template-proj" {
    project = "template-proj"
    account_id = aiven_account.tst-account.account_id
    billing_group = aiven_billing_group.tst-bg.id
    default_cloud = "aws-eu-west-1"
    depends_on = [
      aiven_billing_group.tst-bg,
      aiven_account.tst-account
    ]
}
```

Response on subsequent plan:
<img width="507" alt="image" src="https://user-images.githubusercontent.com/517923/162123810-7cac517b-69fe-4289-82e1-3ca1c169f8be.png">

wanisfahmyDE:
can confirm that it works, thanks!

PR #724, opened 2022-04-12
v2: project resource fix import and read for deprecated billing group fields

ivan-savciuc:
Project resource has deprecated fields related to billing group, and in a scenario when a project is associated with a billing group and these deprecated fields are not set in Terraform manifest TF refresh generates none empty plan. With this patch, we will only refresh those fields if they are set to any non zero value since API retrieves back values transferred from associated billing groups.

v3 has no such problem

### DIFF
--- a/aiven/resource_project.go
+++ b/aiven/resource_project.go
@@ -79,6 +79,7 @@ var aivenProjectSchema = map[string]*schema.Schema{
 	"payment_method": {
 		Type:        schema.TypeString,
 		Computed:    true,
+		Deprecated:  "Please use aiven_billing_group resource to set this value.",
 		Description: "The method of invoicing used for payments for this project, e.g. `card`.",
 	},
 	"billing_group": {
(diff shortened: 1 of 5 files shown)


aiven/resource_project.go
```diff
@@ -516,11 +517,15 @@ func setProjectTerraformProperties(d *schema.ResourceData, client *aiven.Client,
...
+			return diag.FromErr(err)
+		}
 	}
-	if err := d.Set("country_code", project.CountryCode); err != nil {
-		return diag.FromErr(err)
+	if _, ok := d.GetOk("country_code"); ok {
+		if err := d.Set("country_code", project.CountryCode); err != nil {
+			return diag.FromErr(err)
+		}
 	}
```
dmitrii-vasilev, line 529: This `GetOk` / `Set` pattern probably can be extracted to a function

dmitrii-vasilev (dismissed):
LGTM. Just a small comment.

mlatief:
Closes #720.

mlatief (approved):
LGTM.

aiven/resource_project.go
ivan-savciuc: done

PR #728, opened 2022-04-14
v3: add support for account primary_billing_group_id

ivan-savciuc:


dmitrii-vasilev (approved):
LGTM

wanisfahmyDE:
updated to 2.7.2 and tried this out.
if the `primary_billing_group_id` is not set in the `aiven_account` resource, it will attempt to destroy the account..
it seems that this attribute is now mandatory, it only worked when I actually added this attribute.

docs state that it's optional: https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/account#optional

PR #738, opened 2022-05-11
add support for a new user configuration option

ivanr-aiven:


ivan-savciuc:
@ivanr-aiven thanks for this PR, thought we will introduce these new user configuration options here as well https://github.com/aiven/terraform-provider-aiven/pull/736.

PR #741, opened 2022-05-16
add support for project tags

ivan-savciuc:


internal/service/project/resource_project.go
```diff
@@ -68,6 +68,25 @@ var aivenProjectSchema = map[string]*schema.Schema{
 		Description:      schemautil.Complex("The id of the billing group that is linked to this project.").Referenced().Build(),
 		DiffSuppressFunc: schemautil.EmptyObjectDiffSuppressFunc,
 	},
+	"tag": {
```
dmitrii-vasilev, line 71: Why is it `tag` and not `tags`? It looks confusing below with `d.Set("tag", tags)`

internal/service/project/resource_project.go
```diff
@@ -372,5 +393,27 @@ func setProjectTerraformProperties(d *schema.ResourceData, client *aiven.Client,
...
+	for k, v := range project.Tags {
+		tags = append(tags, map[string]interface{}{
+			"key":   k,
+			"value": v,
+		})
+	}
+	if err := d.Set("tag", tags); err != nil {
+		return diag.FromErr(err)
+	}
+
```
dmitrii-vasilev, line 406: Maybe this code should be extracted to a function? Then all properties in this function will have kind of the same set pattern. Something like the following:
```go
if err := setTags(d, project.Tags); err != nil {
    return diag.FromErr(err)
}
```

dmitrii-vasilev (changes requested):
A couple of comments / questions

internal/service/project/resource_project.go
ivan-savciuc: This is why, a user will be adding tag one be one 
```
                  tag {
		    key   = "test"
		    value = "val"
		  }
		  tag {
		    key   = "test"
		    value = "val2"
		  }
```

internal/service/project/resource_project.go
ivan-savciuc: done

dmitrii-vasilev (approved):
LGTM

PR #744, opened 2022-05-26
Removing `partitioned_by` field as Aiven API has no such argument.

dewan-ahmed:
Removing `partitioned_by` field from the resource "aiven_flink_table" as Aiven API has no such argument.


ivan-savciuc (approved):
LGTM

PR #746, opened 2022-05-27
If a project exists, then error if trying to create it again.

### LINKED ISSUE #745: Provider v3.0.0 allows project to look like it is created even though it pre-exists
When running `terraform apply` with a newly-defined `resource "aiven_project"` block, but where the block uses a project name which already exists elsewhere in the account, then instead of throwing a `Project already exists` error, the Terraform provider tells me that it has successfully created the project.

v2.7.3 gives me `Error: 409: {"errors":[{"message":"Project already exists","status":409}],"message":"Project already exists"}`

kevinbowman-aiven:
Fixes #745


ivan-savciuc (approved):
LGTM

PR #749, opened 2022-06-03
Add copy from billing group support [DARK-232]

dangquangdon:
Allow creating billing group by copying from an existing one. For example

```
resource "aiven_billing_group" "bar2" {
    name           			= "copy-test-acc-bg-%s"
    copy_from_billing_group     = aiven_billing_group.bar.id
}
```

[DARK-232]

[DARK-232]: https://aiven.atlassian.net/browse/DARK-232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/service/project/resource_billing_group.go
```diff
@@ -91,6 +91,12 @@ var aivenBillingGroupSchema = map[string]*schema.Schema{
 		DiffSuppressFunc: schemautil.EmptyObjectNoChangeDiffSuppressFunc,
 		Description:      "State",
 	},
+	"copy_from_billing_group": {
```
ivan-savciuc, line 94: Since this field can only be setduring the creation of a resource please add `DiffSuppressFunc: schemautil.CreateOnlyDiffSuppressFunc,` and since it cannot change after creation we do not need `schemautil.EmptyObjectNoChangeDiffSuppressFunc`

ivan-savciuc (changes requested):
LGTM, just a small thing:

PR #753, opened 2022-06-09
Removing "0.0.0.0/0" that allows all IP addresses

dewan-ahmed:
Removing "0.0.0.0/0" since this will allow all IP addresses and the filter will be of no use.


mhoffm-aiven (approved):
lgtm

PR #754, opened 2022-06-10
Update Kafka version, remove Karapace and fix data type

dewan-ahmed:
- Updating the Kafka version in the example. You cannot create services using 2.4 version anymore.
- Karapace Boolean is a throwback to when there was a choice between Confluent Schema Registry and Karapace
- The type for `auto_create_topics_enable` is a Boolean; not a String


ivan-savciuc (approved):
LGTM

PR #760, opened 2022-06-16
Allow retrieving project vpc datasource by id [DARK-628]

dangquangdon:
Project VPC datasource now can be retrieved using `id`. The `id` argument is Optional and would be useful in case there are multiple VPCs in the same cloud name.

User would also be notifed and given some hints from the error message when it happens. For example
image

[DARK-628]

[DARK-628]: https://aiven.atlassian.net/browse/DARK-628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/service/vpc/datasource_project_vpc.go
```diff
@@ -24,24 +32,58 @@ func datasourceProjectVPCRead(_ context.Context, d *schema.ResourceData, m inter
...
 	projectName := d.Get("project").(string)
 	cloudName := d.Get("cloud_name").(string)
+	vpcId, hasId := d.GetOk("id")
 
 	vpcs, err := client.VPCs.List(projectName)
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
+	filtered_vpcs := make([]*aiven.VPC, 0)
```
ivan-savciuc, line 42: please use camel Case for all the variables

internal/service/vpc/datasource_project_vpc.go
```diff
@@ -24,24 +32,58 @@ func datasourceProjectVPCRead(_ context.Context, d *schema.ResourceData, m inter
...
+		}
+		return diag.Errorf("project %s has multiple VPC defined for %s. Please add `id` to get the desired one. The available vpc ids are:\n%s",
+			projectName, cloudName, smg)
+	}
+
+	vpc := filtered_vpcs[0]
+
+	d.SetId(schemautil.BuildResourceID(projectName, vpc.ProjectVPCID))
+
+	err = copyVPCPropertiesFromAPIResponseToTerraform(d, vpc, projectName)
```
ivan-savciuc, line 83: I think it is best that we call here `resourceProjectVPCRead` instead of `copyVPCPropertiesFromAPIResponseToTerraform`

ivan-savciuc (changes requested):
LGTM, just few small comments :+1:

internal/service/vpc/datasource_project_vpc.go
dangquangdon: hmm, using `resourceProjectVPCRead` would call the API once again to get the data when we already have it, so I think it's not necessary. :thinking: Also we used `copyVPCPropertiesFromAPIResponseToTerraform` in the old code as well so using it again here is kinda OK I think :sweat_smile:

internal/service/vpc/datasource_project_vpc.go
```diff
@@ -24,24 +32,56 @@ func datasourceProjectVPCRead(_ context.Context, d *schema.ResourceData, m inter
...
-		if vpc.CloudName == cloudName {
-			d.SetId(schemautil.BuildResourceID(projectName, vpc.ProjectVPCID))
-			err = copyVPCPropertiesFromAPIResponseToTerraform(d, vpc, projectName)
-			if err != nil {
-				return diag.FromErr(err)
+		condition := vpc.CloudName == cloudName
+
+		// We don't need to care about the cloud name if the ID is provided
+		if hasId {
+			split_id := schemautil.SplitResourceID(vpcId.(string), 2)
```
ivan-savciuc, line 49: can we also change `split_id` to camel Case - `splitID`?

PR #771, opened 2022-06-23
Updating service versions and fix namespace issue

dewan-ahmed:
- Updating service versions that were no longer in use
- Fixing M3DB namespace name in the example
- Updating provider version to latest
- Added the change to CHANGELOG


mhoffm-aiven (approved):
lgtm

PR #778, opened 2022-06-28
docs: minor naming fix in README.md

Serpentiel:

minor naming fix for some URLs in README.md

## Why this way

to make it unified with the name of actual docs those links refer to


ivan-savciuc (approved):
LGTM

PR #780, opened 2022-06-28
fix .gitignore typo

Serpentiel:

fixes .gitignore typo

## Why this way

because English exists :)


mhoffm-aiven (approved):
lgtm

PR #781, opened 2022-06-28
enforce changelog entry for every PR

Serpentiel:

enforces `CHANGELOG.md` entry to be expected for every PR, i.e. enables `.github/workflow/changelog-checker.yml` to run on every PR

## Why this way

it's difficult for maintainers to put everything in place when preparing a release


mhoffm-aiven (approved):
lgtm

PR #783, opened 2022-06-29
refactor VPC peering connection import

Serpentiel:

added an internal metadata holder structure which allows to keep some information persistent throughout an on-going operation

## Why this way

- to be able to detect import operations
- to be able to keep track of some other information in future

### DIFF
--- a/internal/meta/meta.go
+++ b/internal/meta/meta.go
@@ -1,0 +1,13 @@
+package meta
+
+import "github.com/aiven/aiven-go-client"
+
+// Meta is the metadata for the Aiven provider.
+type Meta struct {
+	// Client is the Aiven client.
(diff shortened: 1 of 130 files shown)


Serpentiel:
acceptance tests [were also executed](https://github.com/aiven/terraform-provider-aiven/actions/runs/2584717781) manually for as a precaution measure

internal/meta/meta.go
```diff
@@ -0,0 +1,13 @@
+package meta
+
+import "github.com/aiven/aiven-go-client"
+
+// Meta is the metadata for the Aiven provider.
+type Meta struct {
```
Serpentiel, line 6: that's the actual thing that was added

Serpentiel:
a different approach has been discovered, thus there's no need for the metadata holding structure at this very moment

PR #787, opened 2022-07-04
test modification of `static_ips` in Terraform configs

Serpentiel:

- fixes PostgreSQL acceptance test with `static_ips` to actually check for their existence after service's creation
- adds acceptance test coverage for modification of `static_ips` in Terraform configs (via PostgreSQL)
- fixes `CustomizeDiffCheckStaticIpDisassociation` behavior
- made it possible to delete static IPs in a single step, without dissociating them

## Why this way

- to make acceptance tests better
- to improve user experience

### DIFF
--- a/internal/schemautil/custom_diff.go
+++ b/internal/schemautil/custom_diff.go
@@ -28,7 +28,7 @@ func TagsShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
 	return len(new.(*schema.Set).List()) != 0
 }
 
-func CustomizeDiffCheckUniqueTag(ctx context.Context, d *schema.ResourceDiff, m interface{}) error {
+func CustomizeDiffCheckUniqueTag(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
 	t := make(map[string]bool)
 	for _, tag := range d.Get("tag").(*schema.Set).List() {
 		tagVal := tag.(map[string]interface{})
--- a/internal/schemautil/custom_diff.go
+++ b/internal/schemautil/custom_diff.go
@@ -144,14 +144,7 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
 
 	// Check that we block deletions that will fail because the ip belongs to the service
 	for _, sip := range resp.StaticIPs {
-		assignedToTheService := sip.ServiceName == serviceName && sip.State == staticIpAssigned
-		aboutToBeDisassigned := !contains(plannedStaticIps, sip.StaticIPAddressID)
-		if assignedToTheService && aboutToBeDisassigned {
-			return fmt.Errorf("the static ip '%s' is currently assigned and cannot be disassociated from service '%s'", sip.StaticIPAddressID, serviceName)
-		}
-
 		associatedWithDifferentService := sip.ServiceName != "" && sip.ServiceName != serviceName
-
 		if associatedWithDifferentService && contains(plannedStaticIps, sip.StaticIPAddressID) {
 			return fmt.Errorf("the static ip '%s' is currently associated with service '%s'", sip.StaticIPAddressID, sip.ServiceName)
 		}
--- a/internal/service/static_ip/resource_static_ip.go
+++ b/internal/service/static_ip/resource_static_ip.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/aiven/aiven-go-client"
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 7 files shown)


internal/schemautil/custom_diff.go
```diff
@@ -116,6 +116,16 @@ func CustomizeDiffCheckPlanAndStaticIpsCannotBeModifiedTogether(_ context.Contex
 	return nil
 }
 
+// CustomizeDiffCheckStaticIpsExistence prevents users from modifying `aiven_static_ip` resources in
+// the same step with the resources that uses them due to Terraform limitation
+// See: https://terraform.io/plugin/sdkv2/resources/customizing-differences
+func CustomizeDiffCheckStaticIpsExistence(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	if _, exists := d.GetOkExists("static_ips"); d.Id() != "" && !exists {
```
ivan-savciuc, line 123: Also we can replace `d.Id() != ""` to `!d.IsNewResource`, perhaps it is more verbose in the IF statement.

internal/schemautil/custom_diff.go
```diff
@@ -116,6 +116,16 @@ func CustomizeDiffCheckPlanAndStaticIpsCannotBeModifiedTogether(_ context.Contex
 	return nil
 }
 
+// CustomizeDiffCheckStaticIpsExistence prevents users from modifying `aiven_static_ip` resources in
+// the same step with the resources that uses them due to Terraform limitation
+// See: https://terraform.io/plugin/sdkv2/resources/customizing-differences
+func CustomizeDiffCheckStaticIpsExistence(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
```
ivan-savciuc, line 122: Do I get this right? 

Do we error when a service static_ip's list is empty, and a resource already exists? In this case, what would happen if a customer doesn't care about the static IPs and updates his service with something unrelated? I guess a user would get this error in this case.

ivan-savciuc (changes requested):
Please check custom diff function

internal/schemautil/custom_diff.go
Serpentiel: we error in the following scenario:

- there is already a service deployed via Terraform/that Terraform is aware of
- the service has `aiven_static_ip` dependence, i.e. the service has static IPs assigned to it
- user tries to update `aiven_static_ip` resource and the dependent service in one `terraform apply` run

ivan-savciuc, line 122: Can you please add a test that validates this diff function? 
In one step, we should get this error `please modify 'aiven_static_ip' resource in a separate step`, and the next test step should fix that and showcase that it is gone.

ivan-savciuc (changes requested):
Pls add test that is validating `CustomizeDiffCheckStaticIpsExistence`

ivan-savciuc (changes requested):
I did some digging on how static IPs work and have concerns about whether we need `CustomizeDiffCheckStaticIpsExistence` additional custom diff. Please, check comments bellow. 

Moreover, during tests I found a bug in the `CustomizeDiffCheckStaticIpDisassociation` function, but lets fix it in a separate branch.

ivan-savciuc (approved):
LGTM. @MichaHoffmann, would you like to check this one? Since you've been working on static IPs.

Serpentiel:
> Are we not able to do it in one step by dropping that customize diff check?

no, the API fails

Serpentiel:
> > Are we not able to do it in one step by dropping that customize diff check?
> 
> no, the API fails

upd: I was able to achieve it, it's a bit tricky, but it works really good:
```log
2022/07/08 20:05:36 [DEBUG] Creating an instance of TopicCache ...
=== RUN   TestAccAivenPG_static_ips
=== PAUSE TestAccAivenPG_static_ips
=== CONT  TestAccAivenPG_static_ips
--- PASS: TestAccAivenPG_static_ips (415.09s)
PASS
```

internal/service/static_ip/resource_static_ip.go
```diff
@@ -98,19 +97,35 @@ func resourceStaticIPCreate(ctx context.Context, d *schema.ResourceData, m inter
...
 
 	project, staticIPAddressId := schemautil.SplitResourceID2(d.Id())
 
-	err := client.StaticIPs.Delete(
+	staticIP, err := client.StaticIPs.Get(project, staticIPAddressId)
+	if err != nil && !aiven.IsNotFound(err) {
+		return diag.FromErr(err)
+	} else {
+		if staticIP.State == schemautil.StaticIpAssigned || staticIP.State == schemautil.StaticIpAvailable {
+			if err := client.StaticIPs.Dissociate(project, staticIPAddressId); err != nil {
```
ivan-savciuc, line 110: Let's say a service is created with a bunch of static IPs signed (each service has a minimum number, like for PG `startup-2`, it is two static IPs, and for `business-4`, it is 4 IPs). Then we want to disassociate one of these minimum required (these IPs have status `assigned`) IPs from the service API, which will retrieve `409 Static IP XXXX is in the state assigned`. 

So this code would only work if we created a PG service with the plan `startup-2` and added 3 IPs instead of 2 required (in this case first 2 IP addresses will in the state `assigned` and one in `available`)  and wanted to remove 3rd IP (that is in the state `available`). 

So at lest we should change `if staticIP.State == schemautil.StaticIpAssigned || staticIP.State == schemautil.StaticIpAvailable` to `if staticIP.State == schemautil.StaticIpAvailable`

internal/service/static_ip/resource_static_ip.go
```diff
@@ -98,19 +97,35 @@ func resourceStaticIPCreate(ctx context.Context, d *schema.ResourceData, m inter
...
+				return diag.FromErr(err)
+			}
+
+			if err := resourceStaticIPWait(ctx, d, m); err != nil {
+				return diag.FromErr(fmt.Errorf("unable to wait for static ip to become unassigned: %w", err))
```
ivan-savciuc, line 115: Pls use `diag.Errorf` instead of `diag.FromErr(fmt.Errorf(`

internal/service/static_ip/resource_static_ip.go
```diff
@@ -121,7 +136,7 @@ func resourceStaticIPWait(ctx context.Context, d *schema.ResourceData, m interfa
 
 	conf := resource.StateChangeConf{
 		Target:  []string{"created"},
-		Pending: []string{"creating", "waiting"},
+		Pending: []string{"creating", "waiting", schemautil.StaticIpAssigned, schemautil.StaticIpAvailable},
```
ivan-savciuc, line 139: pls add `"creating", "waiting" and "created"` as constants

ivan-savciuc (changes requested):
Pls check the static IPs delete function comments

internal/service/static_ip/resource_static_ip.go
Serpentiel: I did it this way for consistence purposes because it's made this way in `resourceStaticIPCreate` too

PR #789, opened 2022-07-05
ci: add sweep job that runs avery day at 3 AM

ivan-savciuc:


.github/workflows/sweep.yml
```diff
@@ -0,0 +1,30 @@
+name: Sweep
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
```
Serpentiel, line 5: what time zone does it uses?
me and @mhoffm-aiven agreed that it should do that at 3.00 UTC

Serpentiel (changes requested):

.github/workflows/sweep.yml
mhoffm-aiven: its UTC

ivan-savciuc: by default docker Ubuntu image uses UTC, so we are good

ivan-savciuc:
Why 3 AM, not  1 AM or any other time? My idea is to clean up leaked infrastructure by the end of the day, and the earliest convenient time is 1 AM.

mhoffm-aiven (dismissed):
lgtm

Serpentiel (approved):
lgtm

PR #795, opened 2022-07-07
ci: change from master to main branch

ivan-savciuc:


.github/workflows/release-artifacts.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - uses: actions/checkout@master
+    - uses: actions/checkout@main
```
mhoffm-aiven, line 12: i think this should still be master since it refers to the version of `actions/checkout`

mhoffm-aiven (dismissed):
see comment

Serpentiel (approved):
lgtm

Issue #794, opened 2022-07-06
Account Authentication Method resource should be updated to match API spec


encima:
## What is currently missing?

Account Authentication Method [resource](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/account_authentication) does not support `auto_join_team_id` so a default team to join on signup/link cannot be set

## How could this be improved?

Update resource to match [API spec](https://api.aiven.io/doc/#operation/AccountAuthenticationMethodCreate)

## Is this a feature you would work on yourself?

* [ ] I plan to open a pull request for this feature.

Serpentiel:
it will take us a while to publish a new release

in case of urgency, you're very welcome to use the code provided in #797 and build it manually from sources

PR #797, opened 2022-07-07
add support for `auto_join_team_id` in account authentication resource

Serpentiel:

adds support for `auto_join_team_id` in account authentication resource

resolves: #794

## Why this way

to match API docs

### DIFF
--- a/internal/service/account/resource_account_authentication.go
+++ b/internal/service/account/resource_account_authentication.go
@@ -16,22 +16,27 @@ var aivenAccountAuthenticationSchema = map[string]*schema.Schema{
 		Required:    true,
 		Description: "The unique id of the account.",
 	},
+	"name": {
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "The name of the account authentication.",
+	},
 	"type": {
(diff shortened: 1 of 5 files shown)


internal/service/account/resource_account_authentication.go
```diff
@@ -16,22 +16,32 @@ var aivenAccountAuthenticationSchema = map[string]*schema.Schema{
...
+		Required:    true,
+		Description: "The name of the account authentication.",
+	},
 	"type": {
 		Type:         schema.TypeString,
 		Required:     true,
 		ValidateFunc: validation.StringInSlice([]string{"internal", "saml"}, false),
 		Description:  schemautil.Complex("The account authentication type.").PossibleValues("internal", "saml").Build(),
 	},
-	"name": {
```
ivan-savciuc, line 25: we usually arrange fields in this order: 
- first required fields
- then optional
- then computed

ivan-savciuc (changes requested):
Can you add a test checking the `auto_join_team_id`? Otherwise LGTM

ivan-savciuc (approved):
LGTM

PR #800, opened 2022-07-11
fix billing group data-source

ivan-savciuc:

Fixes the `aiven_billing_group` data-source

## Why this way

The billing group name is not unique; therefore, we have to use `billing_group_id` fields instead of `name` as a required field.


Serpentiel (dismissed):
lgtm

ivan-savciuc:
UPDATE: also updated `golangci-lint` to the latest

Serpentiel (dismissed):

ivan-savciuc:
I think we can merge this `go_test (1.16.x, ubuntu-latest)` is no longer valid

Serpentiel (dismissed):
lgtm

PR #804, opened 2022-07-12
fix typo in sweeper

Serpentiel:

fixes typo in sweeper

## Why this way

because typos aren't cool


mhoffm-aiven (approved):
lgtm

PR #805, opened 2022-07-12
small static IP import fix

Serpentiel:

small static IP import fix

## Why this way

to improve static IP import and tests

### DIFF
--- a/internal/service/static_ip/resource_static_ip.go
+++ b/internal/service/static_ip/resource_static_ip.go
@@ -50,7 +50,7 @@ func ResourceStaticIP() *schema.Resource {
 		ReadContext:   resourceStaticIPRead,
 		DeleteContext: resourceStaticIPDelete,
 		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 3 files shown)


internal/service/static_ip/resource_static_ip.go
```diff
@@ -169,7 +169,7 @@ func resourceStaticIPState(_ context.Context, d *schema.ResourceData, m interfac
 
 	sip, err := client.StaticIPs.Get(project, staticIPAddressId)
 	if err != nil {
-		return nil, err
+		return nil, schemautil.ResourceReadHandleNotFound(err, d)
```
ivan-savciuc, line 172: I think we can use here `schema.ImportStatePassthroughContext`

ivan-savciuc (changes requested):
Perhaps we can delete static IP import function and replace it with `schema.ImportStatePassthroughContext`?

ivan-savciuc (approved):
LGTM

PR #806, opened 2022-07-13
Fix testAccKafkaACLWrongUsernameResource

juha-aiven:

Change the username in the test to an invalid value. The test was failing.


mhoffm-aiven (approved):
lgtm

PR #809, opened 2022-07-14
fix: use go117

mhoffm-aiven:
Newer goreleaser versions require go1.17 to build all GOOS/GOARCH combinations


ivan-savciuc (approved):
LGTM

PR #811, opened 2022-07-15
Kafka Topic waiter optimizations

ivan-savciuc:
- User `ImportStatePassthroughContext` for Kafka Topic importer
- Remove v1 API fallback functionality for Kafka Topic availability waiter
- Kafka topic availability waiter optimization by reducing `PollInterval` to `5s`

## Why this way

Before TF v2 (v2.0.7), we've been using the Kafka Topics List endpoint that retrieves a list of all Kafka Topics for service. Around that time, Aiven added support for Kafka Topic config. Unfortunately, the config wasn't included in the Kafka Topic List API response `GET v1/project/{$project}/service/${service}/topic`. As a result, we had to query each topic one by one `GET v1/project/{$project}/service/${service}/topic/${topic_name}`, which is not optimal from the requests per second to API point of view and customer perspective (clearly, this approach would jeopardize performance). Finally, Kafka Topic v2 endpoint was created `POST v2/project/{$project}/service/${service}/topic`, which allowed us to query up to 100 Kafka Topics at the time, and we still use it today. 

Kafka Topic availability waiter is gathering Kafka Topics that need to be refreshed in a batch (up to 100 Topics). Therefore, the waiter queries the v2 endpoint sequentially despite Terraform's asynchronous management of the resources. 

That leads us to the`-parallelism` parameter usage; the default value is 10. But unfortunately, there is no correct number for everyone, since it depends on the number of Kafka topics in manifest and on the virtual machine (available CPU/RAM) where the Terraform command is executed. So from the waiter implementation, it doesn't make sense to have `-parallelism > 100`. 

In my tests, I was using `-parallelism=60` for the 2k Topics clusters; here are results:

**v3.3.0**
```
terraform apply --parallelism=60 | gnomon 
Total   609.3299s
```
**this PR**

```
terraform apply --parallelism=60 | gnomon 
Total   188.6568s
```

As you can see there is some performance gain, and amount of API calls per second is OK.

### DIFF
--- a/internal/service/kafka/kafka_topic_availability.go
+++ b/internal/service/kafka/kafka_topic_availability.go
@@ -125,11 +125,10 @@ func (w *KafkaTopicAvailabilityWaiter) Conf(timeout time.Duration) *resource.Sta
 	log.Printf("[DEBUG] Kafka Topic availability waiter timeout %.0f minutes", timeout.Minutes())
 
 	return &resource.StateChangeConf{
-		Pending:        []string{"CONFIGURING"},
-		Target:         []string{"ACTIVE"},
-		Refresh:        w.RefreshFunc(),
-		Timeout:        timeout,
-		PollInterval:   30 * time.Second,
-		NotFoundChecks: 50,
+		Pending:      []string{"CONFIGURING"},
+		Target:       []string{"ACTIVE"},
+		Refresh:      w.RefreshFunc(),
+		Timeout:      timeout,
+		PollInterval: 5 * time.Second,
 	}
 }
--- a/internal/service/kafka/kafka_topic_availability.go
+++ b/internal/service/kafka/kafka_topic_availability.go
@@ -107,15 +107,6 @@ func (w *KafkaTopicAvailabilityWaiter) refresh() error {
 		log.Printf("[DEBUG] Kafka Topic queue: %+v", queue)
 		v2Topics, err := w.Client.KafkaTopics.V2List(w.Project, w.ServiceName, queue)
 		if err != nil {
-			// if v2 endpoint retrieves 409 response code, it means that Kafka service has old nodes and
-			// v2 endpoint is not available, therefore using v1.
-			if err.(aiven.Error).Status == 409 {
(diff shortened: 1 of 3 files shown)


internal/service/kafka/kafka_topic_availability.go
```diff
@@ -129,29 +120,15 @@ func (w *KafkaTopicAvailabilityWaiter) refresh() error {
...
 	log.Printf("[DEBUG] Kafka Topic availability waiter timeout %.0f minutes", timeout.Minutes())
 
 	return &resource.StateChangeConf{
-		Pending:        []string{"CONFIGURING"},
-		Target:         []string{"ACTIVE"},
-		Refresh:        w.RefreshFunc(),
-		Timeout:        timeout,
-		PollInterval:   30 * time.Second,
-		NotFoundChecks: 50,
+		Pending:      []string{"CONFIGURING"},
```
mhoffm-aiven, line 128: For better error messages ( it currently says still "CONFIGURING" when timing out), can we consider "PENDING" ?

ivan-savciuc: We have the following Kafka Topic States:
```
@enum.unique
class KafkaTopicState(str, enum.Enum):
    active = "ACTIVE"
    configuring = "CONFIGURING"
    deleting = "DELETING"
``` 
so in our case it's either `CONFIGURING` or `ACTIVE`

mhoffm-aiven: ah ok! sorry i thought this was our tf waiter state

mhoffm-aiven (approved):
lgtm

Serpentiel:
@kkonstan-ovo could you please check if https://github.com/aiven/terraform-provider-aiven/pull/848 improves it?

kkonstan-ovo:
Apologies for the late reply, missed the notification. I can confirm it works and we've been using 3.5.0 an now 3.5.1 for a while, thanks

Serpentiel:
you're very welcome!

PR #816, opened 2022-07-19
add Flink SQL validation

Serpentiel:

adds Flink SQL validation

## Why this way

users will struggle less when they know about SQL errors straight away and not during mid-apply

### DIFF
--- a/internal/schemautil/custom_diff.go
+++ b/internal/schemautil/custom_diff.go
@@ -9,15 +9,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func ServiceIntegrationShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
-	return len(new.([]interface{})) != 0
+func ResourceShouldExist(_ context.Context, d *schema.ResourceDiff, _ interface{}) bool {
+	return len(d.Id()) > 0
 }
 
-func CustomizeDiffServiceIntegrationAfterCreation(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
-	if len(d.Id()) > 0 && d.HasChange("service_integrations") && len(d.Get("service_integrations").([]interface{})) != 0 {
-		return fmt.Errorf("service_integrations field can only be set during creation of a service")
-	}
-	return nil
+func ServiceIntegrationShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
+	return len(new.([]interface{})) != 0
 }
 
 func DiskSpaceShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
@@ -26,6 +23,13 @@ func DiskSpaceShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
 
 func TagsShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
 	return len(new.(*schema.Set).List()) != 0
+}
+
+func CustomizeDiffServiceIntegrationAfterCreation(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	if len(d.Id()) > 0 && d.HasChange("service_integrations") && len(d.Get("service_integrations").([]interface{})) != 0 {
+		return fmt.Errorf("service_integrations field can only be set during creation of a service")
+	}
+	return nil
 }
 
 func CustomizeDiffCheckUniqueTag(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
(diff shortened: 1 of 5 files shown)


internal/schemautil/custom_diff.go
```diff
@@ -153,3 +153,49 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
 	// TODO: Check that we block deletions that will result in too few static ips for the plan
 	return nil
 }
+
+// CustomizeDiffValidateFlinkJob is a custom diff validation function for flink jobs
+func CustomizeDiffValidateFlinkJob(_ context.Context, d *schema.ResourceDiff, m interface{}) error {
```
ivan-savciuc, line 158: I think we can move a `CustomizeDiffValidateFlinkJob` to the Flink package, since we are not using it anywhere else.

internal/schemautil/custom_diff.go
```diff
@@ -153,3 +153,49 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
...
+	})
+	if err != nil && aiven.IsNotFound(err) {
+		return nil
+	}
+
+	return err
+}
+
+// CustomizeDiffValidateFlinkTable is a custom diff validation function for flink tables
+func CustomizeDiffValidateFlinkTable(_ context.Context, d *schema.ResourceDiff, m interface{}) error {
```
ivan-savciuc, line 175: same here I think, perhaps we can move it to the `flink` package?

internal/schemautil/custom_diff.go
```diff
@@ -153,3 +153,49 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
...
+func CustomizeDiffValidateFlinkJob(_ context.Context, d *schema.ResourceDiff, m interface{}) error {
+	client := m.(*aiven.Client)
+
+	projectName, serviceName := d.Get("project").(string), d.Get("service_name").(string)
+
+	_, err := client.FlinkJobs.Validate(projectName, serviceName, aiven.ValidateFlinkJobRequest{
+		Statement: d.Get("statement").(string),
+		TableIDs:  FlattenToString(d.Get("table_ids").([]interface{})),
+	})
+	if err != nil && aiven.IsNotFound(err) {
```
ivan-savciuc, line 167: Can you add a comment here regarding why we ignore 404?

ivan-savciuc (changes requested):
A proposal, can we use something like:
```go
CustomizeDiff: customdiff.If(func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
	return len(d.Id()) > 0 // so that we ignore this custom diff when we do create a resource
}, schemautil.CustomizeDiffValidateFlinkTable),
```

Or if we care about a particular field we can explore options to use `IfValueChange`

ivan-savciuc (commented):
Can we add an acceptance test that checks these new validators? A false position one, when `client.FlinkTables.Validate` retrieves an error.

Serpentiel:
> A proposal, can we use something like:
> 
> ```go
> CustomizeDiff: customdiff.If(func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
> 	return len(d.Id()) > 0 // so that we ignore this custom diff when we do create a resource
> }, schemautil.CustomizeDiffValidateFlinkTable),
> ```

I believe this adds more code complexity and is unobvious in general

internal/schemautil/flink.go
```diff
@@ -0,0 +1,27 @@
+package schemautil
+
+import (
+	"github.com/aiven/aiven-go-client"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func FlinkReadUpsertKafkaFromSchema(d ResourceStateOrResourceDiff) *aiven.FlinkTableUpsertKafka {
```
ivan-savciuc, line 8: Why do we need to move `FlinkReadUpsertKafkaFromSchema` from the `flink` package since we only use it there?

internal/schemautil/custom_diff.go
```diff
@@ -28,6 +25,13 @@ func TagsShouldNotBeEmpty(_ context.Context, _, new, _ interface{}) bool {
 	return len(new.(*schema.Set).List()) != 0
 }
 
+func CustomizeDiffServiceIntegrationAfterCreation(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
```
ivan-savciuc, line 28: What `CustomizeDiffServiceIntegrationAfterCreation` has to do with this PR?

ivan-savciuc (changes requested):
Just a few comments, otherwise LGTM

internal/schemautil/flink.go
Serpentiel: whoops, sorry, my mistake, I didn't intend to

internal/schemautil/custom_diff.go
Serpentiel: I just rearranged them cause we have `CustomizeDiffFunc`s mixed with a bunch of `ValueChangeConditionFunc`s, so it makes sense to group them

ivan-savciuc (approved):
LGTM

PR #817, opened 2022-07-20
update account authentication SAML fields

Serpentiel:

- updates account authentication SAML fields

## Why this way

to match current API


ivan-savciuc (approved):
LGTM

PR #818, opened 2022-07-20
drop deprecated resources from sample project

Serpentiel:

drops deprecated resources from sample project

## Why this way

to make sample project up to date


ivan-savciuc (approved):
LGTM

PR #819, opened 2022-07-25
add outputs example to the sample project

### LINKED ISSUE #507: Add outputs to the sample project
The sample project is a great way to try out a bunch of services together, would be nice to have the outputs (like connection string. username, password) so the user knows where to connect to.

Serpentiel:

adds outputs example to the sample project

resolves #507

## Why this way

to improve documentation


ivan-savciuc (approved):
LGTM

Issue #798, opened 2022-07-07
Truncated user agent in logs


st3fan:
## What happened?

In internal Aiven logs we see many calls being made by the terraform provider that have a `User-Agent` that has the version missing.

```
"terraform-provider-aiven/"
```

There is no other indication of the version of the provider used so it is not clear if these are old versions that may not have set the `User-Agent` correctly.

Looking at the current code, I see two things:

https://github.com/aiven/terraform-provider-aiven/blob/121a6e152d9431a66deb4b4cf6664bb5902200e9/internal/provider/provider.go#L244-L246

And:

https://github.com/aiven/terraform-provider-aiven/blob/121a6e152d9431a66deb4b4cf6664bb5902200e9/internal/sweep/sweep.go#L30

Assuming the second parameter to `NewTokenClient` will transfer to the `User-Agent` header - the first one looks good but the second one has the version missing I think? Although that one `-acc` is not what we see in log files.

mhoffm-aiven:
might be old provider too

ivan-savciuc:
We should definitely send an actual provider version in the User-Agent header.

mhoffm-aiven:
Lets use the old linker trick to override the version when we compile releases

st3fan:
@Serpentiel It is very possible that we are seeing older versions being used in the current logs. 

We just see the HTTP requests in the logs and they do not have enough context to find out who or what version is sending those requests.

If you think this is all working as expected in the current shipping release then please feel free to close this issue.

Serpentiel:
> It is very possible that we are seeing older versions being used in the current logs.

could it be that we have a bug in the way it's being logged?

could you use spin the sample project up via Terraform and see what `User-Agent` is logged for the purpose of verifying that the logs aren't broken?

st3fan:
> could you use spin the sample project up via Terraform and see what `User-Agent` is logged for the purpose of verifying that the logs aren't broken?

I don't know how to do this - but maybe we can work on this together? If you spin something up and I know for example the IP address from which these requests are coming, I can follow along in the logs to see what things look like.

Serpentiel:
after an internal discussion with @st3fan we found out that the issue doesn't exists in the current version of Terraform provider

PR #824, opened 2022-07-27
fix version ldflag

Serpentiel:

fixes version ldflag

introduced in #798

## Why this way

currently, our `User-Agent` string contains misleading `dev` version, and this is not correct

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -32,7 +32,7 @@ import (
 )
 
 var (
-	providerVersion = "dev"
+	version = "dev"
 )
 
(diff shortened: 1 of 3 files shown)


internal/provider/provider.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 var (
-	providerVersion = "dev"
+	version = "dev"
```
Serpentiel, line 35: this has been made to follow the ideology of Go, which forbids to prefix variables with package name

ivan-savciuc (approved):
LGTM

PR #825, opened 2022-07-27
improve Makefile

Serpentiel:

- Beautifies and reworks `Makefile`
- Adds `make` targets `build` and `build-dev`

## Why this way

to improve the `Makefile`


ivan-savciuc (approved):
LGTM

PR #829, opened 2022-07-28
use Changelog Enforcer GitHub Action

Serpentiel:

changes changelog enforcement action to a better, more supported one

## Why this way

we better rely on better tools


ivan-savciuc (approved):
LGTM

PR #830, opened 2022-07-28
capitalize variables in sample project

Serpentiel:

capitalizes variables in sample project

## Why this way

if you want to override those variables via environment variables, you have to use the following syntax:

```bash
export TF_VAR_<name>=<value>
```

and having them uncapitalized makes it end up looking:

```bash
export TF_VAR_aiven_api_token=<value>
export TF_VAR_aiven_project_name=<value>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ nav_order: 1
 - Add `make` targets `build` and `build-dev`
 - Add sweeper for account authentication
 - Make use of `BUILD_DEV_DIR` in `Makefile`
+- Capitalize variables in sample project
 
 ## [3.4.0] - 2022-07-26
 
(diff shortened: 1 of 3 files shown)


ivan-savciuc (commented):
Yes indeed it has to be:
```
export TF_VAR_aiven_api_token=<value>
export TF_VAR_aiven_project_name=<value>
```

And this is perfectly in unison with the Terraform naming convention https://www.terraform-best-practices.com/naming#general-conventions

Serpentiel:
> And this is perfectly in unison with the Terraform naming convention https://www.terraform-best-practices.com/naming#general-conventions

but this isn't unison with IEEE Std 1003.1-2017, see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

ivan-savciuc:
env variables have nothing to do with TF variables

Serpentiel:
> env variables have nothing to do with TF variables

they let us manipulate them via env variables, thus they should follow the rules of the env variables for that case

but I guess I'll create an issue to hashicorp/terraform then

PR #831, opened 2022-07-28
fix Makefile test-acc command

ivan-savciuc:

Fix `test-acc` command when user is using PGK


Serpentiel (approved):
lgtm

PR #832, opened 2022-07-28
fix service project vpc custom diff

ivan-savciuc:
fix service project vpc custom diff

## Why this way
do not error when id is empty or not set by a user


ivan-savciuc:
this is hot fix, no change log

Serpentiel (approved):
lgtm

PR #835, opened 2022-07-28
add dependabot.yml

Serpentiel:

adds `dependabot.yml` for Dependabot config

## Why this way

life is awesome when there's no vulnerable code and the dependencies are up to date


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily
+  - package-ecosystem: gomod
+    directory: /
```
ivan-savciuc, line 8: pls add:
```
 ignore:
      - dependency-name: "golang.org/x/tools"
      - dependency-name: "google.golang.org/grpc"
```

ivan-savciuc (changes requested):
just a small thing to add, otherwise LGTM

ivan-savciuc (approved):
LGTM

PR #839, opened 2022-07-29
prevent changelog enforcement for PRs from dependabot[bot]

Serpentiel:

prevents changelog enforcement for PRs from dependabot[bot]

## Why this way

we don't want to have changelog entries for dependency bumps


ivan-savciuc (approved):
LGTM

PR #842, opened 2022-08-01
update Changelog Enforcer workflow

Serpentiel:
delete me

updates Changelog Enforcer workflow

## Why this way

it has built-in functionality for labels that skip changelog enforcement
also we use a different version pattern in changelog, which is different from what [KeepAChangelog](https://keepachangelog.com) offers


mhoffm-aiven (approved):
lgtm

PR #843, opened 2022-08-01
add Dependency Review workflow

Serpentiel:
delete me

add Dependency Review workflow

## Why this way

to prevent usage of outdated dependencies


mhoffm-aiven:
maybe rather a periodic workflow to not fail PRs because some dependency got updated upstream?

Serpentiel:
> maybe rather a periodic workflow to not fail PRs because some dependency got updated upstream?

we have Dependabot configured for this reason now, the point of this is to fail PRs that don't use latest dependencies, if there were new dependencies used within the PR

mhoffm-aiven (approved):
lgtm

PR #845, opened 2022-08-01
add CodeQL workflow

Serpentiel:

adds CodeQL workflow

## Why this way

to have CodeQL analysis :)


mhoffm-aiven (approved):
lgtm

PR #849, opened 2022-08-03
add PostgreSQL max connections and PgBouncer outputs

Serpentiel:

adds PostgreSQL max connections and PgBouncer outputs

## Why this way

to make `sample_project` better, to let users have connection limit information available

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,13 @@
 .PHONY: build build-dev test test-unit test-acc lint lint-go lint-test lint-docs fmt fmt-test docs clean clean-tools sweep
 
-GO := CGO_ENABLED=0 go
+#################################################
+# Global
+#################################################
 
-#################################################
-# Tools
-#################################################
+GO := CGO_ENABLED=0 go
 
 TOOLS_DIR ?= tools
 TOOLS_BIN_DIR ?= $(TOOLS_DIR)/bin
-
 
 $(TOOLS_BIN_DIR):
 	mkdir -p $(TOOLS_BIN_DIR)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -31,19 +30,24 @@ GOLANGCILINT := $(TOOLS_BIN_DIR)/golangci-lint
...
 ARCH ?= $(shell $(GO) env GOOS GOARCH | tr '\n' '_' | sed '$$s/_$$//')
-BUILD_DEV_DIR ?= ~/.terraform.d/plugins/registry.terraform.io/aiven/aiven/0.0.0+dev/$(ARCH)
-
+BUILD_DEV_DIR ?= ~/.terraform.d/plugins/registry.terraform.io/aiven-dev/aiven/0.0.0+dev/$(ARCH)
 
 $(BUILD_DEV_DIR):
 	mkdir -p $(BUILD_DEV_DIR)
 
 
+PROJECT_DIR ?= sample_project
```
Serpentiel, line 42: changeable via env variables

Makefile
```diff
@@ -53,14 +57,28 @@ build:
...
+#################################################
+
+up: project-init project-apply
+
+
+project-init: $(PROJECT_DIR)
+	cd $(PROJECT_DIR) && terraform init -input=false
+
+
+project-apply: $(PROJECT_DIR)
```
ivan-savciuc, line 79: Trying this :
```
) make project-apply
cd sample_project && terraform apply -input=false -auto-approve
╷
│ Error: Unsupported attribute
│ 
│   on outputs.tf line 223, in output "samplepg_max_connections":
│  223:   value       = aiven_pg.samplepg.pg[0].max_connections
│ 
│ This object has no argument, nested block, or exported attribute named "max_connections".
╵
make: *** [Makefile:80: project-apply] Error 1
```

Makefile
```diff
@@ -53,14 +57,28 @@ build:
...
+#      source  = "aiven-dev/aiven"
 #      version = "0.0.0+dev"
 #    }
 #  }
 #}
 build-dev: $(BUILD_DEV_DIR)
 	$(GO) build -o $(BUILD_DEV_DIR)/terraform-provider-aiven_v0.0.0+dev
 
+#################################################
+# Up
```
ivan-savciuc, line 69: What is the point of this block? I do not know how we can benefit of creating a folder and running TF apply as make commands

internal/schemautil/service.go
```diff
@@ -624,7 +624,7 @@ func copyServicePropertiesFromAPIResponseToTerraform(
 		return fmt.Errorf("cannot set `components` : %s", err)
 	}
 
-	return copyConnectionInfoFromAPIResponseToTerraform(d, serviceType, s.ConnectionInfo)
+	return copyConnectionInfoFromAPIResponseToTerraform(d, serviceType, s.ConnectionInfo, s.Metadata)
```
ivan-savciuc, line 627: Can we add a test that checks that we get this property set?

ivan-savciuc (changes requested):
Please add a test

Makefile
Serpentiel: it works for me

internal/schemautil/service.go
Serpentiel: `s.Metadata` always exists in every API response of this type, it also always exists in the Go client, is there really need for a test?

Makefile
Serpentiel: it could be used for quick spin up of any Terraform project, and is useful for development

internal/schemautil/service.go
ivan-savciuc: I mean we can just add this check to one of the existing tests

ivan-savciuc (approved):
LGTM

ivan-savciuc (approved):

PR #852, opened 2022-08-04
fix code warnings and errors

Serpentiel:

fixes code warnings and errors

there's going to be a bunch of clean up PRs, I will add a changelog entry for it with the last PR

## Why this way

to improve code quality

### DIFF
--- a/internal/schemautil/user_config.go
+++ b/internal/schemautil/user_config.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil/templates"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
 // GenerateServiceUserConfigurationSchema generate service user_config
@@ -20,7 +22,7 @@ func GenerateServiceUserConfigurationSchema(t string) *schema.Schema {
 		Type:             schema.TypeList,
 		MaxItems:         1,
 		Optional:         true,
-		Description:      fmt.Sprintf("%s user configurable settings", strings.Title(t)),
+		Description:      fmt.Sprintf("%s user configurable settings", cases.Title(language.Und, cases.NoLower).String(t)),
 		DiffSuppressFunc: EmptyObjectDiffSuppressFuncSkipArrays(s),
 		Elem:             &schema.Resource{Schema: s},
 	}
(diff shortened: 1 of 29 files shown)


internal/schemautil/user_config.go
```diff
@@ -20,7 +22,7 @@ func GenerateServiceUserConfigurationSchema(t string) *schema.Schema {
 		Type:             schema.TypeList,
 		MaxItems:         1,
 		Optional:         true,
-		Description:      fmt.Sprintf("%s user configurable settings", strings.Title(t)),
+		Description:      fmt.Sprintf("%s user configurable settings", cases.Title(language.Und, cases.NoLower).String(t)),
```
ivan-savciuc, line 25: it will make all words in the description field start with the capital, we will have something like `Allow incoming connections from ...` -> `Allow Incoming Connections From ....` 

Lets not alter the description field in this way.

internal/service/kafka/kafka_topic_create.go
```diff
@@ -9,19 +9,19 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
-// KafkaTopicCreateWaiter is used to create topics. Since topics are often
+// TopicCreateWaiter is used to create topics. Since topics are often
```
ivan-savciuc, line 12: I assume we are trying to remove the package name from the name of the struct, but it alters its meaning. The TopicWaiter can be a little misleading, and I suggest we keep Kafka in the name, but let's not export it any longer, and I think liners will be kind to it.

ivan-savciuc (changes requested):
A few small things

internal/schemautil/user_config.go
ivan-savciuc: Checked the code from above, we only pass a a service name here as `t`, so we are good to go!

internal/service/kafka/kafka_topic_create.go
Serpentiel: I'm sorry, but it's forbidden to prefix methods in Go with it's package name, we will have to rename the package otherwise, see https://go.dev/blog/package-names:

> Avoid repetition. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.

ivan-savciuc: if we have this as an internal structure it is going to be OK

internal/schemautil/user_config.go
Serpentiel: `strings.Title` does that already anyway

ivan-savciuc (dismissed):
LGTM

ivan-savciuc:
Pls fix unit tests we we merge this one

ivan-savciuc (approved):
LGTM

PR #855, opened 2022-08-05
update Flink version to 1.15

juha-aiven:

The Flink version in Aiven for be updated to 1.15. The support for 1.13 will be removed. 

Change all references to Flink 1.15.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ nav_order: 1
 
 # Changelog
 
+## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
+
+- Change Flink version to 1.15
+
 ## [3.6.0] - 2022-08-31
 
 - Add additional disk space support
--- a/internal/schemautil/templates/service_user_config_schema.json
+++ b/internal/schemautil/templates/service_user_config_schema.json
@@ -736,7 +736,7 @@
       },
       "flink_version": {
         "enum": [
-          "1.13"
+          "1.15"
         ],
         "title": "Flink major version",
         "type": [
--- a/docs/resources/flink.md
+++ b/docs/resources/flink.md
@@ -22,7 +22,7 @@ resource "aiven_flink" "flink" {
   maintenance_window_time = "10:00:00"
 
   flink_user_config {
-    flink_version = 1.13
+    flink_version = 1.15
   }
 }
 ```
--- a/examples/resources/aiven_flink/resource.tf
+++ b/examples/resources/aiven_flink/resource.tf
@@ -7,6 +7,6 @@ resource "aiven_flink" "flink" {
   maintenance_window_time = "10:00:00"
 
   flink_user_config {
-    flink_version = 1.13
+    flink_version = 1.15
   }
 }
--- a/internal/schemautil/templates/service_user_config_schema.go
+++ b/internal/schemautil/templates/service_user_config_schema.go
@@ -746,7 +746,7 @@ var (
       },
       "flink_version": {
         "enum": [
-          "1.13"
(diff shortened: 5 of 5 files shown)


mhoffm-aiven (dismissed):
lgtm

juha-aiven:
This can be / needs to be merged _after_ the change of the Flink version has been done in Aiven prod.

mhoffm-aiven:
> This can be / needs to be merged _after_ the change of the Flink version has been done in Aiven prod.

understood, can you convert to draft until its in?

ivan-savciuc (approved):
LGTM

internal/schemautil/templates/service_user_config_schema.json
```diff
@@ -685,7 +685,7 @@
       },
       "flink_version": {
         "enum": [
-          "1.13"
+          "1.15"
```
ivan-savciuc, line 688: Oh wait, here we should preserve the old version as well, so it should be:
```
 "enum": [
          "1.13",
          "1.15"

**Review by ivan-savciuc — CHANGES_REQUESTED** [review]
We should keep an old version

File: internal/schemautil/templates/service_user_config_schema.json (hunk 1 above)
Comment (juha-aiven) [reply internal/schemautil/templates/service_user_config_schema.json:688]: We don't currently supported different versions of Flink in Aiven. Thus when Flink 1.15.1 goes live in Aiven, all new services will be 1.15.1.

Is there some reason why 1.13 should be preserved here?

Comment (ivan-savciuc) [reply internal/schemautil/templates/service_user_config_schema.json:688]: When a user upgrades the TF provider and uses 1.13, she will get an error that this value is no longer valid, this is why we only remove old values only during the major releases. 

Are you saying it is impossible that we have any users on 1.13?

**Comment by juha-aiven** [comment]
> When a user upgrades the TF provider and uses 1.13, she will get an error that this value is no longer valid

Let me find out the exact way the upgrade will happen, and I'll come back to this.

**Comment by juha-aiven** [comment]
Flink 1.15 is now live, 1.13 Flink is no longer available for new services. 

Screenshot from 2022-09-01 08-17-26

File: internal/schemautil/templates/service_user_config_schema.json (hunk 1 above)
Comment (juha-aiven) [reply internal/schemautil/templates/service_user_config_schema.json:739]: I think we can ignore this problem for now.

There might still be some users running 1.13, but they'll switch to 1.15 sooner than later. And new Flink services can only be created using 1.15.

**Comment by juha-aiven** [comment]
I did not manually test this. Was there a test somewhere that actually creates a Flink service in aiven.io - should we try out that?

**Comment by Serpentiel** [comment]
are those versions backwards compatible?

**Comment by juha-aiven** [comment]
> are those versions backwards compatible?

They are not. But we'll fix the prod issues related to the incompatibility if those happen.

**Review by Serpentiel — APPROVED** [review]
well, there's nothing really we can do about it if it's already unsupported in prod
we need to have this working in Terraform anyway, even if the old version won't work
so lgtm

File: CHANGELOG.md [hunk 2]
```diff
@@ -7,6 +7,10 @@ nav_order: 1
 
 # Changelog
 
+## Unreleased
```
Serpentiel, line 10: please change the placeholder to be `[MAJOR.MINOR.PATCH] - YYYY-MM-DD`

Serpentiel (dismissed):
small change, otherwise lgtm

CHANGELOG.md
juha-aiven: Done.

Serpentiel (dismissed):
lgtm

Serpentiel (approved):

PR #856, opened 2022-08-05
small .golangci.yml fix

Serpentiel:

- dropped unused exclusion rules
- rearranged linters alphabetically
- drop timeout setting, the linter never takes more than several seconds to run—default is 30 minutes
- set linters to believe that we use Go 1.17, see https://github.com/golangci/golangci-lint/issues/2649
- replace `vet` with `govet`

there's going to be a bunch of clean up PRs, I will add a changelog entry for it with the last PR

## Why this way

to improve `.golangci.yml`

### DIFF
--- a/internal/schemautil/common.go
+++ b/internal/schemautil/common.go
@@ -10,14 +10,14 @@ import (
 )
 
 //goland:noinspection GoDeprecation
-func GetACLUserValidateFunc() schema.SchemaValidateFunc {
+func GetACLUserValidateFunc() schema.SchemaValidateFunc { //nolint:staticcheck
 	return validation.StringMatch(
 		regexp.MustCompile(`^[-._*?A-Za-z0-9]+$`),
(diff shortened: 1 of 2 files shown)


internal/schemautil/common.go
```diff
@@ -10,14 +10,14 @@ import (
 )
 
 //goland:noinspection GoDeprecation
-func GetACLUserValidateFunc() schema.SchemaValidateFunc {
+func GetACLUserValidateFunc() schema.SchemaValidateFunc { //nolint:staticcheck
```
Serpentiel, line 13: this says that we use deprecated `schema.SchemaValidateFunc`

I believe we shouldn't mute deprecation issues

Serpentiel, line 13: this says that we use deprecated schema.SchemaValidateFunc

I believe we shouldn't mute deprecation issues

ivan-savciuc (approved):
LGTM

PR #861, opened 2022-08-10
fix release artifacts workflow

Serpentiel:

fixes release artifacts workflow

## Why this way

it was broken


ivan-savciuc (approved):
LGTM

PR #865, opened 2022-08-11
fix title case in unit-tests.yml

Serpentiel:

fixes title case in `unit-tests.yml`, so that the name of this workflow is unified with all other workflows

## Why this way

to unify the naming with other workflows


ivan-savciuc (approved):
LGTM

PR #867, opened 2022-08-15
fix vpc peering connection custom diff

ivan-savciuc:

Fixes vpc peering connection custom diff logic

## Why this way

During the plan we do not know the project vpc id therefore custom diff logic should be moved to the create method


Serpentiel (approved):
lgtm, but we really need to work on adding tests here

rslinckx:
It seems the GCP variant got forgotten in this fix ? https://github.com/aiven/terraform-provider-aiven/blob/8fdb15eef39d0ced492087dd57d827458d063af5/internal/service/vpc/resource_gcp_vpc_peering_connection.go#L69

I can't create a new gcp vpc peering with 3.6.0

Serpentiel:
@ivan-savciuc

PR #876, opened 2022-08-30
remove retention_hours from the kafka topic resource

ivan-savciuc:

retention_hours field is deprecated in v2 and gone from the resource in v3, and the code that sets this value can be deleted.


Serpentiel (approved):
lgtm

PR #877, opened 2022-08-30
update disk space on refresh

ivan-savciuc:

Disk space is not updated during the TF refresh/read, and it may cause some discrepancy between TF state and remove the Aiven state of the resource


Serpentiel (approved):
might be worth to fix typo in 1add162 by replacing `disc` with `disk`, otherwise lgtm

ivan-savciuc:
@Serpentiel updated commit message, pls check

PR #880, opened 2022-08-31
Update official provider name

dewan-ahmed:

The public Terraform Registry docs included some instances of "Aiven Terraform Provider" whereas the official name of the product is "Aiven Provider for Terraform".

This PR updates those texts as well as some outdated docs link for creating authentication tokens.

## Why this way

This is the only way to make a docs change.

## Note to the reviewer

I didn't make any change to `docs/` directory since I believe that the docs are auto-generated from the template. Please review if I made changes to the correct place/directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ nav_order: 1
 
 # Changelog
 
-## [X.X.X] - not yet released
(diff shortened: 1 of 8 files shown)


Serpentiel (changes requested):
please run `make docs`

Serpentiel (approved):
lgtm

PR #882, opened 2022-09-02
feat: extract exp tool output via jsongo

Serpentiel:

adds automatic extraction of exp tool output via jsongo and `go generate`

## Why this way

this is a first step towards automatic user configurable options schema generation


byashimov (approved):

Serpentiel:
will be implemented in https://github.com/aiven/aiven-go-client

PR #884, opened 2022-09-05
add is_account_owner computed field

byashimov:

Add support is_account_owner field https://api.aiven.io/doc/#operation/AccountList


Serpentiel (approved):
lgtm, next time also run `go mod tidy` please
we should probably add `go mod tidy` check in our linter

PR #891, opened 2022-09-09
update examples, set v3 for provider

byashimov:

Updates examples with v3 provider version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,15 @@ nav_order: 1
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
+## [3.6.1] - 2022-09-05
+
 - Update to the official name Aiven Provider for Terraform
 - Replaced older links from help.aiven.io to docs.aiven.io
 - Change Flink version to 1.15
 - Fix empty `user_config` and empty `ip_filters`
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
```diff
@@ -9,15 +9,15 @@ nav_order: 1
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
+## [3.6.1] - 2022-09-05
+
```
Serpentiel, line 13: please move all these changes under `[MAJOR.MINOR.PATCH] - YYYY-MM-DD` section

Serpentiel, line 13: please move all these changes under `[MAJOR.MINOR.PATCH] - YYYY-MM-DD` section
we add version number when we're ready to release

Serpentiel (changes requested):
there might be some chance that these examples won't exactly work with v3.x of our provider

also, at least one of these resources should be removed, that is the `elasticsearch` resource, and we should probably update it to use `opensearch`

we should probably discuss this a bit more before proceeding next

ivan-savciuc (changes requested):
Just a minor thing, can you put a space in between `>=` and the version?

ivan-savciuc:
@byashimov can you check comments for this PR please?

PR #894, opened 2022-09-13
feat: user config options schema generation

Serpentiel:

adds user config options schema generation

Dependabot will now automatically update our schema whenever new options become available

## Why this way

Aiven users are going to be happy because of our new fast time-to-market approach in terms of user config options availability :)


Serpentiel:
acceptance tests:

https://github.com/aiven/terraform-provider-aiven/actions/runs/3045945001

byashimov (approved):

PR #897, opened 2022-09-15
ci(codeowners): change team

Serpentiel:

changes team in `CODEOWNERS` file

## Why this way

GitHub doesn't supports multiple `CODEOWNERS` entries so we decided to create a new team for it


mhoffm-aiven (approved):
lgtm

PR #898, opened 2022-09-19
feat: user config static schema generator, objects only

Serpentiel:

adds a user config static schema generator, without support for arrays at the moment

## Why this way

to simplify debugging, to improve code readability, to keep user config options up-to-date automatically and let the community to also be able to easily watch for changes / propose changes in our user config options


internal/schemautil/user_config/service_types.go
```diff
@@ -0,0 +1,1419 @@
...
+// ServiceTypeOpensearch is a generated function returning the schema of the opensearch ServiceType.
+func ServiceTypeOpensearch() *schema.Schema {
+	s := map[string]*schema.Schema{
+		"custom_domain": {
+			Description: "Serve the web frontend using a custom CNAME pointing to the Aiven DNS name",
+			Optional:    true,
+			Required:    false,
+			Type:        schema.TypeString,
+		},
+		"disable_replication_factor_adjustment": {
```
ivan-savciuc, line 19: I see a lot of `deprecated` words in the description. Can we add a logic that checks the description, and in case it contains a deprecated word, we also make this schema field deprecated, WDYT?

internal/schemautil/user_config/service_types.go
```diff
@@ -0,0 +1,1419 @@
...
+			Required:    true,
+			Type:        schema.TypeBool,
+		},
+		"ip_filter_map": {
+			Description: "Allow incoming connections from CIDR address block, e.g. '10.20.0.0/16'",
+			Optional:    false,
+			Required:    true,
+			Type:        schema.TypeMap,
+		},
+		"ip_filter_string": {
```
ivan-savciuc, line 354: It is best we keep an existing fields like `ip_filter` intact, without the `_string` postfix.

internal/schemautil/user_config/service_types.go
```diff
@@ -0,0 +1,1419 @@
...
+		},
+		"migration": {
+			Optional: true,
+			Required: false,
+			Type:     schema.TypeMap,
+		},
+		"pg": {
+			Optional: false,
+			Required: true,
+			Type:     schema.TypeMap,
```
ivan-savciuc, line 368: Are we going to expand these nested fields with the schema.Set?

internal/schemautil/user_config/service_types.go
```diff
@@ -0,0 +1,1419 @@
...
+		"max_index_count": {
+			Description: "Maximum number of indexes to keep before deleting the oldest one",
+			Optional:    false,
+			Required:    true,
+			Type:        schema.TypeInt,
+		},
+		"opensearch_version": {
+			Optional: true,
+			Required: false,
+			Type:     schema.TypeString,
```
ivan-savciuc, line 581: We should have a validation to the fields that have it.

internal/schemautil/user_config/util.go
```diff
@@ -0,0 +1,70 @@
...
+			r = append(r, "TypeBool")
+		case "integer":
+			r = append(r, "TypeInt")
+		case "number":
+			r = append(r, "TypeFloat")
+		case "string":
+			r = append(r, "TypeString")
+		case "array":
+			r = append(r, "TypeSet")
+		case "object":
```
ivan-savciuc, line 26: Object should be TypeSet I think

ivan-savciuc (commented):
Did a first quick read trough and lest some comments

internal/schemautil/user_config/service_types.go
Serpentiel: could you please elaborate?

internal/schemautil/user_config/service_types.go
Serpentiel: we already discussed this internally with you and yes, I will make it this way when the final version of this PR is ready

internal/schemautil/user_config/service_types.go
Serpentiel: could you please elaborate?

internal/schemautil/user_config/util.go
Serpentiel: this should be discussed additionally

ivan-savciuc (changes requested):
If you run `terraform providers schema --json` and take a close look for example on `kafka_user_config.kafka` object since we use `TypeMap` it generates the following: 
```json
"kafka": {
  "type": [
    "map",
    "string"
  ],
  "description": "Kafka broker configuration values",
  "description_kind": "plain",
  "optional": true
},
```
even thought we have this code generated: 
image

Objects in the schema should have either `TypeList` (like it was in the old version) or `TypeSet` ( I'm in favor of using this one), but instead we use `TypeMap` https://www.terraform.io/plugin/sdkv2/schemas/schema-types#typemap that is not generating a desired schema.

ivan-savciuc (changes requested):
After running `go generate` I get this: 
image
Newly generated code is scuffled around randomly, despite no change. Can you do something about it?

Serpentiel:
> Objects in the schema should have either `TypeList` (like it was in the old version) or `TypeSet` ( I'm in favor of using this one), but instead we use `TypeMap` https://www.terraform.io/plugin/sdkv2/schemas/schema-types#typemap that is not generating a desired schema.

done



> Newly generated code is scuffled around randomly, despite no change. Can you do something about it?

fixed

ivan-savciuc (approved):
LGTM

PR #899, opened 2022-09-20
fix: static_ips schema type change from list to set

ivan-savciuc:

The TF plan, diff sometimes is shown incorrectly because the list type may generate diff when an order of elements has changed, set type will fix that


Serpentiel (approved):
lgtm

byashimov (approved):

PR #900, opened 2022-09-20
add postgres example test

byashimov:

Adds a test for a postgres example using terratest.


byashimov:
- [ ] remove binaries for postgres
- [ ] add CI tests
- [ ] add retries for aiven client

examples_test/utils.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aiven/aiven-go-client"
+	"github.com/gruntwork-io/terratest/modules/random"
+)
+
+func randNamef(f string) string {
```
ivan-savciuc, line 12:
```
examples_test/utils.go:16:6: `randName` is unused (deadcode)
func randName() string {
     ^
examples_test/utils.go:20:6: `newClient` is unused (deadcode)
func newClient(token string) (*aiven.Client, error) {
     ^
examples_test/utils.go:37:6: `newTFConfig` is unused (deadcode)
func newTFConfig(providerPath string) (string, error) {
     ^
examples_test/utils.go:12:6: func `randNamef` is unused (unused)
func randNamef(f string) string {
```

ivan-savciuc: Do we need this file?

ivan-savciuc (changes requested):
Some ideas:
- Let's check if the `AIVEN_TOKEN` and `AIVEN_PROJECT_NAME` environment vars are set and, if not, error with the meaningful message before running the tests. WDYT?
- When we run example tests, can we add some prefixes to a service name? Something like `ex-test`? So that we know these services are related to the tests and we can set our sweeper to look for the services with this prefix. 
- Not sure if it is a good idea, but perhaps it is better to keep tests in the example folder as we do with the standard Go tests.

Also I did run the test: 
```
--- PASS: TestExamplesTestSuite (0.00s)
    --- PASS: TestExamplesTestSuite/TestPostgresExample (676.46s)
PASS
``` 
:+1:

examples_test/utils.go
byashimov: > Let's check if the AIVEN_TOKEN and AIVEN_PROJECT_NAME environment vars are set and, if not, error with the meaningful message before running the tests. WDYT?

There is a test for that.

> When we run example tests, can we add some prefixes to a service name? Something like ex-test? 

I can do that.

> Do we need this file?

That file is used by the test. But test marked with tag and the file is not. I'll mark them both, hope that will help.

examples_tests/utils.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aiven/aiven-go-client"
+	"github.com/gruntwork-io/terratest/modules/random"
+)
+
+// randName Returns randomized name for a given format specifier randName("foo-%s")
+func randName(format string) string {
```
ivan-savciuc, line 13: Not sure if we really need this function, we can perhaps move it's content to the `randNameGen`

ivan-savciuc (approved):
LGTM 
Added a minor sugestion

examples_tests/utils.go
byashimov: That's a handy one for small tests with a single name, so you don't have to make a generator just for a one var.

byashimov:
rebased to main

PR #910, opened 2022-10-03
feat: add support for arrays with one_of property

Serpentiel:

adds support for arrays with `one_of` property

## Why this way

to make them supported too


ivan-savciuc (approved):
LGTM

Issue #901, opened 2022-09-21
aiven_project_vpc will be read during apply - breaks aiven_azure_vpc_peering_connection


dkirrane:
Trying to upgrade to the lastest `3.6.0` version and move to using the new `aiven_azure_vpc_peering_connection` resource.
This resource is failing during Terraform plan with:

```
╷
│ Error: Missing required argument
│
│   with module.aiven.module.aiven_peering[0].aiven_azure_vpc_peering_connection.this,
│   on ../../modules/aiven-peering/main.tf line 59, in resource "aiven_azure_vpc_peering_connection" "this":
│   59:   vpc_id                = data.aiven_project_vpc.vpc.id
│
│ The argument "vpc_id" is required, but no definition was found.****
```

If I temporarily remove this resource I can see that Terraform does not read `aiven_project_vpc` datasource during planning phase.
So this is likely causing the issue above:
```
  # module.aiven.module.aiven_peering[0].data.aiven_project_vpc.vpc will be read during apply
  # (depends on a resource or a module with changes pending)
 <= data "aiven_project_vpc" "vpc" {
      + cloud_name   = "azure-westus"
      + network_cidr = (known after apply)
      + project      = "project-xyz"
      + state        = (known after apply)
    }
```

```
terraform version
Terraform v1.2.9
on linux_amd64
+ provider registry.terraform.io/aiven/aiven v3.6.0
+ provider registry.terraform.io/hashicorp/azuread v2.28.1
+ provider registry.terraform.io/hashicorp/azurerm v3.23.0
```

```hcl
data "aiven_project_vpc" "vpc" {
  project    = "project-xyz"
  cloud_name = "azure-westus"
}

resource "aiven_azure_vpc_peering_connection" "this" {
  vpc_id                = data.aiven_project_vpc.vpc.id
  azure_subscription_id = data.azurerm_subscription.current.subscription_id
  peer_resource_group   = local.resource_group_name
  vnet_name             = local.vnet_name
  peer_azure_app_id     = data.azuread_service_principal.peering.application_id
  peer_azure_tenant_id  = data.azurerm_subscription.current.tenant_id

  depends_on = [
    azurerm_role_assignment.peering,
    azurerm_role_assignment.aiven
  ]
}
```

# other info
When I switch back to `2.7.5` - `aiven_vpc_peering_connection` doesn't suffer from this issue.

```
  # module.aiven[0].module.aiven_peering[0].aiven_vpc_peering_connection.this will be created
[...]

PR #912, opened 2022-10-04
fix: change aiven_project_vpc datasource

ivan-savciuc:

Change `aiven_project_vpc` data source behavior and rename `id` field to `vpc_id`

Resolves: #901.


Serpentiel (dismissed):
lgtm

internal/service/vpc/datasource_project_vpc.go
```diff
@@ -12,12 +16,45 @@ import (
...
-	aivenProjectVPCDataSourceSchema := schemautil.ResourceSchemaAsDatasourceSchema(aivenProjectVPCSchema,
-		"project", "cloud_name")
-	aivenProjectVPCDataSourceSchema["id"] = &schema.Schema{
-		Type:        schema.TypeString,
-		Description: "ID of the VPC. This can be used to filter out the specific VPC if there are more than one datasource returned.",
-		Optional:    true,
+	aivenProjectVPCDataSourceSchema := map[string]*schema.Schema{
+		"project": {
+			Type:          schema.TypeString,
+			ValidateFunc:  validation.StringMatch(regexp.MustCompile("^[a-zA-Z0-9_-]*$"), "project name should be alphanumeric"),
```
byashimov, line 22: nit: I would leave that for backend. Or go client if it's made from schema. This regexp covers basic validation, it even works with empty strings and leading dash/underscore or just dash-only. It doesn't cover length validation either.

internal/service/vpc/datasource_project_vpc.go
```diff
@@ -28,34 +65,30 @@ func DatasourceProjectVPC() *schema.Resource {
...
 	projectName := d.Get("project").(string)
 	cloudName := d.Get("cloud_name").(string)
-	vpcId, hasId := d.GetOk("id")
+	if id, hasId := d.GetOk("vpc_id"); hasId {
+		var err error
+		splitId, err = schemautil.SplitResourceID(id.(string), 2)
+		if err != nil {
+			return diag.Errorf("error splitting vpc_id: %s:", err)
+		}
+
```
byashimov, line 79: how about merging this validation and ValidateDiagFunc from the Schema?

internal/service/vpc/datasource_project_vpc.go
```diff
@@ -28,34 +65,30 @@ func DatasourceProjectVPC() *schema.Resource {
...
-				return diag.FromErr(err)
-			}
-
-			id := splitId[1]
-			condition = vpc.ProjectVPCID == id
-		}
-
-		if condition {
+		if vpc.CloudName == cloudName || (vpcId != "" && vpc.ProjectVPCID == vpcId) {
 			filteredVPCs = append(filteredVPCs, vpc)
```
byashimov, line 92: What if cloudName and vpc.CloudName are both empty? Is that possible? Or should it be validated separately:
```go
if cloudName == "" && vpcId == "" {
    // one must be provided
}

if vpc.CloudName == cloudName || vpc.ProjectVPCID == vpcId {
    filteredVPCs = append(filteredVPCs, vpc)
}
```

byashimov (dismissed):

internal/service/vpc/datasource_project_vpc.go
Serpentiel: this is supposed to prevent being rate limited by the backend and to reduce waiting time for the backend to report validation problem, and is actually supposed to be simple enough to cover most cases

backend is going to fail anyway with empty strings, so I believe this could be left as is

ivan-savciuc: As @Serpentiel mentioned we already have this validation on the back-end and this is done to communicate the correct formatting on the `terraform plan` stage.

Serpentiel (dismissed):
lgtm

byashimov (approved):

Serpentiel (approved):
lgtm

PR #917, opened 2022-10-05
fix account authentication method update, add tests

byashimov:

Fixes account authentication method update, adds tests.
Blocked by https://github.com/aiven/aiven-go-client/pull/204


internal/schemautil/schemautil.go
```diff
@@ -201,6 +201,10 @@ func IpFilterValueDiffSuppressFunc(k, old, new string, _ *schema.ResourceData) b
 	return old == "0.0.0.0/0" && new == "" && strings.HasSuffix(k, ".ip_filter.0")
 }
 
+func TrimSpaceDiffSuppressFunc(k, old, new string, _ *schema.ResourceData) bool {
+	return strings.TrimSpace(old) == strings.TrimSpace(new)
```
Serpentiel, line 205: is this really necessary?

byashimov: Yes. Backend returns certificate with trailing \n.

internal/service/account/resource_account_authentication_test.go
```diff
@@ -35,6 +46,109 @@ func TestAccAivenAccountAuthentication_basic(t *testing.T) {
 	})
 }
 
+func testAccAccountAuthenticationResourceSAMLCertificate(rName, certificate string) string {
```
ivan-savciuc, line 49: It seems to be that we can fit all the newly created tests into one test function with multiple steps, for example: 
`TestAccAivenAccountAuthentication_saml_valid_certificate` is basically the same with `TestAccAivenAccountAuthentication_saml_valid_certificate_update_delete` but we just have one extra `update` step.  

Moreover we no longer need the `TestAccAivenAccountAuthentication_basic` test so please drop it or move all the new tests inside.

ivan-savciuc (changes requested):
This is blocked now by the corresponding go-client PR.

Also had few tiny suggestions otherwise LGTM

internal/service/account/resource_account_authentication_test.go
byashimov: basic test doesn't have certificate file

internal/service/account/resource_account_authentication.go
```diff
@@ -39,9 +41,10 @@ var aivenAccountAuthenticationSchema = map[string]*schema.Schema{
 		Description: "Team ID",
 	},
 	"saml_certificate": {
-		Type:        schema.TypeString,
-		Optional:    true,
-		Description: "SAML Certificate",
+		Type:             schema.TypeString,
+		Optional:         true,
+		Description:      "SAML Certificate",
+		DiffSuppressFunc: schemautil.TrimSpaceDiffSuppressFunc,
```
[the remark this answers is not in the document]
byashimov: It said for DuffSuppressFunc:

> // This allows comparison based on something other than primitive, list
> // or map equality - for example SSH public keys may be considered
> // equivalent regardless of trailing whitespace.

I can change it for StateFunc I think they are equal here.

internal/schemautil/schemautil.go
Serpentiel: why can't we have it with the trailing `\n`?

byashimov: Meaning? it's compared with local one and doesn't match.  
So either you add trailing to one or trim it from another :)

Serpentiel: I'm sorry but according to [RFC 7468 ("Textual Encodings of PKIX, PKCS, and CMS Structures", section 3: ABNF)](https://rfc-editor.org/rfc/rfc7468#section-3), we can't trim the trailing whitespace at the very end of the string, could you please confirm that only the `\n` is being trimmed?

ivan-savciuc: `strings.TrimSpace` trims `\n`, I guess we should use https://golang.org/pkg/strings/#TrimLeft instead?

internal/service/account/resource_account_authentication_test.go
```diff
@@ -35,6 +46,109 @@ func TestAccAivenAccountAuthentication_basic(t *testing.T) {
...
+  type             = "saml"
+  name             = "test-acc-auth-method-%[1]s"
+  saml_certificate = <<-EOT
+  %[2]s
+  EOT
+}
+`, rName, certificate)
+}
+
+func TestAccAivenAccountAuthentication_saml_valid_certificate(t *testing.T) {
```
ivan-savciuc, line 66: I still this it is a good idea to combine `TestAccAivenAccountAuthentication_saml_valid_certificate` and `TestAccAivenAccountAuthentication_saml_valid_certificate_update` since `TestAccAivenAccountAuthentication_saml_valid_certificate` is a first step for the `TestAccAivenAccountAuthentication_saml_valid_certificate_update`

ivan-savciuc (commented):
WDYT about those test?

internal/schemautil/schemautil.go
Serpentiel: after discussing this internally, this has been concluded to be left as is

byashimov: The certificate is encapsulated between `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----`. Terraform's multiline markup is quite tricky and produces unwanted whitespaces https://www.terraform.io/language/expressions/strings#heredoc-strings. It's more user friendly just trim that out. Certificate won't be damaged. A real-world ACC test example with server side validation is attached to the PR.

PR #918, opened 2022-10-05
rename examples-tests job

byashimov:

Copy/pasted job name, should be renamed.


Serpentiel (approved):
lgtm

Issue #620, opened 2021-11-10
Wrong type rendering between schema and the docs


z0r1k:
Hi,

In the schema (https://github.com/aiven/terraform-provider-aiven/blob/c0182fce7e2af8869d5877031a2f82ec7704c04a/aiven/templates/service_user_config_schema.json#L4508-L4511) type is defined as list of `null` and `boolean`. 

I assume when property is not set default value is `null` and the main use-case is to have it set to `true` or `false` if one would use it. 
When docs are being generated aforementioned type is being rendered as `string` instead - https://github.com/aiven/terraform-provider-aiven/blob/c0182fce7e2af8869d5877031a2f82ec7704c04a/docs/resources/service.md#L948
This is a big misleading. 

It seems to happen to other properties as well. 

Ty.

ivan-savciuc:
@z0r1k, we know about this issue, all user configuration options have the type `string` in Terraform. We auto-generate all of them from the JSON schema you mentioned above. And our documentation is generated based on resource schema, which is why you get this issue. 

We plan to rework the generation of user configuration options which will automatically fix this issue.

Serpentiel:
after further evaluation, due to the deprecation mandate from Hashicorp, we will be unable to change those types in v3.x, and this will only be adjusted in v4.0, which is currently an on-going project of ours

Serpentiel:
v4 was released, and the docs were adjusted accordingly

PR #920, opened 2022-10-06
refactor: switch to new user config generator

Serpentiel:

switching to the new user config generator, completely removing the old one

## Why this way

to let it go :)


internal/service/service_integration/resource_service_integration_endpoint.go
```diff
@@ -37,116 +37,17 @@ var aivenServiceIntegrationEndpointSchema = map[string]*schema.Schema{
...
-		Description: "External AWS cloudwatch mertrics specific user configurable settings",
-		Elem: &schema.Resource{
-			Schema: schemautil.GenerateTerraformUserConfigSchema(
-				templates.GetUserConfigSchema("endpoint")["external_aws_cloudwatch_metrics"].(map[string]interface{})),
-		},
-		MaxItems: 1,
-		Optional: true,
-		Type:     schema.TypeList,
-	},
+	"datadog_user_config":                         dist.IntegrationEndpointTypeDatadog(),
```
Serpentiel, line 40: maybe we should consider also generating this so we can automatically add new integration endpoint types

Serpentiel:
temporarily converting this to draft as [acceptance tests are broken](https://github.com/aiven/terraform-provider-aiven/actions/runs/3198538292) at the moment

ivan-savciuc:
I think this PR fixes https://github.com/aiven/terraform-provider-aiven/issues/620, WDYT?

Serpentiel:
> I think this PR fixes #620, WDYT?

it does

Serpentiel:
acceptance tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/3330440281

ivan-savciuc (commented):
Some of the finding during tests: 
- By some reason M3DB `rules` object has `max_imtes:1` if one look at `terraform providers schema --json`
image Of course we should be able to have multiple rules but I cannot do something like : 
```
m3db_user_config {
  rules {...}
  rules {...}
```
- This code works: 
```     
  rules {
      mapping {
        name = "test1"
        filter = "__name__:disk_* host:important-42 mount:!*/sda"
        namespaces = [
          "aggregated_*"
        ]
      }
    }
```
but if I try `namespaces_object`, I'm getting this error: 
```
 rules {
      mapping {
        name = "test2"
        filter = "__name__:disk_* host:important-42 mount:!*/sda"
        namespaces_object {
          resolution = "1h"
          retention = "10h"
        }
      }
    }
...
```

ivan-savciuc:
Experimenting with namespaces_object, this code works: 
```     
  rules {
      mapping {
        name = "test1"
        filter = "__name__:disk_* host:important-42 mount:!*/sda"
        namespaces = [
          "aggregated_*"
        ]
      }
    }
```
but if I try `namespaces_object`, I'm getting this error: 
```
 rules {
      mapping {
        name = "test2"
        filter = "__name__:disk_* host:important-42 mount:!*/sda"
        namespaces_object {
          resolution = "1h"
          retention = "10h"
        }
      }
    }
...
error updating (ivans-m3db) service: 400: {"errors":[{"message":"Invalid 'user_config': Invalid input for rules: {'resolution': '1h', 'retention': '10h'} is not valid under any of the given schemas","status":400}],"message":"Invalid 'user_config': Invalid input for rules: {'resolution': '1h', 'retention': '10h'} is not valid under any of the given schemas"} - 
```

ivan-savciuc:
More experiments with `one_of` fields, if we try something like:
```
   ip_filter = ["200.1.2.3/32"]

    ip_filter_object {
      description = "test"
      network = "10.20.0.0/16"
    }
```
Only `200.1.2.3/32` filter will be added, expected two, or we should add `conflictingWith` to the schema so that a user cannot use both notation at the same time.

ivan-savciuc:
After successful creation of a service via TF if a user alters some of the complex structures via web console we get a diff during the plan, for example: 
```
  ~ resource "aiven_m3db" "m3db" {
        id                      = "aiven-ci-kubernetes-operator/ivans-m3db"
        # (21 unchanged attributes hidden)

      ~ m3db_user_config {
            # (5 unchanged attributes hidden)

          - ip_filter_object {
              - description = "sdfsdf" -> null
              - network     = "200.1.2.3/32" -> null
            }

          - namespaces {
              - name       = "ns10" -> null
              - resolution = "1d" -> null
              - type       = "aggregated" -> null

              - options {
                  - snapshot_enabled    = false -> null
                  - writes_to_commitlog = false -> null

                  - retention_options {
                      - blocksize_duration        = "30m" -> null
                      - retention_period_duration = "10h" -> null
                    }
                }
            }
```
But If a user updates any property of an existing namespace, we are not showing any diff. This is inconsistent with the behavior of user_config options in v2 and v3

ivan-savciuc (changes requested):
Can you please add comments and an explanation of what is happening with the schema or API request in `schemautil/userconfig`. It is tough to understand what is happening there and why. This component is a crucial part of the TF provider; we need it well-documented and understandable.

Also, we could add unit tests that validate all possible cases of conversations of a manifest payload to API request and viscus versa from API response back to a schema. 

Something like: 
From schema to API, if we can isolate if doe each type of schema properties: int, bool, string, object, array of objects:
- Given schema: ....
- Payload for the schema: ...
- Expected JSON of the request: ...

And the other way around: 
- JSON that we presume is coming from Aiven API
- Expected values in Schema also isolated for each time scenario

We need more unit tests for such a complex piece of engineering, perhaps even testing each function for the edge cases and, of course, more complex tests mentioned above.

internal/schemautil/userconfig/userconfig_test.go
```diff
@@ -11,16 +12,15 @@ import (
...
 	"golang.org/x/exp/slices"
-	"gopkg.in/yaml.v3"
 )
 
 // generateSchema is a function that generates Terraform schema via its map representation.
 func generateSchema(n string, m map[string]interface{}) {
 	np := fmt.Sprintf("%ss", n)
 
-	f := jen.NewFile("user_config")
+	f := jen.NewFile("dist")
```
byashimov, line 21: Let's move it to a param, on :76 too.

internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -0,0 +1,281 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var typedKeyRegexp = regexp.MustCompile(`^[A-z]*_(boolean|integer|number|string|array|object)$`)
```
byashimov, line 12: Just noticing this is not equal to `a-zA-Z_-`. It's shorter but has more cases also.

internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -0,0 +1,281 @@
...
+	d *schema.ResourceData,
+) (interface{}, bool) {
+	var res = make([]interface{}, len(v))
+
+	fks := strings.Join(fk, ".")
+
+	// TODO: Remove when this is fixed on backend.
+	if k == "additional_backup_regions" {
+		return res, true
+	}
```
byashimov, line 31: What fix it waits for? What `true` stands for?

byashimov (changes requested):
To me personally it's too much panics there. I would suggest to return errors instead so it can be unit-tested.

Serpentiel:
> By some reason M3DB `rules` object has `max_imtes:1` if one look at `terraform providers schema --json` image Of course we should be able to have multiple rules but I cannot do something like :

this should be filled as:

```hcl
    rules {
      mapping {
        name = "test1"
        filter = "__name__:disk_* host:important-42 mount:!*/sda"
        namespaces = [
          "aggregated_*"
        ]
      }

      mapping {
        name = "test1"
        filter = "__name__:disk_* host:important-42 mount:!*/sda"
        namespaces = [
          "aggregated_*"
        ]
      }
    }
```

> Only 200.1.2.3/32 filter will be added, expected two, or we should add conflictingWith to the schema so that a user cannot use both notation at the same time.

the `ConflictsWith` behavior is planned to be added in a separate PR

> panic: runtime error: index out of range [0] with length 0

this will be fixed with the new update of this PR

> But If a user updates any property of an existing namespace, we are not showing any diff. This is inconsistent with the behavior of user_config options in v2 and v3.

will be fixed

internal/schemautil/schemautil.go
```diff
@@ -134,7 +134,7 @@ func EmptyObjectDiffSuppressFunc(k, old, new string, _ *schema.ResourceData) boo
 
 	// When a field is not set to any value and consequently is null (empty string) but had
 	// a non-empty parameter before. Allow ignoring those.
-	if new == "" && old != "" {
+	if new == "" && old != "" || (!d.HasChange(k) && (new == "0" && old != "0" || new == "false" && old == "true")) {
```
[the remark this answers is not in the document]
Serpentiel: there is a comment at lines 135-136 :)

Serpentiel:
> To me personally it's too much panics there. I would suggest to return errors instead so it can be unit-tested.

panics are on purpose, if this code ever panics then our Terraform provider is broken, these errors can't be handled in any other way, safely, without affecting the overall functionality of the Terraform provider

in fact, the previous code we had in place there also works this way, cf. `internal/schemautil/user_config.go`@`main`

internal/schemautil/schemautil.go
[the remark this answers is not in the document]
Serpentiel: > Let's turn those strings to bool:

I believe these conversions might be a bit unnecessary

> And please provide some tests, this is very tricky.

the only way to test this is by acceptance testing this, as the only proper way to trigger this function is by actually spinning a service

in fact, the user config options convertor/generator is the most test-covered place in our code

internal/schemautil/userconfig/userconfig_test.go
Serpentiel: sorry, but why do we need this?

internal/schemautil/userconfig/apiconvert/toapi.go
Serpentiel: changed it to match `.*`

internal/schemautil/userconfig/apiconvert/toapi.go
Serpentiel: it waits for the fix of `additional_backup_regions`, `true` stands for omit

internal/schemautil/schemautil.go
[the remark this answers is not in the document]
ivan-savciuc: comment at lines 135-136 explain the `new == "" && old != "" ` part.
I suggest we add another if clause with `new == "0" && old != "0"` and explain that this is something for integer and float and `new == "false" && old == "true"` is related to boolean.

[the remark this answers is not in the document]
ivan-savciuc: Or at lest expand the commend with those cases, now it is a bit misleading because we are talking about string but have criteria for other types.

internal/schemautil/schemautil.go
```diff
@@ -119,22 +119,22 @@ func ParseOptionalStringToBool(val interface{}) *bool {
 }
 
 func CreateOnlyDiffSuppressFunc(_, _, _ string, d *schema.ResourceData) bool {
-	return len(d.Id()) > 0
+	return d.IsNewResource()
```
byashimov, line 122: Do we still need this? It looks like we can go with ForceNew.

Serpentiel: > Do we still need this? It looks like we can go with ForceNew.

we want to keep it the way it is in `main` at the moment

internal/schemautil/userconfig/userconfig_test.go
byashimov: Because it's a hardcoded config. If you try to write a test for this, you will face the problem that you can't use tmp or other dir for that.

Serpentiel: well then we will make it a param when this is needed, overengineering is generally a bad idea

ivan-savciuc (approved):
LGTM

PR #922, opened 2022-10-07
fix: update auth fields for sweep

byashimov:

Sweep refers to old and non-existing AuthenticationMethod fields.


ivan-savciuc (approved):
LGTM

PR #929, opened 2022-10-17
fix(ssg): return mirrormaker user config option

Serpentiel:

returns mirrormaker user config option—it will be dropped on next major release

also updates user config options with new options

## Why this way

for backwards compatibility


ivan-savciuc (approved):
LGTM

Issue #914, opened 2022-10-04
updating to 3.7.0 crashes


wanisfahmyDE:
Hi there,

When upgrading from version `3.6.0` to `3.7.0` the terraform provider crashes with the following:

```
The plugin.(*GRPCProvider).ReadResource request was cancelled.
╵

Stack trace from the terraform-provider-aiven_v3.7.0 plugin:

panic: interface conversion: interface {} is map[string]interface {}, not string

goroutine 49 [running]:
github.com/aiven/terraform-provider-aiven/internal/schemautil.toStringSlice(...)
	github.com/aiven/terraform-provider-aiven/internal/schemautil/ipfilter.go:84
github.com/aiven/terraform-provider-aiven/internal/schemautil.NormalizeIpFilter({0xc15b00?, 0xc00086de90?}, {0xc0002886b0?, 0x1, 0x1})
	github.com/aiven/terraform-provider-aiven/internal/schemautil/ipfilter.go:28 +0xccf
github.com/aiven/terraform-provider-aiven/internal/schemautil.copyServicePropertiesFromAPIResponseToTerraform(0xc00000e900?, 0xc00023a800, {0xc00035a028?, 0xc000424480?, 0xc000394010?}, {0xc000424480, 0xe})
	github.com/aiven/terraform-provider-aiven/internal/schemautil/service.go:622 +0xba5
github.com/aiven/terraform-provider-aiven/internal/schemautil.ResourceServiceRead({0xea3628, 0xc00078e480}, 0xc000293200, {0xd05600?, 0xc0000d6b00?})
	github.com/aiven/terraform-provider-aiven/internal/schemautil/service.go:328 +0x55c
github.com/aiven/terraform-provider-aiven/internal/service/kafka.resourceKafkaRead({0xea3628, 0xc00078e480}, 0xc000293200, {0xd05600?, 0xc0000d6b00?})
	github.com/aiven/terraform-provider-aiven/internal/service/kafka/resource_kafka.go:223 +0x4b0
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0xc0005947e0, {0xea3660, 0xc00079d290}, 0xd?, {0xd05600, 0xc0000d6b00})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.22.0/helper/schema/resource.go:724 +0x12e
[...]

Issue #909, opened 2022-09-30
Unable to copy api response into terraform schema: cannot set `kafka_user_config`


wex-ryan-whitwell:
## What happened?

I attempted to use data resource for `aiven_kafka` that was working in version 3.6.0 of the provider. I received the folliwing error

```
Error: unable to copy api response into terraform schema: cannot set `kafka_user_config` : kafka_user_config.0.ip_filter.0: '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[description: network:10.0.0.0/8]'; Please make sure that all Aiven services have unique s names
```

Terraform threw the following error after failing to resolve the data resource.
```
Error: unable to copy api response into terraform schema: cannot set `kafka_user_config` : kafka_user_config.0.ip_filter.0: '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[description: network:10.0.0.0/8]'; Please make sure that all Aiven services have unique s names
```

## What did you expect to happen?

I expected the data resource to resolve the same way it did in version 3.6.0

## What else do we need to know?

The version that causes the error is 3.7.0

It seems to be associated with the following update: https://github.com/aiven/terraform-provider-aiven/blob/main/CHANGELOG.md#370---2022-09-30

Regarding item `Fix empty user_config and empty ip_filters`

ivan-savciuc:
Hi @wex-ryan-whitwell, it seems to be a general limitation not related to our last release. Recently, the `ip_filter` field got extended. Now users can set it simply as an array of IPs, for example: `ip_filter = [ "10.0.0.0/8" ]` and this is something what we already support on the Terraform provider side and recently added another way to set it as a list of objects `ip_filter{ description = "xx", network ="10.0.0.0/8"}` and currently we do not support it (this one https://github.com/aiven/terraform-provider-aiven/pull/907 and some upcoming PR's would bring it to Terraform provider). 

We are working on adding support for the new API, and in the meantime, you can change how `static_ip` was set to an array notation.

Serpentiel:
hi one more time, thank you for your report!
I will also check this out just in case, as I am the author of those changes, and will fix ASAP, if there is any issue in the Terraform provider itself

mkhl:
fwiw: we (different customer) get these kinds of errors whenever one of us has changed something in the "Allowed IP Addresses" settings interface

some of us make temporary changes there to help them debug stuff; doing that seems to convert each network string into a map, which the terraform provider then has trouble with

[...]

PR #932, opened 2022-10-18
feat: add a converter for the service user configuration options ip_f…

ivan-savciuc:

add a converter for the service user configuration options `ip_filter` object format to the string slice
Resolves: #914, #909.


Serpentiel (approved):
lgtm

PR #940, opened 2022-10-25
add a reference to casandra description

byashimov:

Having a reference to Aiven Cassandra page might be helpful.


ivan-savciuc:
We have helper functions that dress the description field, and we could add a helper for all the services.

PR #941, opened 2022-10-26
fix vpc peering connection `terraform import`

byashimov:

VPC peering connection id parser doesn't validate user input and panics.

## Why this way

New parser:
- validates user input has valid size
- won't panic because of validation

### DIFF
--- a/internal/service/vpc/resource_aws_vpc_peering_connection.go
+++ b/internal/service/vpc/resource_aws_vpc_peering_connection.go
@@ -5,11 +5,11 @@ import (
 	"time"
 
 	"github.com/aiven/aiven-go-client"
-	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
 )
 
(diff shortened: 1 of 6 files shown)


Serpentiel:
could you please run the acceptance tests?

internal/service/vpc/resource_aws_vpc_peering_connection.go
```diff
@@ -168,27 +168,34 @@ func resourceAWSVPCPeeringConnectionCreate(ctx context.Context, d *schema.Resour
 func resourceAWSVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	client := m.(*aiven.Client)
 
-	projectName, vpcID, peerCloudAccount, peerVPC, peerRegion := parsePeeringVPCId(d.Id())
+	p, err := parsePeerVPCIDWithRegion(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
```
ivan-savciuc, line 173: Can you please use `diag.Errorf("Error ....: %s", err)` instead of FromError?

byashimov:
> could you please run the acceptance tests?

There is [only one](https://github.com/aiven/terraform-provider-aiven/blob/28d5d1af7e8235b5cc51e747c346db2006adc2e3/internal/service/vpc/resource_aws_vpc_peering_connection_test.go#L17) test for AWS and it requires AWS credentials.

ivan-savciuc (approved):
Would be nice to have more meaningful error messages,o otherwise LGTM

ivan-savciuc (approved):
LGTM

PR #945, opened 2022-10-28
fix Azure vpc peering connection import

byashimov:

Fixes Azure vpc peering connection import.

### DIFF
--- a/internal/service/vpc/resource_vpc_peering_connection.go
+++ b/internal/service/vpc/resource_vpc_peering_connection.go
@@ -287,9 +287,10 @@ func resourceVPCPeeringConnectionRead(_ context.Context, d *schema.ResourceData,
 	}
 
 	if isAzure {
-		if peerResourceGroup, ok := d.GetOk("peer_resource_group"); ok {
+		peerResourceGroup := schemautil.OptionalStringPointer(d, "peer_resource_group")
+		if peerResourceGroup != nil {
 			pc, err = client.VPCPeeringConnections.GetVPCPeeringWithResourceGroup(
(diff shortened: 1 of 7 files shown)


ivan-savciuc:
@muralibasani have you tried to import an Azure VPC manually with this patch?

internal/service/vpc/resource_vpc_peering_connection.go
```diff
@@ -634,3 +635,11 @@ func validateVPCID(i interface{}, k string) (warnings []string, errors []error)
 
 	return warnings, errors
 }
+
+// ToNilOrStringPointer returns string pointer or nil
+func ToNilOrStringPointer(s string) *string {
```
ivan-savciuc, line 640: It looks we can reuse `OptionalStringPointer` instead of this function

ivan-savciuc (changes requested):
I haven't tested it, since it request manual testing but code looks Ok.

internal/service/vpc/resource_vpc_peering_connection.go
byashimov: Yeah, have tested too. Unfortunately due to some unknown issues I can't provide a working test. I'll try to do that one more time. Function has been renamed.

ivan-savciuc (changes requested):
Thanks for testing locally.

Regarding the function name, I meant to say that we already have it in the schemautil package: https://github.com/aiven/terraform-provider-aiven/blob/0c7d157d35054bcf8b57aae9774b3d78ace22db8/internal/schemautil/schemautil.go#L26

Can you use it?

ivan-savciuc (approved):
LGTM

Issue #942, opened 2022-10-26
Terraform Import command for aiven_pg_user resource


barywhyte:
## What can we help you with?

Is there no way to import existing aiven pg database user?

```
resource "aiven_pg_user" "aiven_db_users" {
  project      = "my-${var.environment}"
  service_name = "service_name"
  username     = test-user
}
```
I am unable to see any import commands to import pg users that have been manually created.

I like to see something like: ` terraform import aiven_pg_user.aiven_db_users     whatever-follows` just like traditional terraform import commands
## Where would you expect to find this information?
https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/pg_user#import

I like to see something like:

## Import is supported using the following syntax:

`terraform import aiven_pg_user.aiven_db_users whatever-follows` 

in import section of https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/pg_user

byashimov:
Hi! Thanks for the question. I'll update the docs, there is a way to do that:

```
$ terraform import aiven_pg_user.aiven_pg_user my-project/aiven_pg_user/test-user
```

So it's basically `project_name/service_name/user_name`.

PR #949, opened 2022-11-02
add aiven_pg_user import example to docs

byashimov:

Adds pg user import example, covers #942


Serpentiel (approved):
lgtm

PR #952, opened 2022-11-03
Add three ClickHouse examples (standalone service, Postgres & Kafka integrations)

aris-aiven:
## About this change

This PR adds three ClickHouse examples: 
- a standalone service example showcasing privilege grants, and 
- two service integration examples for Kafka and PostgreSQL sources respectively

## Why this way

The standalone ClickHouse service example is on the toplevel examples dir to stay consistent with other services' examples.

The service integrations examples are grouped in a `clickhouse_integrations` directory: there may be more sources and sinks in the future. The two examples are split into two sub-directories for clarity and because users may want to tweak and test them separately as a first step.

The examples are not very worked in terms of access controls from the PostgreSQL and Kafka side of things because the goal is to highlight the integrations more than the other aspects of the provider's usage.

*Note*: The examples' ClickHouse plans will have to be updated when the service leaves Beta.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ nav_order: 1
 # Changelog
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
+- Add ClickHouse examples:
+  - Standalone service
 
 ## [3.9.0] - 2022-12-01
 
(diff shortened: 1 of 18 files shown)


ivan-savciuc (approved):
LGTM

PR #953, opened 2022-11-07
add aiven_azure_privatelink_connection_approval test

byashimov:

- adds aiven_azure_privatelink_connection_approval test
- fixes add aiven_azure_privatelink_connection_approval docs
- downgrades terraform-plugin-sdk to run multiple providers in [tests](https://github.com/hashicorp/terraform-plugin-sdk/issues/1064)

### DIFF
--- a/internal/service/vpc/resource_azure_privatelink_connection_approve_test.go
+++ b/internal/service/vpc/resource_azure_privatelink_connection_approve_test.go
@@ -1,0 +1,186 @@
+package vpc_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/kelseyhightower/envconfig"
+
+	acc "github.com/aiven/terraform-provider-aiven/internal/acctest"
+)
+
+// azurePLCASecrets AzurePrivateLinkConnectionApproval secrets
+type azurePLCASecrets struct {
+	Project        string `envconfig:"AIVEN_PROJECT_NAME" required:"true"`
+	ClientID       string `envconfig:"AZURE_CLIENT_ID" required:"true"`
+	ClientSecret   string `envconfig:"AZURE_CLIENT_SECRET" required:"true"`
+	TenantID       string `envconfig:"AZURE_TENANT_ID" required:"true"`
+	SubscriptionID string `envconfig:"AZURE_SUBSCRIPTION_ID" required:"true"`
+}
+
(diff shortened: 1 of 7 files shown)


internal/service/vpc/resource_azure_privatelink_connection_approve_test.go
```diff
@@ -0,0 +1,162 @@
...
+// azurePLCASecrets AzurePrivateLinkConnectionApproval secrets
+type azurePLCASecrets struct {
+	Project        string `envconfig:"AIVEN_PROJECT_NAME" required:"true"`
+	ClientID       string `envconfig:"AZURE_CLIENT_ID" required:"true"`
+	ClientSecret   string `envconfig:"AZURE_CLIENT_SECRET" required:"true"`
+	TenantID       string `envconfig:"AZURE_TENANT_ID" required:"true"`
+	SubscriptionID string `envconfig:"AZURE_SUBSCRIPTION_ID" required:"true"`
+}
+
+func TestAccAivenAzurePrivateLinkConnectionApproval_basic(t *testing.T) {
```
ivan-savciuc, line 23: Can you also add an `import` test?

ivan-savciuc (changes requested):
Two small things:
- Please run `make docs`, these two files should be added to this PR:
```
modified:   docs/data-sources/grafana.md
modified:   docs/resources/grafana.md
```

- Can you please run it locally and add here results?

internal/service/vpc/resource_azure_privatelink_connection_approve_test.go
```diff
@@ -0,0 +1,162 @@
+package vpc_test
```
Serpentiel, line 1: I believe we don't add a suffix to the name of the test packages in this project

Serpentiel (changes requested):
perhaps we should also have a changelog entry for this

internal/service/vpc/resource_azure_privatelink_connection_approve_test.go
byashimov: 1. It doesn't add changes to the provider.
2. vpc_test(s) are in a separate package because of cycle imports (the package already been there).

Serpentiel: > It doesn't add changes to the provider.

we try to list every change in the changelog, even when the docs were updated

the `no changelog` label should only be used, for example, when fixing a typo, or similar

this PR introduces new code so it definitely makes sense to add a changelog entry

internal/service/vpc/resource_azure_privatelink_connection_approve_test.go
byashimov: Done.

ivan-savciuc (approved):
LGTM, thought would be great to check Aiven VPC resources fields with `ImportStateCheck`, but it could be done in a separate PR

Serpentiel (approved):
lgtm

PR #954, opened 2022-11-08
fix: flatten ip_filter in case it contains objects

ivan-savciuc:

Flatten `ip_filter` from object to []string, this is an extension of this https://github.com/aiven/terraform-provider-aiven/pull/932 patch happens to be not converting `ip_filter` in case a user created a service and added IP Filters using Aiven Console and then uses data-source or imports such resource to Terraform.

Resolves: #909.


Serpentiel (approved):
lgtm

PR #962, opened 2022-11-15
feat(ssg): many to one conflicts with

Serpentiel:

disallows to set both `ip_filter` and `ip_filter_object`

## Why this way

because Terraform's `ConflictsWith` doesn't really works

### DIFF
--- a/internal/schemautil/custom_diff.go
+++ b/internal/schemautil/custom_diff.go
@@ -3,9 +3,12 @@ package schemautil
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/aiven/aiven-go-client"
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
 	"github.com/docker/go-units"
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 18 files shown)


internal/schemautil/custom_diff.go
```diff
@@ -173,3 +175,70 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
 	// TODO: Check that we block deletions that will result in too few static ips for the plan
 	return nil
 }
+
+// CustomizeDiffDisallowMultipleManyToOneKeys checks that we don't have multiple keys that are going to be converted to
+// a single key in the API request, e.g. 'ip_filter' and 'ip_filter_object' in the same diff.
+func CustomizeDiffDisallowMultipleManyToOneKeys(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	ks := d.GetRawConfig().AsValueMap()
+
+	tks := map[string]struct{}{}
```
ivan-savciuc, line 184: I think this can be a `[]string` string slice since we only use it as a storage for keys

internal/schemautil/custom_diff.go
```diff
@@ -173,3 +175,70 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
 	// TODO: Check that we block deletions that will result in too few static ips for the plan
 	return nil
 }
+
+// CustomizeDiffDisallowMultipleManyToOneKeys checks that we don't have multiple keys that are going to be converted to
+// a single key in the API request, e.g. 'ip_filter' and 'ip_filter_object' in the same diff.
+func CustomizeDiffDisallowMultipleManyToOneKeys(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+	ks := d.GetRawConfig().AsValueMap()
```
ivan-savciuc, line 182: We actually never use this variable other than as an argument for a loop. Can we do something like?:  
```
for k, v := range d.GetRawConfig().AsValueMap() {
...

// the same can be done here
for k, v := range  v.AsValueSlice()[0].AsValueMap() {
```

internal/schemautil/custom_diff.go
```diff
@@ -173,3 +175,70 @@ func CustomizeDiffCheckStaticIpDisassociation(_ context.Context, d *schema.Resou
...
+			}
+
+			ak := k
+
+			// TODO: Remove this in the next major version.
+			if k == "ip_filter" || k == "namespaces" {
+				ak = fmt.Sprintf("%s_string", k)
+			}
+
+			if userconfig.IsKeyTyped(ak) {
```
ivan-savciuc, line 209: We only explore the first level of user config, but we may have fields that we are looking for deeper, like `namespaces`

ivan-savciuc (dismissed):
If it try something like:
```
  m3db_user_config {
    rules {
      mapping {
        namespaces = ["test1"]
        namespaces_object {
          resolution = "test1"
        }
      }
    }
```

`terraform plan` is not showing any errors, as suspected here: https://github.com/aiven/terraform-provider-aiven/pull/962#discussion_r1022958736

internal/schemautil/custom_diff.go
Serpentiel: this increases code complexity

internal/schemautil/custom_diff.go
Serpentiel: fixed

internal/schemautil/custom_diff.go
ivan-savciuc: I think a bit differently, I always tend to check where the variable is used, and since we only use it for things like: 
```
nks := v.AsValueSlice()[0].AsValueMap()
tks = typedKeys(nks)
```
I vote for something like: 
```
tks = typedKeys(v.AsValueSlice()[0].AsValueMap())
```

The less lines of code the better :)

ivan-savciuc (approved):
LGTM

PR #964, opened 2022-11-17
increase default timeout for kafka_connect service integration create, add example test

byashimov:

Adds extra timeout for kafka_connect create. Adds example test.

### DIFF
--- a/internal/service/service_integration/resource_service_integration.go
+++ b/internal/service/service_integration/resource_service_integration.go
@@ -8,15 +8,15 @@ import (
 	"time"
 
 	"github.com/aiven/aiven-go-client"
-	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
-	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
-	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig/apiconvert"
-	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig/dist"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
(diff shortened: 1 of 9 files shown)


internal/sweep/sweep.go
```diff
@@ -55,7 +55,7 @@ func SweepServices(region, t string) error {
 			continue
 		}
 
-		if !strings.Contains(s.Name, "test-acc-") {
```
ivan-savciuc, line 58: Can you also add the a K8s acc and e2e tests prefix?

internal/service/service_integration/resource_service_integration.go
```diff
@@ -148,7 +148,7 @@ func ResourceServiceIntegration() *schema.Resource {
 			StateContext: schema.ImportStatePassthroughContext,
 		},
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(5 * time.Minute),
+			Create: schema.DefaultTimeout(10 * time.Minute),
```
ivan-savciuc, line 151: Customers can configure this field for themselves using something like: 
```
resource "aiven_service_integration" "test" {
  ....
  timeout {
    create = 10m
  }
}
```

ivan-savciuc (changes requested):
Can you change the commit message? 

For example, we claim `fix kafka_connect service integration`, but we have no fix in the code, just a default timeout has been updated, but any user can do that by themselves. What fixed is example, and the current commit message is a bit misleading .

internal/service/service_integration/resource_service_integration.go
byashimov: Yeah, I know. But _default_ is not enough.

internal/sweep/sweep.go
byashimov: Done. But actually operators tests are mess. Naming may vary and doesn't have strict pattern. Can't we just sweep all?

ivan-savciuc: I think we can be on the same side if we use `k8s-*` for K8s tests, all of the tests start with this prefix

ivan-savciuc (dismissed):
LGTM

ivan-savciuc:
@byashimov  pls rebase to the latest main, it has some conflicts: 
```
internal/service/service_integration/resource_service_integration.go
```

byashimov:
> @byashimov pls rebase to the latest main, it has some conflicts:
> 
> ```
> internal/service/service_integration/resource_service_integration.go
> ```

Done, thanks.

Serpentiel (approved):
lgtm

PR #976, opened 2022-11-28
refactor: add a deprecation notice for ip_filter and namespaces

Serpentiel:

adds deprecation notice for `ip_filter` and `namespaces`

## Why this way

to inform our users about the upcoming changes


Serpentiel:
acceptance tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/3565594846

byashimov:
I guess we can leave the old field name for compatibility and a new one.

ivan-savciuc:
I think we should mark them as deprecated and can only drop in v5

ivan-savciuc (approved):
LGTM

PR #979, opened 2022-11-30
fix(ssg): prevent generation of Default field

Serpentiel:

`Default` field works a bit unexpected and makes the diff to have those values set even if the user never requested it

## Why this way

to fix it


Serpentiel:
acceptance tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/3584155860

ivan-savciuc (approved):
LGTM

ivan-savciuc (approved):
LGTM

PR #986, opened 2022-12-07
replace aiven_service in various places

byashimov:

Replaces deprecated `aiven_service` resource in examples.  
This request doesn't update ES examples.


Serpentiel (approved):
lgtm

Issue #988, opened 2022-12-07
Cannot upgrade provider from v3.6.0 to v3.9.0


gandelman-a:
## What happened?

Attempted to bump provider v3.6.0 -> v3.9.0 to test a fix.  After doing so, I cannot refresh terraform state. It fails while refreshing an `aiven_vpc_peering_connection` resource.  The VPC peering connection is used to connect Azure VPCs:

```
resource "aiven_vpc_peering_connection" "peering" {
  vpc_id               = aiven_project_vpc.vpc.id
  peer_vpc             = var.azure_networking.virtual_network.id
  peer_azure_app_id    = azuread_application.space_peering.application_id
  peer_azure_tenant_id = data.azurerm_client_config.current.tenant_id
  peer_cloud_account   = data.azurerm_subscription.current.subscription_id
  peer_resource_group  = var.resource_group.name
}
```

We currently pass the azure vnet name to `peer_vpc`, as per [docs](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/vpc_peering_connection) this is supported for AWS/GCP. This has worked fine to date:

![image](https://user-images.githubusercontent.com/274055/206264362-2ad148cc-a556-41e8-a1c0-48a44a152b54.png)

However, bumping to `v3.9.0` of the provider, cannot refresh state:

```
│ Error: error parsing peering VPC ID: expected unix path-like string with 5 chunks
│
│   with module.aiven_networking.aiven_vpc_peering_connection.peering,
│   on ../../../modules/aiven/azure-networking/main.tf line 16, in resource "aiven_vpc_peering_connection" "peering":
│   16: resource "aiven_vpc_peering_connection" "peering" {
│
```

## What did you expect to happen?

Init/plan/apply with no changes

PR #989, opened 2022-12-07
fix vpc peering id parser

byashimov:

VPC peering CRUD functions accept IDs both 4 and 5 chunks length,
so parser must expect different sizes.  
Resolves: #988.


Serpentiel (approved):
lgtm

PR #991, opened 2022-12-08
add kafka_authentication_method support

malonsocasas:
Would it be possible to add `kafka_authentication_method` attribute to the list of components from the output of an Aiven service  ? 

We would like to be able to differentiate components between "sasl" and "certificate" and as of today, this value (`aiven.kafka.components[].kafka_authentication_method`) is always an empty string. 

## Why this way

The [Aiven API](https://api.aiven.io/doc/#tag/Service/operation/ServiceGet) returns this attribute (it's not required though) and it seems to be supported by the [Aiven Go Client](https://github.com/aiven/aiven-go-client/blob/master/service.go#L84)

### DIFF
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -693,11 +693,12 @@ func FlattenServiceComponents(r *aiven.Service) []map[string]interface{} {
 
 	for _, c := range r.Components {
 		component := map[string]interface{}{
-			"component": c.Component,
-			"host":      c.Host,
-			"port":      c.Port,
-			"route":     c.Route,
-			"usage":     c.Usage,
+			"component":                   c.Component,
+			"host":                        c.Host,
+			"port":                        c.Port,
+			"route":                       c.Route,
+			"usage":                       c.Usage,
+			"kafka_authentication_method": c.KafkaAuthenticationMethod,
 		}
 		components = append(components, component)
 	}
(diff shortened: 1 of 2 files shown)


byashimov (dismissed):

internal/schemautil/service.go
```diff
@@ -693,11 +693,12 @@ func FlattenServiceComponents(r *aiven.Service) []map[string]interface{} {
...
-			"host":      c.Host,
-			"port":      c.Port,
-			"route":     c.Route,
-			"usage":     c.Usage,
+			"component":                   c.Component,
+			"host":                        c.Host,
+			"port":                        c.Port,
+			"route":                       c.Route,
+			"usage":                       c.Usage,
+			"kafka_authentication_method": c.KafkaAuthenticationMethod,
```
Serpentiel, line 701: could this please be moved to be in a specific service that this is related to?

Serpentiel (changes requested):

ivan-savciuc (commented):
@byashimov I think we also need to add this property to `aiven_service_component` data-source

byashimov:
@ivan-savciuc it's already [there](https://github.com/aiven/terraform-provider-aiven/blob/795427f1d07b51e8cbf3ebd98dc00de8f1ad693c/internal/service/service_component/datasource_service_component_test.go#L91-L91).

internal/schemautil/service.go
Serpentiel: never mind, this would require serious refactoring, and is inappropriate to be done in v3.x, since we already have the field in every service

Serpentiel (approved):
lgtm

PR #994, opened 2022-12-12
fix(ip_filter): misconversion when using frontend

### LINKED ISSUE #987: `v3.9.0` : existing `ip_filter` gets destroyed because of `ip_filter_object`
after upgrading to `v3.9.0` terraform destroys the existing  `ip_filter`  attribute in `aiven_kafka` and adds an empty `ip_filter_object` removing all IP filtering from our kafka cluster!

plan output:

```terraform

 ~ kafka_user_config {
          ~ ip_filter                 = [
              + "xx.xx.xx.xx/xx",
              + "xx.xx.xx.xx/xx",
              + "xx.xx.xx.xx/xx"
            ]

          - ip_filter_object {
              - network = "xx.xx.xx.xx/xx" -> null
            }
          - ip_filter_object {
              - network = "xx.xx.xx.xx/xx" -> null
            }
          - ip_filter_object {
              - network = "xx.xx.xx.xx/xx" -> null
            }

            # (3 unchanged blocks hidden)
        }

```

checking the aiven console event log, all IPs were removed

This problem exists when there is some IPs managed in the UI and the rest in terraform, which was not a problem before 3.9.0

Serpentiel:

Fix changes being incorrectly picked up by the `ip_filter` field after changing it in frontend

resolves #987 

## Why this way

currently, the backend logic for the `ip_filter` field is mutable with schema changes from other API clients, e.g. the frontend client

this has to be addressed both in frontend and backend, for now we will introduce this small fix on our side

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,9 @@ nav_order: 1
   - Standalone service
   - Integration with Kafka source
   - Integration with PostgreSQL source
-- Fix vpc peering id parser
+- Fix VPC peering ID parser
 - Add `offset_syncs_topic_location` support for `aiven_mirrormaker_replication_flow` resource 
+- Fix changes being incorrectly picked up by the `ip_filter` field after changing it in frontend
 
 ## [3.9.0] - 2022-12-01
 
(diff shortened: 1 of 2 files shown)


jlippmann:
Hi, this PR has been CLOSED but not MERGED, are there any reasons for that? We were affected by #987 and we are wondering whether the problem is solved with 3.10.0 or if we should stick with 3.8.1

Serpentiel:
hey, @jlippmann! 👋 

we will soon get back to you with more information regarding this question, this PR appeared to break in certain cases

if it is possible, please stick to v3.8.1 for now, thank you!

PR #995, opened 2022-12-12
feat(ssg): strict types

Serpentiel:

makes SSG generate schema with strict types

## Why this way

to have strict types in place


ivan-savciuc (approved):
LGTM

PR #996, opened 2022-12-12
expose ssl field in service components

### LINKED ISSUE #992: aiven_pg: Add an attribute for pgbouncer service port
## What is currently missing?

With the `aiven_pg` resource, there is a `service_port` attribute, and a `pg.port` attribute. There is also a `components` array of objects (which cannot be referenced easily). There is no `pgbouncer_port` attribute. Right now, we have to assume this is `service_port` + 1.

Create a dedicated attribute for referencing the pgbouncer port.

Currently the `components` attribute looks like this:

```
    components              = [
        {
            component                   = "pg"
            host                        = "myservice-myproject.a.timescaledb.io"
            kafka_authentication_method = ""
            port                        = 23514
            route                       = "dynamic"
            ssl                         = false
            usage                       = "primary"
        },
        {
            component                   = "pg"
            host                        = "public-myservice-myproject.a.timescaledb.io"
            kafka_authentication_method = ""
            port                        = 23514
            route                       = "public"
            ssl                         = false
            usage                       = "primary"
        },
        {
            component                   = "pgbouncer"
            host                        = "myservice-myproject.a.timescaledb.io"
            kafka_authentication_method = ""
            port                        = 23515
            route                       = "dynamic"
            ssl                         = false
            usage                       = "primary"
        },
    ]

```

Note: there is a different bug here where `ssl=false` but TLS / SSL is required.

byashimov:

- Exposes `ssl` field in service components.
- Adds test for `kafka_authentication_method` field

Resolves #992.

### DIFF
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -7,9 +7,10 @@ import (
 	"strings"
 
 	"github.com/aiven/aiven-go-client"
+	"github.com/docker/go-units"
+
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig/apiconvert"
-	"github.com/docker/go-units"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -699,6 +700,9 @@ func FlattenServiceComponents(r *aiven.Service) []map[string]interface{} {
 			"route":                       c.Route,
 			"usage":                       c.Usage,
 			"kafka_authentication_method": c.KafkaAuthenticationMethod,
+			// By default, endpoints are always encrypted and
+			// this property is only included for service components that may disable encryption.
+			"ssl": PointerValueOrDefault(c.Ssl, true),
 		}
 		components = append(components, component)
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ nav_order: 1
   - Integration with PostgreSQL source
 - Fix vpc peering id parser
 - Add `offset_syncs_topic_location` support for `aiven_mirrormaker_replication_flow` resource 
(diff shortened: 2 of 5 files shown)


byashimov:
=== RUN   TestAccAiven_kafka
=== PAUSE TestAccAiven_kafka
=== CONT  TestAccAiven_kafka
--- PASS: TestAccAiven_kafka (651.30s)
PASS

internal/schemautil/service.go
```diff
@@ -699,6 +700,9 @@ func FlattenServiceComponents(r *aiven.Service) []map[string]interface{} {
 			"route":                       c.Route,
 			"usage":                       c.Usage,
 			"kafka_authentication_method": c.KafkaAuthenticationMethod,
+			// By default, endpoints are always encrypted and
+			// this property is only included for service components that may disable encryption.
+			"ssl": PointerValueOrDefault(c.Ssl, true),
```
ivan-savciuc, line 705: We should always propagate what is coming from the API response, and often this field is not set in the response, and in these situations, we keep it empty in Terraform state.

ivan-savciuc (changes requested):
I'm not sure about the default `true` when `ssl` is null or not present in API response, I think it should be empty in TF state as well.

internal/schemautil/service.go
byashimov: It is in a state anyway, but with with a wrong value:

```json
"ssl": false,
```

And since it is computed, we can't set it's default value.  
You can also see it in the [issue](https://github.com/aiven/terraform-provider-aiven/issues/992).

ivan-savciuc: Then I think we can set `null` https://developer.hashicorp.com/terraform/language/expressions/types#null to this field or, in the worst case scenario, change a type to a string and when it is not set, set it empty.

ivan-savciuc: Hey, I did some experiments with the `null` value and it is not that simple. So I guess it is a good idea in the end to set it `true`

ivan-savciuc (approved):
LGTM

PR #999, opened 2022-12-13
feat(ssg): support strict types in diff funcs

Serpentiel:

adds support for strict types in diff functions

## Why this way

to make the strict types function properly


Serpentiel:
acceptance tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/3686897628

Issue #997, opened 2022-12-13
admin_username usage causes resource recreation in 3.9.0


joonvena:
## What happened?

When trying to update from provider version 3.8.0 > 3.9.0 `terraform plan` wants to replace `aiven_pg` resource:

```
~ pg_user_config {
          - additional_backup_regions = [] -> null
          + admin_username            = "avnadmin" # forces replacement
```

## What did you expect to happen?

I would expect this not to cause replace for the resource

## What else do we need to know?

If i remove the `admin_username` setting from the IaC code then it works fine but does this really means that this settings should be only used when creating the service initially and then removing it from the IaC code as the documentation suggest?

`Custom username for admin user. This must be set only when a new service is being created.`

This same happens also with the `aiven_mysql` resource.

Serpentiel:
hey, @joonvena 👋 

thank you for reporting this, this is currently being investigated by the team members

PR #1000, opened 2022-12-13
fix admin credentials diff

byashimov:

Saves admin credentials to the state to make plan work correctly 

Resolves: #997.

### DIFF
--- a/internal/schemautil/ipfilter.go
+++ b/internal/schemautil/ipfilter.go
@@ -7,29 +7,29 @@ import (
 // NormalizeIpFilter compares a list of IP filters set in TF and a sorted version coming
 // from Aiven and takes sort IP filters such that all matching entries will be in
 // the same order as defined in the TF manifest.
-func NormalizeIpFilter(tfUserConfig interface{}, userConfig []map[string]interface{}) []map[string]interface{} {
+func NormalizeIpFilter(tfUserConfig interface{}, userConfig []map[string]interface{}) {
 	tfInt, ok := tfUserConfig.([]interface{})
 
 	if len(userConfig) == 0 {
-		return userConfig
+		return
 	}
 
 	if _, ok := userConfig[0]["ip_filter"]; !ok {
-		return userConfig
+		return
 	}
 
 	userConfig[0]["ip_filter"] = toStringSlice(userConfig[0]["ip_filter"].([]interface{}))
 
 	if !ok || len(tfInt) == 0 {
-		return userConfig
+		return
 	}
 
 	if tfInt[0] == nil {
-		return userConfig
+		return
 	}
 
 	if _, ok := tfInt[0].(map[string]interface{})["ip_filter"]; !ok {
-		return userConfig
+		return
 	}
 
 	api := userConfig[0]["ip_filter"].([]string)
@@ -77,7 +77,6 @@ func NormalizeIpFilter(tfUserConfig interface{}, userConfig []map[string]interfa
 	}
 
 	userConfig[0]["ip_filter"] = trim(append(newIpFilters, diff...))
-	return userConfig
 }
 
 // toStringSlice converts []interface to a []string
--- a/internal/schemautil/sensitive.go
+++ b/internal/schemautil/sensitive.go
@@ -1,0 +1,27 @@
+package schemautil
+
+import (
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
+)
+
+// Those fields are not returned by API,
+// but should be in the state to make terraform work correctly
+var sensitiveFields = []string{
+	"admin_username",
+	"admin_password",
+}
+
+// copySensitiveFields copies sensitive fields to the state which not returned by API,
+// but exist in the manifest
+func copySensitiveFields(oldSrc interface{}, new []map[string]interface{}) {
+	old := userconfig.UnmarshalUserConfig(oldSrc)
+	if len(old)*len(new) == 0 {
+		return
+	}
+
+	for _, k := range sensitiveFields {
+		if v, ok := old[k]; ok {
+			new[0][k] = v
+		}
+	}
+}
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -636,12 +636,17 @@ func copyServicePropertiesFromAPIResponseToTerraform(
 		}
 	}
 
-	userConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
+	newUserConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
 	if err != nil {
 		return err
 	}
 
-	if err := d.Set(serviceType+"_user_config", NormalizeIpFilter(d.Get(serviceType+"_user_config"), userConfig)); err != nil {
+	// Mutates user config inplace
+	oldUserConfig := d.Get(serviceType + "_user_config")
+	copySensitiveFields(oldUserConfig, newUserConfig)
+	NormalizeIpFilter(oldUserConfig, newUserConfig)
+
+	if err := d.Set(serviceType+"_user_config", newUserConfig); err != nil {
 		return fmt.Errorf("cannot set `%s_user_config` : %s; Please make sure that all Aiven services have unique s names", serviceType, err)
 	}
 
--- a/internal/schemautil/userconfig/util.go
+++ b/internal/schemautil/userconfig/util.go
@@ -174,3 +174,11 @@ func DecodeKey(k string) string {
 func IsKeyTyped(k string) bool {
 	return regexp.MustCompile(`^.*_(boolean|integer|number|string|array|object)$`).MatchString(k)
 }
+
+func UnmarshalUserConfig(src interface{}) map[string]interface{} {
+	configList, ok := src.([]interface{})
+	if !ok || len(configList) == 0 {
+		return nil
+	}
+	return configList[0].(map[string]interface{})
+}
--- a/internal/service/pg/resource_pg_test.go
+++ b/internal/service/pg/resource_pg_test.go
@@ -4,11 +4,13 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 	"testing"
 
-	acc "github.com/aiven/terraform-provider-aiven/internal/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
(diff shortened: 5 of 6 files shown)


internal/schemautil/service.go
```diff
@@ -636,12 +636,17 @@ func copyServicePropertiesFromAPIResponseToTerraform(
...
-	userConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
+	newUserConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
 	if err != nil {
 		return err
 	}
 
-	if err := d.Set(serviceType+"_user_config", NormalizeIpFilter(d.Get(serviceType+"_user_config"), userConfig)); err != nil {
+	// Mutates user config inplace
+	oldUserConfig := d.Get(serviceType + "_user_config")
+	copySensitiveFields(oldUserConfig, newUserConfig)
```
Serpentiel, line 646: I'm not sure whether this is an idiomatic Go approach, e.g. the function `append` does not mutates data, but returns the new version of it instead

internal/schemautil/service.go
```diff
@@ -636,12 +636,17 @@ func copyServicePropertiesFromAPIResponseToTerraform(
...
 
-	userConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
+	newUserConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
 	if err != nil {
 		return err
 	}
 
-	if err := d.Set(serviceType+"_user_config", NormalizeIpFilter(d.Get(serviceType+"_user_config"), userConfig)); err != nil {
+	// Mutates user config inplace
+	oldUserConfig := d.Get(serviceType + "_user_config")
```
Serpentiel, line 645: could `oldUserConfig` please be renamed to `curUserConfig` or similar, since at this point in time in the code, the user config was not changed yet, thus it is not an old one, yet

internal/schemautil/service.go
Serpentiel, line 646: could we please have some kind of a wrapper which would call both of these functions, e.g. `mutateUserConfig`

internal/schemautil/userconfig/util.go
```diff
@@ -174,3 +174,11 @@ func DecodeKey(k string) string {
 func IsKeyTyped(k string) bool {
 	return regexp.MustCompile(`^.*_(boolean|integer|number|string|array|object)$`).MatchString(k)
 }
+
+func UnmarshalUserConfig(src interface{}) map[string]interface{} {
+	configList, ok := src.([]interface{})
+	if !ok || len(configList) == 0 {
+		return nil
+	}
+	return configList[0].(map[string]interface{})
```
Serpentiel, line 183: this should be type asserted

internal/service/pg/resource_pg_test.go
```diff
@@ -598,3 +600,46 @@ resource "aiven_pg" "bar" {
 }
 `
 }
+
+func TestAccAivenPG_admin_creds(t *testing.T) {
```
Serpentiel, line 604: add a comment: we comment all functions, tests too

internal/schemautil/service.go
```diff
@@ -636,12 +636,17 @@ func copyServicePropertiesFromAPIResponseToTerraform(
 		}
 	}
 
-	userConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
+	newUserConfig, err := apiconvert.FromAPI(userconfig.ServiceTypes, serviceType, s.UserConfig)
```
Serpentiel, line 639: what would be the need to rename this variable?

internal/schemautil/sensitive.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// Those fields are not returned by API,
+// but should be in the state to make terraform work correctly
+var sensitiveFields = []string{
+	"admin_username",
+	"admin_password",
+}
+
+func copySensitiveFields(oldSrc interface{}, new []map[string]interface{}) {
```
Serpentiel, line 14: please add a comment

internal/service/pg/resource_pg_test.go
```diff
@@ -598,3 +600,46 @@ resource "aiven_pg" "bar" {
...
+					}),
+					resource.TestCheckResourceAttr(resourceName, "pg_user_config.0.admin_username", "root"),
+					resource.TestCheckResourceAttr(resourceName, "pg_user_config.0.admin_password", prefix+"-password"),
+				),
+			},
+		},
+	})
+}
+
+func testAccPGResourceAdminCreds(prefix, project string) string {
```
Serpentiel, line 631: add a comment

internal/schemautil/ipfilter.go
```diff
@@ -7,29 +7,29 @@ import (
 // NormalizeIpFilter compares a list of IP filters set in TF and a sorted version coming
 // from Aiven and takes sort IP filters such that all matching entries will be in
 // the same order as defined in the TF manifest.
-func NormalizeIpFilter(tfUserConfig interface{}, userConfig []map[string]interface{}) []map[string]interface{} {
+func NormalizeIpFilter(tfUserConfig interface{}, userConfig []map[string]interface{}) {
 	tfInt, ok := tfUserConfig.([]interface{})
 
 	if len(userConfig) == 0 {
-		return userConfig
```
Serpentiel, line 14: what was the need to get rid of the idiomatic Go way to modify data?

although, I do agree that this would need some extra code which copies the map

Serpentiel (changes requested):

Serpentiel (changes requested):
please also fix lint errors

internal/schemautil/ipfilter.go
byashimov: You don't create new map, you modify existing https://go.dev/play/p/TjqI6b4VDjL. Returning a value looks like you return a new value, but it's not.

internal/schemautil/service.go
byashimov: NormalizeIpFilter mutates data. See comment below.

internal/schemautil/service.go
byashimov: TF uses naming old and new for diff suppress func.

internal/schemautil/service.go
byashimov: To keep tf style old vs new.

internal/schemautil/ipfilter.go
Serpentiel: I understand how maps function; as I said above, this would require to add extra code which copies the map, e.g. by applying `maps.Clone`, so I do agree that the code does not works as expected the way we have it at the moment

internal/service/pg/resource_pg_test.go
byashimov: [Not sure](https://github.com/aiven/terraform-provider-aiven/blob/1068fb7728f2c5df59c59e987ce5be330bff7f43/internal/service/pg/resource_pg_test.go#L15-L15) if that's so. OK I'll add comment to this.

internal/schemautil/sensitive.go
byashimov: It's just few rows upper. But added more.

internal/schemautil/service.go
byashimov: I won't modify NormalizeIpFilter. It has quite confusing code and doesn't provide extra tests for that complexity. Just check out first few 20-30 lines and early exits.

Serpentiel: this function is too long as it is already, and moving those call into a wrapper, would actually reduce complexity since the reader shouldn't care what exactly is being mutated under the hood, rather it should be visible that the data is being changed

internal/service/pg/resource_pg_test.go
Serpentiel: > [Not sure](https://github.com/aiven/terraform-provider-aiven/blob/1068fb7728f2c5df59c59e987ce5be330bff7f43/internal/service/pg/resource_pg_test.go#L15-L15) if that's so

we don't have necessary capacity to add comments everywhere, but we are supposed to be commenting each new function that we add

internal/schemautil/service.go
Serpentiel: please see https://github.com/aiven/terraform-provider-aiven/pull/1000/files#r1048277751

internal/schemautil/userconfig/util.go
Serpentiel: we already do type assertion at the beginning of this function, how exactly would it be different?

byashimov: Well, I can remove "ok" check, because it doensn't matter. But I can't cast it in one line, golang doesn't let it do so. That's why there is two steps.

Serpentiel: after discussing internally, this has been addressed

internal/schemautil/userconfig/util.go
```diff
@@ -174,3 +176,20 @@ func DecodeKey(k string) string {
 func IsKeyTyped(k string) bool {
 	return regexp.MustCompile(`^.*_(boolean|integer|number|string|array|object)$`).MatchString(k)
 }
+
+func UnmarshalUserConfig(src interface{}) (map[string]interface{}, error) {
```
ivan-savciuc, line 180: I think we can move `UnmarshalUserConfig` to the `internal/schemautiul` package and do not export this function `UnmarshalUserConfig -> unmarshalUserConfig` since we only use it there.

ivan-savciuc (commented):
Perhaps we should focus on a Naming Convention/Code style document to avoid lengthy discussions in the future regarding the names of the local variables. Ideally, something that we can enable in the CI linter or IDE.

My only suggestion is perhaps to move  `UnmarshalUserConfig` to a different package, but this should not block the PR; we can do such things later in a separate one.

internal/schemautil/service.go
```diff
@@ -636,12 +636,21 @@ func copyServicePropertiesFromAPIResponseToTerraform(
...
 	if err != nil {
 		return err
 	}
 
-	if err := d.Set(serviceType+"_user_config", NormalizeIpFilter(d.Get(serviceType+"_user_config"), userConfig)); err != nil {
+	// Mutates user config inplace
```
ivan-savciuc, line 644: I think it is a typo `in place` -> `in place` ?

ivan-savciuc (approved):
LGTM

Serpentiel (approved):
lgtm

PR #1007, opened 2022-12-21
fix concurrent map writes

byashimov:

Fixes races in tests because of concurrent map writes to `cachedRepresentationMaps`.


ivan-savciuc (approved):
LGTM

Issue #1004, opened 2022-12-16
Handle correctly missing (or deleted) `kafka_topics`


bkamin29:
## What is currently missing?

Currently, when you stop and then restart your kafka service, all kafka topics are deleted.
There is a consequence on the terraform state, because the provider return a `404` issue.

``` error
Error: error waiting for Aiven Kafka topic to be ACTIVE: one of the Kafka Topics from the queue [[mytopic]] is not found: 404: {"errors":[{"message":"Topic(s) not found","status":404}],"message":"Topic(s) not found"} - 
```

How to reproduce ? by using this terraform code:

``` terraform
resource "aiven_kafka" "kafka1" {
  project                 = "sandbox-bki"
  cloud_name              = "google-europe-west1"
  plan                    = "startup-2"
  service_name            = "my-kafka1"
  maintenance_window_dow  = "monday"
  maintenance_window_time = "10:00:00"
}

resource "aiven_kafka_topic" "mytesttopic" {
  project                = aiven_kafka.kafka1.project
  service_name           = aiven_kafka.kafka1.service_name
  topic_name             = "mytopic"
  partitions             = 5
  replication            = 3

  config {
    flush_ms                       = 10
    unclean_leader_election_enable = true
    cleanup_policy                 = "compact,delete"
  }
}

terraform {
  required_providers {
    aiven = {
      source = "aiven/aiven"
      version = "= 3.10.0"
    }
  }
}

provider "aiven" {
  api_token = var.aiven_api_token
}

variable "aiven_api_token" {
  
}
```

## How could this be improved?

The provider should handle correctly deleted or missing kafka topics.
During terraform refresh process, if Aiven API return a `404`, the provider should delete the resource from state to, don't block terraform and to re-create the kafka topic during the next plan.

## Is this a feature you would work on yourself?

* [ ] I plan to open a pull request for this feature.

byashimov:
Hi @bkamin29!  
Thanks for the reporting. It is in the backlog now.

ivan-savciuc:
@bkamin29 what version of Terraform provider are you using?

bkamin29:
> @bkamin29 what version of Terraform provider are you using?

Oh sorry Ivan, it was from the latest version, `v3.10.0`

bkamin29:
perfect, thank you @ivan-savciuc

PR #1008, opened 2022-12-22
recreate missing kafka topics

byashimov:

Recreates missing kafka topics (if service was off and on)
Resolves: #1004.

### DIFF
--- a/examples_tests/base.go
+++ b/examples_tests/base.go
@@ -10,6 +10,8 @@ import (
 	"github.com/gruntwork-io/terratest/modules/terraform"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/stretchr/testify/suite"
+
+	"github.com/aiven/terraform-provider-aiven/internal/common"
 )
 
 type envConfig struct {
@@ -107,7 +109,7 @@ func examplesRandPrefix() func(string) string {
 }
 
 func newClient(token string) (*aiven.Client, error) {
-	client, err := aiven.NewTokenClient(token, "terraform-provider-aiven/0.11+compatible/dev")
+	client, err := common.NewAivenClientWithToken(token)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/common/client.go
+++ b/internal/common/client.go
@@ -1,0 +1,34 @@
+package common
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/aiven/aiven-go-client"
+)
+
+func NewAivenClient() (*aiven.Client, error) {
+	return NewAivenClientWithToken(os.Getenv("AIVEN_TOKEN"))
+}
+
+func NewAivenClientWithToken(token string) (*aiven.Client, error) {
+	return NewCustomAivenClient(token, "", "")
+}
+
+func NewCustomAivenClient(token, tfVersion, buildVersion string) (*aiven.Client, error) {
+	if token == "" {
+		return nil, fmt.Errorf("token is required for Aiven client")
+	}
+
+	if tfVersion == "" {
+		// Terraform 0.12 introduced this field to the protocol
+		// We can therefore assume that if it's missing it's 0.10 or 0.11
+		tfVersion = "0.11+compatible"
+	}
+
+	if buildVersion == "" {
+		buildVersion = "dev"
+	}
+
+	return aiven.NewTokenClient(token, fmt.Sprintf("terraform-provider-aiven/%s/%s", tfVersion, buildVersion))
+}
--- a/internal/service/kafka/kafka_topic_cache.go
+++ b/internal/service/kafka/kafka_topic_cache.go
@@ -136,3 +136,13 @@ func (t *kafkaTopicCache) GetQueue(projectName, serviceName string) []string {
 
 	return t.inQueue[projectName+serviceName]
 }
+
+// FlushTopicCache for tests only!
+func FlushTopicCache() {
+	c := getTopicCache()
+	c.Lock()
+	for k := range c.internal {
+		delete(c.internal, k)
+	}
+	c.Unlock()
+}
--- a/internal/service/kafka/resource_kafka_topic.go
+++ b/internal/service/kafka/resource_kafka_topic.go
@@ -7,12 +7,12 @@ import (
 	"time"
 
 	"github.com/aiven/aiven-go-client"
-	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-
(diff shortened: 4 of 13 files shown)


internal/service/kafka/resource_kafka_topic_test.go
```diff
@@ -341,3 +345,144 @@ func TestPartitions(t *testing.T) {
...
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(kafkaResource, "id", kafkaID),
+					resource.TestCheckResourceAttr(topicResource, "id", topicID),
+				),
+			},
+			{
+				// Step 2: deletes topic, then runs apply, same config & checks
+				PreConfig: func() {
+					client, err := provider.DefaultAivenClient()
```
Serpentiel, line 379: an instance of `*aiven.Client` is already available by calling `provider.Provider().Meta().(*aiven.Client)`, so I'm not sure if this is justified to have a separate exported constructor function for the client

moreover, the constructor function appears to violate the `provider` package's semantic, and even if we do want to have it this way, it should probably be moved to some other package that makes more sense for it, e.g. a new package called `client` or `clientutil`, although, it probably worths to address the aforementioned necessity of obtaining the client in this manner, first

internal/service/kafka/kafka_topic_cache.go
```diff
@@ -136,3 +136,13 @@ func (t *kafkaTopicCache) GetQueue(projectName, serviceName string) []string {
 
 	return t.inQueue[projectName+serviceName]
 }
+
+// FlushTopicCache for tests only!
+func FlushTopicCache() {
```
Serpentiel, line 141: could this be moved to the test package itself? iirc, test packages also have access to non-exported functions

Serpentiel (changes requested):

internal/service/kafka/resource_kafka_topic_test.go
Serpentiel: moreover, the constructor function appears to violate the `provider` package's semantics, and even if we do want to have it this way, it should probably be moved to some other package that makes more sense for it, e.g. a new package called `client` or `clientutil`, although, it probably worths to address the aforementioned necessity of obtaining the client in this manner, first

byashimov: `provider.Provider().Meta().(*aiven.Client)` 
Doesn't work. It doesn't cast to the client.  
Moved client constructors to a separate package.

byashimov:
ACC https://github.com/aiven/terraform-provider-aiven/actions/runs/3760191113/jobs/6390628545

internal/service/kafka/resource_kafka_topic_test.go
byashimov: Found it `acc.TestAccProvider.Meta().(*aiven.Client)`

Serpentiel: > Moved client constructors to a separate package.

I don't quite understand why do we need to have them at all, since we shouldn't be constructing multiple clients in the provider anyway

byashimov: But we have them in provider, sweep and example tests.

Serpentiel: did you forget to update the example tests?

byashimov: No. It's [there](https://github.com/aiven/terraform-provider-aiven/pull/1008/files#diff-25fd8eb50f1be757367ab21a98d4ee45ac5eae017f561fa4fa72a2f41430a4cbR91).

internal/service/kafka/resource_kafka_topic.go
```diff
@@ -330,13 +331,27 @@ func getKafkaTopicConfig(d *schema.ResourceData) aiven.KafkaTopicConfig {
...
+	topic, err := getTopic(ctx, d, m)
+
+	// Topics are destroyed when kafka is off
+	// https://docs.aiven.io/docs/platform/concepts/service-power-cycle
+	// So it's better to recreate them, than make user to clear the state manually
+	// Recreates missing topics:
+	// 1. if server returns 404
+	// 2. only for "resource" type (not datasource)
+	// 3. only for resources with existing state, not imports
+	if errors.Is(err, errTopicNotFound) && isResource && !d.IsNewResource() {
```
ivan-savciuc, line 349: Btw, we already have this logic down below in the next if statement:
```
if err != nil {
  return diag.FromErr(schemautil.ResourceReadHandleNotFound(err, d))
}
```

So I think we can drop this IF and moreover we can reuse it in data-source, no need for a separate functions for each case. Data-source read function will not error I think in case of `d.SetId("")`

ivan-savciuc (changes requested):
It seems to be working, I tested creation of 2500 topics: 
```
Apply complete! Resources: 2501 added, 0 changed, 0 destroyed.
```

It is a bit surprising since Aiven has a flaky Kafka Topic API, that may return 404 after creation, but I wasn't able to reproduce it in 2.5k topics test. So I guess we are good to go. 

Please check some of the comments below.

internal/common/client.go
```diff
@@ -0,0 +1,34 @@
+package common
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/aiven/aiven-go-client"
+)
+
+func NewAivenClient() (*aiven.Client, error) {
```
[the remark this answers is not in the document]
byashimov: I think it is better to incapsulate env variable and not hardocode it everywhere.

internal/service/kafka/resource_kafka_topic.go
byashimov: Nice. It didn't work because cache overrides error, I'll try to fix it.

byashimov:
ACC https://github.com/aiven/terraform-provider-aiven/actions/runs/3796247765/jobs/6456520791

ivan-savciuc (dismissed):
LGTM

examples_tests/base.go
```diff
@@ -86,7 +88,7 @@ func randNameGen(format string) func() string {
 }
 
 func newClient(token string) (*aiven.Client, error) {
```
Serpentiel, line 90: lgtm, but please, comment all functions / type definitions (e.g. structs) / top level constants or variables, if any

Serpentiel (changes requested):

examples_tests/base.go
byashimov: Please configure your IDE or tell me what functions needs docs and why. Comments without sense are waste of time.

Serpentiel (approved):
lgtm

PR #1009, opened 2022-12-22
feat(schemautil): normalize ip_filter_object

### LINKED ISSUE #987: `v3.9.0` : existing `ip_filter` gets destroyed because of `ip_filter_object`

Serpentiel:

this PR adds support for normalization of `ip_filter_object`

additionally, it introduces a small fix to the `apiconvert` package which is going to prevent the omit of nested items if the top item has been changed, e.g. if we modify only the `description` field of an `ip_filter_object`, the API expects us to send a full `ip_filter_object` array, instead of only modified items; this should as well be the case for other nested structures

resolves #987

## Why this way

currently, we didn't have a way to utilize the functionality that is provided via `ip_filter_object`

this PR makes `ip_filter_object` to function properly

### DIFF
--- a/internal/schemautil/mutations.go
+++ b/internal/schemautil/mutations.go
@@ -16,18 +16,30 @@ func copySensitiveFields(old, new map[string]interface{}) {
 	}
 }
 
-// normalizeIpFilter compares a list of IP filters set in the oldUserConfig and a sorted version coming from
-// newUserConfig and returns the re-sorted IP filters, such that all matching entries will be in
-// the same order as defined in the oldUserConfig.
+// normalizeIpFilter compares a list of IP filters set in the old user config and a sorted version coming from the new
+// user config and returns the re-sorted IP filters, such that all matching entries will be in the same order as
+// defined in the old user config.
 func normalizeIpFilter(old, new map[string]interface{}) {
-	oldIpFilters, ok := old["ip_filter"].([]interface{})
-	if !ok {
-		return
-	}
+	oldIpFilters, _ := old["ip_filter"].([]interface{})
 
-	newIpFilters, ok := new["ip_filter"].([]interface{})
-	if !ok {
-		return
+	newIpFilters, _ := new["ip_filter"].([]interface{})
+
+	fieldToWrite := "ip_filter"
+
+	if oldIpFilters == nil || newIpFilters == nil {
+		var ok bool
+
+		oldIpFilters, ok = old["ip_filter_object"].([]interface{})
+		if !ok {
+			return
+		}
+
+		newIpFilters, ok = new["ip_filter_object"].([]interface{})
+		if !ok {
+			return
+		}
+
+		fieldToWrite = "ip_filter_object"
 	}
 
 	var normalizedIpFilters []interface{}
@@ -40,7 +52,24 @@ func normalizeIpFilter(old, new map[string]interface{}) {
 		found := false
 
 		for _, n := range newIpFilters {
-			if o == n {
+			// Define two comparison variables to avoid code duplication in the loop.
+			var comparableO interface{}
+
+			var comparableN interface{}
+
+			// If we're dealing with a string format, we need to compare the values directly.
+			if fieldToWrite == "ip_filter" {
+				comparableO = o
+
+				comparableN = n
+			} else {
+				// If we're dealing with an object format, we need to compare the values of the "network" field.
+				comparableO = o.(map[string]interface{})["network"]
+
+				comparableN = n.(map[string]interface{})["network"]
+			}
+
+			if comparableO == comparableN {
 				normalizedIpFilters = append(normalizedIpFilters, o)
 
 				found = true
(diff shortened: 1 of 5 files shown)


Serpentiel:
acceptance tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/3759796259

internal/schemautil/userconfig/apiconvert/toapi_test.go
```diff
@@ -678,7 +683,8 @@ func TestToAPI(t *testing.T) {
 						map[string]interface{}{
 							"namespaces": []interface{}{
 								map[string]interface{}{
-									"retention": "48h",
+									"resolution": "30s",
+									"retention":  "48h",
```
byashimov, line 687: Why you change namespaces tests?

internal/schemautil/mutations.go
```diff
@@ -73,5 +120,5 @@ func normalizeIpFilter(old, new map[string]interface{}) {
 		}
 	}
 
-	new["ip_filter"] = append(normalizedIpFilters, nonexistentIpFilters...)
+	new[fieldToWrite] = append(normalizedIpFilters, nonexistentIpFilters...)
```
byashimov, line 123: Please provide tests for this, with a good coverage.

internal/schemautil/mutations.go
```diff
@@ -30,6 +30,22 @@ func normalizeIpFilter(old, new map[string]interface{}) {
 		return
 	}
 
+	fieldToWrite := "ip_filter"
+
+	if len(oldIpFilters)*len(newIpFilters) == 0 {
```
byashimov, line 35: So if newIpFilters is empty and oldIpFilters is not should write to ip_filter_object?
Is this correct?

internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -169,6 +169,15 @@ func itemToAPI(
...
 	o := !d.HasChange(fks)
 
+	// We need to make sure that if there were any changes to the parent object, we also send the value, even if it
+	// was not changed.
+	//
+	// We check that there are more than three elements in the fk slice, because we don't want to send the value if
+	// the parent object is the root object.
+	if o && len(fk) > 3 && d.HasChange(fks[:strings.LastIndex(fks, ".0")]) {
+		o = false
+	}
```
byashimov, line 179: Please provide tests for this.

internal/schemautil/mutations.go
```diff
@@ -61,7 +94,21 @@ func normalizeIpFilter(old, new map[string]interface{}) {
...
+
+				comparableN = n
+			} else {
+				comparableO = o.(map[string]interface{})["network"]
+
+				comparableN = n.(map[string]interface{})["network"]
+			}
+
+			if comparableO == comparableN {
 				found = true
```
byashimov, line 112: I think this can be easier:
1. create a slice and a map `seen` for old values
2. iterate over old slice and append to new, and `seen`
3. iterate over new slice and if value is not in seen, then append it

byashimov (changes requested):

internal/schemautil/userconfig/apiconvert/toapi_test.go
Serpentiel: because the behavior of the `ToAPI` changed, we had a wrong behavior before

internal/schemautil/mutations.go
Serpentiel: yes

internal/schemautil/userconfig/apiconvert/toapi.go
Serpentiel: https://github.com/aiven/terraform-provider-aiven/pull/1009#discussion_r1055746378

internal/schemautil/mutations.go
Serpentiel: this would be good to consider for refactoring it to work this way in future

internal/schemautil/mutations.go
Serpentiel: done

byashimov (approved):

PR #1011, opened 2022-12-27
feat: add stateupgrader and Grafana PoC

Serpentiel:

- adds a `stateupgrader` package which aims to assist with upgrading from v3.x to v4.0.0
- adds PoC for Grafana

## Why this way

we need to be able to automatically change types in the states of v3.x users


internal/schemautil/userconfig/stateupgrader/typeupgrader/typeupgrader.go
```diff
@@ -0,0 +1,62 @@
...
+		case "int":
+			if va == "" {
+				va = "0"
+			}
+
+			s[i], err = strconv.Atoi(va)
+			if err != nil {
+				return err
+			}
+		}
```
byashimov, line 58: Let's put both "switch" into one function and use it in Slice and Map.

byashimov (approved):

internal/schemautil/userconfig/stateupgrader/typeupgrader/typeupgrader.go
Serpentiel: I'll see if that's possible

PR #1016, opened 2022-12-29
improve team member deletion

byashimov:

If user's email not on the members list terraform won't warn.
With this fix it will return an error.


byashimov:
Can't provide tests, because can't accept invitation code from email in tests.

Serpentiel (approved):
lgtm

PR #1021, opened 2023-01-03
refactor(typeupgrader): move switch to convert fn

Serpentiel:

moves `switch` to `convert` fn

## Why this way

deduplicates code and reduces complexity

### DIFF
--- a/internal/schemautil/userconfig/stateupgrader/typeupgrader/typeupgrader.go
+++ b/internal/schemautil/userconfig/stateupgrader/typeupgrader/typeupgrader.go
@@ -1,6 +1,7 @@
 package typeupgrader
 
 import (
+	"fmt"
 	"strconv"
 )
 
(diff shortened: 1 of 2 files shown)


internal/schemautil/userconfig/stateupgrader/typeupgrader/typeupgrader.go
```diff
@@ -45,18 +29,37 @@ func Slice(s []interface{}, t string) (err error) {
...
+			v = "0"
+		}
+
+		res, err = strconv.Atoi(v)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return res, nil
```
byashimov, line 64: The behaviour has changed. Now it fallbacks for an empty interface if `t` is not `bool` or `int`.  
More important that this hadn't been caught by tests. Please provide a test for the case.

Serpentiel: addressed

byashimov (approved):

PR #1032, opened 2023-01-10
add service integration endpoints sweeper

byashimov:

Adds service integration endpoints sweeper

### DIFF
--- a/internal/service/account/sweep.go
+++ b/internal/service/account/sweep.go
@@ -1,5 +1,4 @@
 //go:build sweep
-// +build sweep
 
 package account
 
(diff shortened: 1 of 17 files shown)


internal/service/account/sweep.go
```diff
@@ -1,5 +1,4 @@
 //go:build sweep
-// +build sweep
```
Serpentiel, line 2: why would we want to remove this directive?

Serpentiel (changes requested):

internal/service/account/sweep.go
byashimov: That is for older [versions](https://stackoverflow.com/a/68361186).

Serpentiel (approved):
lgtm

PR #1040, opened 2023-01-16
refactor(stateupgrader): match with repo struct

Serpentiel:

refactored `stateupgrader` package structure to match overall repository struct introduced by @ivan-savciuc 

## Why this way

to keep things identical


ivan-savciuc (approved):
LGTM

PR #1042, opened 2023-01-17
sweep test-tf prefix too

byashimov:

Sweeps services with `test-tf` prefix

### DIFF
--- a/internal/sweep/sweep.go
+++ b/internal/sweep/sweep.go
@@ -52,7 +52,7 @@ func SweepServices(region, t string) error {
 			continue
 		}
 
-		if !hasPrefixAny(s.Name, "test-acc", "test-examples", "k8s-") {
+		if !hasPrefixAny(s.Name, "test-tf", "test-acc", "test-examples", "k8s-") {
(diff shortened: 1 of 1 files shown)


internal/sweep/sweep.go
```diff
@@ -52,7 +52,7 @@ func SweepServices(region, t string) error {
 			continue
 		}
 
-		if !hasPrefixAny(s.Name, "test-acc", "test-examples", "k8s-") {
+		if !hasPrefixAny(s.Name, "test-tf", "test-acc", "test-examples", "k8s-") {
```
Serpentiel, line 55: nit: perhaps we should move `k8s-` sweeper to the respective repo

Serpentiel (approved):
lgtm

internal/sweep/sweep.go
ivan-savciuc: sounds like a good idea, but we usually have services with k8s prefix in a different project

ivan-savciuc (approved):
LGTM

PR #1046, opened 2023-01-24
feat(userconfig): required fields

Serpentiel:

adds support for required fields to userconfig generator

## Why this way

to have required fields in place

### DIFF
--- a/internal/schemautil/userconfig/handle.go
+++ b/internal/schemautil/userconfig/handle.go
@@ -9,12 +9,17 @@ import (
 )
 
 // handlePrimitiveTypeProperty is a function that converts a primitive type property to a Terraform schema.
-func handlePrimitiveTypeProperty(n string, p map[string]interface{}, t string) map[string]*jen.Statement {
-	return map[string]*jen.Statement{n: jen.Values(convertPropertyToSchema(n, p, t, true))}
+func handlePrimitiveTypeProperty(n string, p map[string]interface{}, t string, ireq bool) map[string]*jen.Statement {
+	return map[string]*jen.Statement{n: jen.Values(convertPropertyToSchema(n, p, t, true, ireq))}
 }
 
 // handleObjectProperty is a function that converts an object type property to a Terraform schema.
-func handleObjectProperty(n string, p map[string]interface{}, t string) (map[string]*jen.Statement, error) {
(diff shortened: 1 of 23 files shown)


internal/schemautil/userconfig/handle.go
```diff
@@ -161,13 +170,19 @@ func handleArrayProperty(n string, p map[string]interface{}, t string) (map[stri
...
 			}
 
-			e, err = handleArrayOfAggregateTypeProperty(ipa)
+			req := map[string]struct{}{}
+
+			if sreq, ok := ia["required"].([]interface{}); ok {
+				req = SliceToKeyedMap(sreq)
+			}
+
+			e, err = handleArrayOfAggregateTypeProperty(ipa, req)
```
byashimov, line 179: Please translate: what are iof, ipa, ia?

internal/schemautil/userconfig/handle.go
```diff
@@ -219,9 +234,15 @@ func handleAggregateTypeProperty(
...
+	req := map[string]struct{}{}
+
+	if sreq, ok := p["required"].([]interface{}); ok {
+		req = SliceToKeyedMap(sreq)
+	}
+
 	switch at {
 	case "object":
-		v, err := handleObjectProperty(n, p, t)
+		v, err := handleObjectProperty(n, p, t, req)
```
byashimov, line 245: If `handleObjectProperty` handles just one property, then it is possible to this:
```go
requiredList, ok := p["required"].([]interface{})

// This is mandatory
if !ok {
    return fmt.Errorf("...")
}

handleObjectProperty(a, b, c, slices.Contains(requiredList, fieldName)
```

internal/schemautil/userconfig/handle.go
Serpentiel: this is planned to be changed in future, after we release v4, but here are the abbreviations used:

`iof` is item one of, `ipa` is item properties asserted, `ia` is item asserted

internal/schemautil/userconfig/handle.go
Serpentiel: it handles multiple properties in downstream call of `convertPropertiesToSchemaMap`

byashimov:
lgtm

PR #1055, opened 2023-02-09
fix(kafka): default_acl

Serpentiel:

fixes `default_acl` field incorrect behavior during import operation

## Why this way

to deliver quality product

### DIFF
--- a/internal/service/kafka/kafka.go
+++ b/internal/service/kafka/kafka.go
@@ -23,10 +23,12 @@ func aivenKafkaSchema() map[string]*schema.Schema {
 		DiffSuppressFunc: schemautil.EmptyObjectDiffSuppressFunc,
 	}
 	aivenKafkaSchema["default_acl"] = &schema.Schema{
-		Type:        schema.TypeBool,
-		Optional:    true,
-		Default:     true,
-		Description: "Create default wildcard Kafka ACL",
+		Type:             schema.TypeBool,
+		Optional:         true,
+		ForceNew:         true,
+		Default:          true,
+		DiffSuppressFunc: schemautil.CreateOnlyDiffSuppressFunc,
+		Description:      "Create default wildcard Kafka ACL",
 	}
 	aivenKafkaSchema[schemautil.ServiceTypeKafka] = &schema.Schema{
 		Type:        schema.TypeList,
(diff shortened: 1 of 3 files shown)


Serpentiel:
@dewan-ahmed could you please check if this resolves your issue?

dewan-ahmed:
Hi @Serpentiel thanks for this change. I tested with the dev version of the provider. With the new change, terraform plan is attempting a force replacement and wants to recreate the resource. I don’t think this is what we intend to do. I have shared a screen-recording of the terraform plan command with you.

internal/service/kafka/kafka.go
```diff
@@ -25,6 +25,7 @@ func aivenKafkaSchema() map[string]*schema.Schema {
 	aivenKafkaSchema["default_acl"] = &schema.Schema{
 		Type:        schema.TypeBool,
 		Optional:    true,
+		ForceNew:    true,
```
ivan-savciuc, line 28: Why do we force recreation of the resource here? I think this filed only makes sense when during resource creation

internal/service/kafka/kafka.go
```diff
@@ -226,5 +227,12 @@ func resourceKafkaRead(ctx context.Context, d *schema.ResourceData, m interface{
 		})
 	}
 
+	if d.IsNewResource() {
```
ivan-savciuc, line 230: I think it is up to the user to set this value. What is the reason why we set it to false here?

ivan-savciuc (changes requested):
Please explain what was the reason for this change?

internal/service/kafka/kafka.go
Serpentiel: it works incorrectly during import

internal/service/kafka/kafka.go
Serpentiel: we cannot have this field edited, editing it should require a resource recreation

ivan-savciuc: Can we make it `Computed` after creation?  Or make it use `CreateOnlyDiffSuppressFunc`?

internal/service/kafka/kafka.go
ivan-savciuc: ok ok, now i get it, I think we should make it computed when imported or basically in any other case other then creation.

ivan-savciuc (approved):
LGTM

dewan-ahmed:
Thanks for reviewing @ivan-savciuc and thanks @Serpentiel for the change.

Adding a note here that I retested and the change is working as expected for `terraform import` 🎉

PR #1061, opened 2023-02-16
fix(schemautil): maintenance_window_dow=never resource update

byashimov:

When `maintenance_window_dow` set to `never` on the backend terraform send it back on updates. And that fails with the validation, because `never` is not available through the API.


Serpentiel (approved):
lgtm

Issue #1054, opened 2023-02-08
Error creating resources: unable to wait for service state change: context deadline exceeded


AaronFriel:
As a maintainer of [Pulumi Aiven](github.com/pulumi/pulumi-aiven), we run tests against the Aiven source code in our provider regularly on cronjobs and on pull requests. Beginning in November, we began seeing intermittent failures with the error:

```
unable to wait for service state change: context deadline exceeded
```

## Suggested resolution

Pulumi's perspective: we recommend removing or significantly increasing timeouts to at least a large multiple of observed 100th percentile times.

Hypothetically, setting latencies to values such as the 99th percentile ensures that 1 in 100 programs with one resource will fail, and a program managing ten resources fails 1 in 10 times (1-.99<sup>10</sup>). When a provider times out, the result is usually an inconsistent state left in the service it manages and this leads to surprising behavior and manual steps to resolve. We believe infra-as-code tools should optimize for reducing manual work, even at the expense of potentially long running CI jobs.

Because of this, Pulumi defaults to not setting a timeout on the providers we author, relying on the user or the CI environment to decide when they have waited too long. This helps keep Pulumi resilient to long tail latencies.

## Steps to reproduce

I believe that either infra-as-code tool would reproduce this error, if you're more familiar with HCL, it's possible to adapt our acceptance test code below to create a similar TF configuration.

This is one of the examples tests that we saw intermittent errors on:

From https://github.com/pulumi/pulumi-aiven/blob/master/examples/service/index.ts
```ts
import * as pulumi from "@pulumi/pulumi";
import * as aiven from "@pulumi/aiven";

const config = new pulumi.Config();
const projectName = config.require("projectName");


const service = new aiven.Grafana("my-new-service", {
    project: projectName,
    cloudName: "google-europe-west1",
    plan:"startup-4",
    grafanaUserConfig: {
        publicAccess: {
            grafana: "true",
        }
    }
});

export const serviceName = service.serviceName;
```

To get started with this example, [install Pulumi](https://www.pulumi.com/docs/get-started/install/) and then in an empty directory run: 

```shell
pulumi new https://github.com/pulumi/pulumi-aiven/examples/service
```

This will create the TypeScript project and its dependencies. You'll need to set the environment variables `AIVEN_PROJECT_NAME` and `AIVEN_TOKEN`, and then you can run:

```
pulumi up
```

## Expected behavior

[...]

PR #1063, opened 2023-02-17
feat(service): unify timeouts

Serpentiel:

this PR unifies the timeout values across all resources so that they are managed in a more obvious way, and uses a constant value so we are able to easily adjust it when/if needed

this PR also keeps an ability for users to manage the timeout values for their resources, and also adds `Default` timeout option to all the resources where it was missing, as well as all the other CRDU timeouts

this PR is linked to and was discovered during #1054

## Why this way

we want to allow our users to manage their timeouts, while both having a reasonable default timeout value, easily modifiable if needed, and unified across all the resources so they become functioning in a more obvious manner for the end users

### DIFF
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aiven/aiven-go-client"
 	"github.com/docker/go-units"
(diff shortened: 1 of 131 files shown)


byashimov (approved):

internal/schemautil/service.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+// DefaultTimeoutMinutes is the default timeout for service operations.
+const DefaultTimeoutMinutes = 20
```
ivan-savciuc, line 20: By the var name I would naturally expect it to be of type `time` with some timeout

ivan-savciuc (dismissed):
LGTM

internal/schemautil/service.go
Serpentiel: this has been changed completely

byashimov (approved):

internal/schemautil/service.go
Serpentiel: please check it again, I moved it to a function completely, also I checked with other open source projects, most use `Minutes` suffix when denoting an integer constant

byashimov:
lgtm

ivan-savciuc (approved):
LGTM

PR #1064, opened 2023-02-17
docs(version): update from v3 to v4

Serpentiel:

updates version from v3 to v4

## Why this way

to have up-to-date docs in place


ivan-savciuc (approved):
LGTM

PR #1066, opened 2023-02-21
ci(lint): add commitlint

ivan-savciuc:

Add commitlint to CI 

## Why this way

Standardize commit message prefixes


Serpentiel (approved):
lgtm

PR #1067, opened 2023-02-21
fix(sweep): use accounts with test prefix

m-novikov:

Fixes not all account sweeping functions operate only on test accounts' objects.

## Why this way

Related accounts objects should only be removed from accounts created by test runs

### DIFF
--- a/internal/service/account/sweep.go
+++ b/internal/service/account/sweep.go
@@ -42,6 +42,22 @@ func init() {
 	})
 }
 
+func listTestAccounts(conn *aiven.Client) ([]aiven.Account, error) {
+	testAccounts := []aiven.Account{}
+	r, err := conn.Accounts.List()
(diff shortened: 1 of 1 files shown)


Serpentiel (changes requested):
could you please fix your commit message to follow https://conventionalcommits.org/en/v1.0.0/ spec?

you are able to simply copy the name of this PR to achieve that

m-novikov:
> could you please fix your commit message to follow https://conventionalcommits.org/en/v1.0.0/ spec?
> 
> you are able to simply copy the name of this PR to achieve that

Yep, thanks just did that :)

Serpentiel:
I'll rebase your branch to `main` and run the acceptance tests to see if this resolves the problem

Serpentiel (approved):
lgtm

PR #1068, opened 2023-02-23
feat: terraform plugin framework

Serpentiel:

adds support for Terraform plugin framework and migrates our existing provider to it

## Why this way

to support Terraform protocol version 6


ivan-savciuc (approved):
Tested with a butch on manifests all working well! :+1:

PR #1080, opened 2023-03-02
chore(timeouts): set default read timeout

byashimov:

Adds missing read timeout. Which also fixes kafka topic acc test.


ivan-savciuc (approved):
LGTM

PR #1085, opened 2023-03-09
refactor: nest internal contents to sdkprovider

Serpentiel:

nests some internal contents to sdkprovider

## Why this way

to have room to introduce the new provider via Plugin Framework


ivan-savciuc (approved):
LGTM

PR #1087, opened 2023-03-14
fix(docs): update flink example version

byashimov:

Fixes flink version in example. 1.15 is not supported anymore.


Serpentiel (approved):
lgtm

PR #1089, opened 2023-03-14
fix(kafkatopic): conflict on create existing topic

byashimov:

Because of retries, kafka topic creation process may get already existing topic.


Serpentiel (approved):
lgtm

PR #1090, opened 2023-03-14
feat(provider): new framework boilerplate

Serpentiel:

adds new framework boilerplate

## Why this way

to begin moving to the new framework


byashimov (approved):
lgtm

PR #1091, opened 2023-03-14
ci(acctest): fix project structure

Serpentiel:

fixes incorrect project structure specified in acctest make target

## Why this way

to have them working


ivan-savciuc (approved):
LGTM

PR #1112, opened 2023-03-27
feat(userconfig): add ip_filter, namespaces string alias fields

Serpentiel:

adds `ip_filter` and `namespaces` string-suffixed alias fields

## Why this way

to provide an alternative before deprecation is in effect


ivan-savciuc (approved):
LGTM

PR #1117, opened 2023-04-04
fix(ip_filter): sort ip filters

byashimov:

Sorts all ip_filters a-z.


ivan-savciuc (commented):
@muralibasani, you are probably planning to do that anyway, but I will say this anyways - can you please add an acceptance test that will show 0 diffs when we reshuffle the IP filters of the service? Unit tests are great, but I would also love to see that in acc tests.

PR #1118, opened 2023-04-04
fix(apiconvert): double apply necessity on migration

### LINKED ISSUE #1073: 4.0.0 pg_user_config.ip_filter_object does not set ips on aiven_pg
## What happened?
Following a deprecated warning on `ip_filter` after migrating from aiven provider 3.5.0 to 4.0.0, 
i change my code from
```
locals {
  aiven_static_ips = toset([
    "a.a.a.a/24",
    "b.b.b.b/24", 
  ])

resource "aiven_pg" "default" {
  ....
  pg_user_config {
    ip_filter = local.aiven_static_ips
  }
}
```
to 
```
resource "aiven_pg" "default" {
  ....
  pg_user_config {
    ip_filter_object {
      network     = "a.a.a.a/24"
      description = "some comment"
    }
    ip_filter_object {
      network     = "b.b.b.b/24"
      description = "some comment"
    }
  }
}
```

As effect, it remove all filters in aiven pg web interface, although there is no error or indication in `terraform apply`
Seems this append with a lot of filter applied

Serpentiel:

fixes double apply necessity on migration from typeless keys to typed keys

resolves #1073 

## Why this way

to eliminate the bug

### DIFF
--- a/internal/schemautil/userconfig/apiconvert/toapi.go
+++ b/internal/schemautil/userconfig/apiconvert/toapi.go
@@ -199,7 +199,7 @@ func itemToAPI(
 
 	fks := strings.Join(fk, ".")
 
-	// We omit the value if has no changes in the Terraform user configuration.
+	// We omit the value if it has no changes in the Terraform user configuration.
 	o := !d.HasChange(fks)
 
 	// We need to make sure that if there were any changes to the parent object, we also send the value, even if it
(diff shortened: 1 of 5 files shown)


internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -270,6 +270,57 @@ func itemToAPI(
 	return res, o, nil
 }
 
+// processManyToOneKeys is a function that processes many to one keys.
+func processManyToOneKeys(res map[string]interface{}, depthIdx int) {
```
ivan-savciuc, line 274: Why do we need the `depthIdx` ? I see that we are using it conclusively like this `processManyToOneKeys(va, depthIdx+1)`, but I do not see it used anywhere else.

internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -270,6 +270,57 @@ func itemToAPI(
...
+		}
+
+		if _, ok := mto[rk]; !ok {
+			mto[rk] = []string{}
+		}
+
+		mto[rk] = append(mto[rk], k)
+	}
+
+	for k, v := range mto {
```
ivan-savciuc, line 305: It seems to me that we can simplify this function by removing this additional for and handling new values in the array from above. WDYT?

internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -270,6 +270,57 @@ func itemToAPI(
 	return res, o, nil
 }
 
+// processManyToOneKeys is a function that processes many to one keys.
```
ivan-savciuc, line 273: Please add another sentence to the comments describing how we are processing this. If I gather it correctly, we set only one value for a key for ip_filters and namespaces regardless of where it was _string or _object.

ivan-savciuc (commented):
Some minor suggestions

internal/schemautil/userconfig/apiconvert/toapi.go
Serpentiel: ah, thanks for noticing this, looks like I used it in previous implementation

internal/schemautil/userconfig/apiconvert/toapi.go
Serpentiel: I checked further, looks like this can't be simplified any further, but thanks for the heads up!

internal/schemautil/userconfig/apiconvert/toapi.go
Serpentiel: added more comments

ivan-savciuc (approved):
LGTM

PR #1119, opened 2023-04-05
fix: add conflicting with logic for namespaces and ip_filters

ivan-savciuc:

IP filters and M3 namespaces have only one field in the API request and response but can be set in 3 different ways using Terraform. However, adding a custom diff logic that triggers an error if a user sets more than one field in Terraform for these two corresponding properties.


Serpentiel (dismissed):
lgtm, but please fix the following `commitlint` error:

```log
✖   body's lines must not be longer than 100 characters [body-max-line-length]
```

Serpentiel (approved):

PR #1121, opened 2023-04-05
ci(lint): drop forced Go version in golangci-lint

Serpentiel:

drops forced Go version in `golangci-lint`

## Why this way

https://github.com/golangci/golangci-lint/issues/2649 was resolved


ivan-savciuc (approved):
LGTM

PR #1135, opened 2023-04-18
fix(mysql): user creation authentication field

ivan-savciuc:

When a MySQL user has no password field set but has an authentication, it won't be propagated to Aiven. Lets fix that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ nav_order: 1
 
 - Added docs and validation for `aiven_service_integration_endpoint`
 - Dropped `signalfx` from supported integration types
+- Fix MySQL user creation authentication field 
 
 ## [4.2.1] - 2023-04-06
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ nav_order: 1
 
 - Added docs and validation for `aiven_service_integration_endpoint`
 - Dropped `signalfx` from supported integration types
+- Fix MySql user creation authentication field
```
Serpentiel, line 13: nit: can we have it "MySQL"?

Serpentiel (approved):
lgtm, but check if this is something you want to change

Serpentiel:
also perhaps `fix(mysql): user creation authentication field` commit/PR message would work better?

Serpentiel (approved):
lgtm

PR #1141, opened 2023-04-26
fix(account): authentication SAML field mapping set

ivan-savciuc:

Fix account authentication SAML field mapping set method


Serpentiel (approved):
lgtm

PR #1143, opened 2023-04-27
docs(static_ip_addresses): instructions for creating static IP addresses

staceysalamon-aiven:
Add instructions for creating static IP addresses.


staceysalamon-aiven:
For docs only updates, is a changelog entry needed? If so, should it go [here](https://github.com/aiven/terraform-provider-aiven/blob/main/CHANGELOG.md#majorminorpatch---yyyy-mm-dd)?

Serpentiel (approved):
lgtm

PR #1147, opened 2023-05-02
fix: adjust generated SQL for ClickHouse privilege grants

kmichel-aiven:

1. Fix `" WITH GRANT"` that should be `" WITH GRANT OPTION"`.

2. When a table is unspecified, write `.*` instead of nothing.
    A query like `GRANT SELECT ON foo TO bar` grants `SELECT` on the table `foo` of the current database of the connection.
    Instead we want to generate `GRANT SELECT ON foo.* TO bar` to grant `SELECT` on all tables of the `foo` database.

3. Adjust the documentation to not suggest using `"*"` as the table name.
    A `"*"` table name does not grant privilege on all tables, it grants privilege on the table named `*`.
    To grant privilege on all tables, the table name should be completely omitted.

Reference: https://clickhouse.com/docs/en/sql-reference/statements/grant#granting-privilege-syntax

### DIFF
--- a/internal/sdkprovider/service/clickhouse/grant.go
+++ b/internal/sdkprovider/service/clickhouse/grant.go
@@ -145,12 +145,14 @@ func createPrivilegeGrantStatement(grant PrivilegeGrant) string {
 
 	if len(grant.Table) > 0 {
 		b.WriteString(fmt.Sprintf(".%s", escape(grant.Table)))
+	} else {
+		b.WriteString(".*")
 	}
 
 	b.WriteString(fmt.Sprintf(" TO %s", escape(userOrRole(grant.Grantee))))
 
 	if grant.WithGrant {
-		b.WriteString(" WITH GRANT")
+		b.WriteString(" WITH GRANT OPTION")
 	}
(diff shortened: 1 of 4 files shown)


internal/sdkprovider/service/clickhouse/grant.go
```diff
@@ -145,12 +145,14 @@ func createPrivilegeGrantStatement(grant PrivilegeGrant) string {
 
 	if len(grant.Table) > 0 {
 		b.WriteString(fmt.Sprintf(".%s", escape(grant.Table)))
+	} else {
+		b.WriteString(".*")
```
Serpentiel, line 149: perhaps we want to have something similar at line 176 too?

Serpentiel (changes requested):

internal/sdkprovider/service/clickhouse/grant.go
kmichel-aiven: We do :+1:

Serpentiel (approved):
lgtm

PR #1150, opened 2023-05-05
docs(guides): fix outdated articles

staceysalamon-aiven:

Updates outdated guide articles and removes the Using Terraform v0.12 article.


Serpentiel (approved):
lgtm

Issue #1142, opened 2023-04-27
Invalid input: 'tls' is a required property


m0ps:
I'm trying to configure Syslog Service Integration, but it fails with the following error:
>Error: 400: {"errors":[{"message":"Invalid 'user_config': Invalid input: 'tls' is a required property","status":400}],"message":"Invalid 'user_config': Invalid input: 'tls' is a required property"} -

Resource configuration:
```hcl
resource "aiven_service_integration_endpoint" "rsyslog" {
  project       = "my-aiven-project"
  endpoint_name = "rsyslog-server"
  endpoint_type = "rsyslog"
  rsyslog_user_config {
    server = "syslogserver.domain.com"
    port   = 5142
    tls    = false
    format = "rfc5424"
  }
}
```
tf and provider version:
```
terraform --version
Terraform v1.3.9
on darwin_arm64
+ provider registry.terraform.io/aiven/aiven v4.2.1
```
Manual configuration via web interface works correct.

It works fine if I'm trying to create a resource with `tls = true`. In the case of changing `tls = true` to `tls = false` after resource creation, the API call fails with the following error:
>Error: 400: {"errors":[{"message":"Invalid 'user_config': Invalid input: 'server' is a required property","status":400}],"message":"Invalid 'user_config': Invalid input: 'server' is a required property"} -
But 'server' is defined.

m0ps:
It seems this issue is closely related to the following one: https://github.com/aiven/terraform-provider-aiven/issues/1028

Serpentiel:
hey, @m0ps! 👋

thanks for reporting this, I'll take a look and will get back to you shortly

m0ps:
The issue with this resource is also reproduced when you are trying to change any service_integration_endpoint configuration parameter from the `rsyslog_user_config` block.

Serpentiel:
hey! 👋

thank you for your follow-up on this matter, I'm taking a look :)

Serpentiel:
hey, @m0ps! 👋 

I investigated this and prepared a PR!

#1154 is supposed to address this issue

Serpentiel:
hey, @m0ps! 👋

this has now been merged, we will create a release soon 😄

PR #1153, opened 2023-05-05
fix(userconfig): required for top level fields

Serpentiel:

- `required` was not generated for top level fields previously
- this PR also updates userconfig options to the latest
  TODO: automate this the same way we have it in https://github.com/aiven/aiven-operator

found while investigating #1142 

## Why this way

to ensure consistency


ivan-savciuc (approved):
LGTM

PR #1156, opened 2023-05-08
feat: rework to api logic

ivan-savciuc:

This is an idea for `toApi` refactoring, similarly reusing this logic we can expand this to the `fromApi`.

This solution is rather more simple than what we have now, please not be nit picky by checking this PR but rather lets use it to discuss an idea in general, we will agree to move forward I will add unit tests to the each function and will spend more time testing this solution, so far it works really well. 

We can compare this to https://github.com/aiven/terraform-provider-aiven/blob/b99e718b65d6227cc56be0e6c26419a11188ae23/internal/schemautil/userconfig/apiconvert/toapi.go almost 500 lines of very complex code vs [internal/schemautil/service_user_config.go](https://github.com/aiven/terraform-provider-aiven/blob/b99e718b65d6227cc56be0e6c26419a11188ae23/internal/schemautil/service_user_config.go)


internal/schemautil/service_user_config.go
```diff
@@ -0,0 +1,98 @@
...
+
+// flattenUserConfigIPFilter convert oneOf ip_filter to API compatible format
+func flattenUserConfigIPFilter(c map[string]interface{}) {
+	if v, ok := c["ip_filter_string"]; ok && !isUserConfigFieldEmpty(v) {
+		c["ip_filter"] = v
+	}
+
+	if v, ok := c["ip_filter_object"]; ok && !isUserConfigFieldEmpty(v) {
+		c["ip_filter"] = v
+	}
```
byashimov, line 97: nit: turns out, we don't have ConflictsWith constraint on those fields.

internal/schemautil/userconfig/toapi/get.go
```diff
@@ -0,0 +1,38 @@
...
+		return ServiceTypeM3dbToAPIBlueprint()
+	case "mysql":
+		return ServiceTypeMysqlToAPIBlueprint()
+	case "opensearch":
+		return ServiceTypeOpensearchToAPIBlueprint()
+	case "pg":
+		return ServiceTypePgToAPIBlueprint()
+	case "redis":
+		return ServiceTypeRedisToAPIBlueprint()
+	}
```
byashimov, line 35: I think, it should has the default statement, and return an error if key does not exist.

internal/schemautil/service_user_config.go
```diff
@@ -0,0 +1,98 @@
...
+	for ck, cv := range c {
+		if isUserConfigFieldEmpty(cv) {
+			continue
+		}
+
+		for bk, bv := range b {
+			// if key from the user_config is not equal to blueprint skip it
+			if ck != bk {
+				continue
+			}
```
byashimov, line 45: Is it possible to use map lookup?
```go
bk, ok := b[ck]
if !ok {continue}
```
That would do O(n + 1) instead of n*m.

internal/schemautil/service_user_config.go
```diff
@@ -0,0 +1,98 @@
...
+
+			if reflect.TypeOf(bv).Kind() == reflect.Slice { // array of objects
+				bv := bv.([]interface{})
+				buf := []interface{}{}
+				for _, s := range cv.([]interface{}) {
+					buf = append(buf, walkBlueprint(s.(map[string]interface{}), bv[0].(map[string]interface{})))
+				}
+
+				m[bk] = buf
+			}
```
byashimov, line 69: This probably should be switch statement.
And sure return errors.

internal/schemautil/service.go
```diff
@@ -488,8 +489,7 @@ func ResourceServiceUpdate(ctx context.Context, d *schema.ResourceData, m interf
 	}
 
 	st := d.Get("service_type").(string)
-
-	cuc, err := apiconvert.ToAPI(userconfig.ServiceTypes, st, d)
+	cuc := fillInServiceUserConfigData(d.Get(st+"_user_config"), toapi.GetToAPIBlueprint(st))
```
byashimov, line 492: Please, keep with error signature.

internal/schemautil/service_user_config.go
```diff
@@ -0,0 +1,98 @@
...
+			continue
+		}
+
+		for bk, bv := range b {
+			// if key from the user_config is not equal to blueprint skip it
+			if ck != bk {
+				continue
+			}
+
+			if bv == "primitive" { // primitive type
```
byashimov, line 47: Since it doesn't really matter what value is,
I would check it for the string type, and write expected type, i.e. float, int, etc.
That would help for debugging.

internal/schemautil/service_user_config.go
```diff
@@ -0,0 +1,98 @@
...
+// walkBlueprint is a function that recursively traverses two maps c and b
+// and generates API compatible user config structure.
+// Where:
+//   - c is a user config that we get from Terraform state using d.Get(...)
+//   - b is a blueprint of the user config API request
+//
+// It iterates over the data from the Terraform and converts, if necessary,
+// to the API-compatible format.
+func walkBlueprint(c map[string]interface{}, b map[string]interface{}) map[string]interface{} {
+	m := make(map[string]interface{})
```
byashimov, line 35: I guess you can set map size here, it won't be larger than `b`.

internal/schemautil/service_user_config.go
byashimov: And probably warn or error. User shouldn't have things not in blueprint.

byashimov (changes requested):
I like it, bravo! Let's do this way.
(I've started review in case you'll proceed with this pr)

internal/schemautil/userconfig/userconfig_test.go
```diff
@@ -14,6 +16,167 @@ import (
...
+		typePrimitive,
+		typeObject,
+		typeArray,
+		typeArrayOfObjects,
+		typeOneOf,
+	}
+}
+
+func listOfPrimitiveTypes() []string {
+	return []string{"boolean", "integer", "string", "number"}
```
byashimov, line 38: You can use a switch, it works without Contains.

internal/schemautil/service_user_config.go
Serpentiel: we [do](https://github.com/aiven/terraform-provider-aiven/blob/9bbd1db4a57facf04dccfc8da3249a2359123d5b/internal/schemautil/custom_diff.go#L168-L233) have them, they're implemented in a different way because `ConflictsWith` does NOT works with arrays/objects

internal/schemautil/userconfig/userconfig_test.go
```diff
@@ -14,6 +16,167 @@ import (
...
+		}
+
+		va, ok := m[k].(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		switch getElementType(va) {
+		case typePrimitive, typeArray, typeOneOf:
+			d[jen.Lit(k)] = jen.Lit("primitive")
```
Serpentiel, line 160: this has to be replaced with `struct{}`: https://medium.com/@l.peppoloni/how-to-improve-your-go-code-with-empty-structs-3bd0c66bc531

or do we plan to have anything else instead of "primitive" over here?

ivan-savciuc: nope, only this

Serpentiel: please, change this then 🙂

PR #1161, opened 2023-05-10
ci: add userconfig-generate-schema.yml workflow

Serpentiel:

adds `userconfig-generate-schema.yml` workflow

## Why this way

currently, when dependabot's PRs are merged, we need to manually run `go generate ./...`

this PR makes the whole update process fully automatic


ivan-savciuc (approved):
LGTM

PR #1166, opened 2023-05-11
ci(userconfig-generate-schema): run make docs

Serpentiel:

also runs `make docs` when generating userconfig schema

## Why this way

it's missing ☹️


ivan-savciuc (approved):
LGTM

PR #1179, opened 2023-05-25
fix(kafka_topic): 404 error handling

ivan-savciuc:

We query Kafka Topics in batches with up to 100 topics simultaneously using the v2 list endpoint. If we add a topic name that does not exist to the query batch, API retrieves a 404 response. Furthermore, it retrieves a 404 even in case all topics in the batch do exist except one.

This API behavior causes problems when Terraform clients accidentally delete one or more topics outside of Terraform and then try to refresh a state. In this case, an nonexistent topic in the state will trigger a 404 response from the API.

In case we get 404 from the v2 list endpoint we will query v1 list (that has all topics for a services but topics data is not fully present in the response) and use to to determine which of the Kafka Topics do not exist and propagate 404 further for these topics only.

### DIFF
--- a/internal/sdkprovider/service/kafka/kafka_topic_wait.go
+++ b/internal/sdkprovider/service/kafka/kafka_topic_wait.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/aiven/aiven-go-client"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"golang.org/x/exp/slices"
 	"golang.org/x/sync/semaphore"
 )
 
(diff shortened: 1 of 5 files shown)


byashimov (dismissed):

Serpentiel (dismissed):

Serpentiel (approved):

internal/sdkprovider/service/kafka/kafka_topic_wait.go
```diff
@@ -96,6 +112,18 @@ func (w *kafkaTopicAvailabilityWaiter) refresh() error {
...
+			// identify the none existing topics.
+			if aiven.IsNotFound(err) {
+				// If topic is missing in V1 list then it does not exist, flagging it as missing
+				for _, t := range queue {
+					if !slices.Contains(c.GetV1List(w.Project, w.ServiceName), t) {
+						c.DeleteFromQueueAndMarkMissing(w.Project, w.ServiceName, t)
+					}
+				}
+				return nil
+			}
```
byashimov, line 126: I think if you put this right before line 113 that would safe you the API call.
Especially if you try to create multiple topics: if I get it right you'll get 404 first, and then it goes to be deleted.

PR #1193, opened 2023-06-05
fix(docs): mark available_credits as read-only

byashimov:

Removes `optional:true` so docs rendered correctly.


Serpentiel (approved):
lgtm

PR #1194, opened 2023-06-05
fix(apiconvert): not being able to unset ip_filter and alike

Serpentiel:

fixes not being able to unset `ip_filter` and similar user config option properties of `array` type

## Why this way

🚀


ivan-savciuc (approved):
LGTM

PR #1197, opened 2023-06-06
chore(kafka_topic): deprecating unclean_leader_election_enable config option

ivan-savciuc:

Deprecating `aiven_kafka_topic.unclean_leader_election_enable`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ nav_order: 1
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
 - Fix not being able to be set `ip_filter` and similar array fields in user config options to an empty array
+- `aiven_kafka_topic` field `unclean_leader_election_enable` is deprecated
 
 ## [4.4.1] - 2023-06-01
 
--- a/docs/resources/kafka_topic.md
+++ b/docs/resources/kafka_topic.md
@@ -23,7 +23,6 @@ resource "aiven_kafka_topic" "mytesttopic" {
 
   config {
     flush_ms                       = 10
-    unclean_leader_election_enable = true
     cleanup_policy                 = "compact,delete"
   }
 
@@ -85,7 +84,7 @@ Optional:
 - `segment_index_bytes` (String) segment.index.bytes value
 - `segment_jitter_ms` (String) segment.jitter.ms value
 - `segment_ms` (String) segment.ms value
-- `unclean_leader_election_enable` (Boolean) unclean.leader.election.enable value
+- `unclean_leader_election_enable` (Boolean, Deprecated) unclean.leader.election.enable value; This field is deprecated and no longer functional.
 
 
 <a id="nestedblock--tag"></a>
--- a/internal/schemautil/userconfig/stateupgrader/v0/kafka/kafka_topic.go
+++ b/internal/schemautil/userconfig/stateupgrader/v0/kafka/kafka_topic.go
@@ -206,9 +206,10 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 				},
 				"unclean_leader_election_enable": {
 					Type:             schema.TypeString,
-					Description:      "unclean.leader.election.enable value",
+					Description:      "unclean.leader.election.enable value; This field is deprecated and no longer functional.",
 					Optional:         true,
 					DiffSuppressFunc: schemautil.EmptyObjectDiffSuppressFunc,
+					Deprecated:       "This field is deprecated and no longer functional.",
 				},
 			},
 		},
@@ -250,11 +251,9 @@ func ResourceKafkaTopicStateUpgrade(
 	}
 
 	err := typeupgrader.Map(config, map[string]string{
-		"message_downconversion_enable":  "bool",
-		"min_cleanable_dirty_ratio":      "float",
-		"preallocate":                    "bool",
-		"unclean_leader_election_enable": "bool",
-	})
+		"message_downconversion_enable": "bool",
+		"min_cleanable_dirty_ratio":     "float",
+		"preallocate":                   "bool"})
 	if err != nil {
 		return rawState, err
 	}
--- a/internal/sdkprovider/service/kafka/kafka_topic.go
+++ b/internal/sdkprovider/service/kafka/kafka_topic.go
@@ -213,9 +213,10 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 				},
 				"unclean_leader_election_enable": {
 					Type:             schema.TypeBool,
-					Description:      "unclean.leader.election.enable value",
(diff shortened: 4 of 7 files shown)


internal/schemautil/userconfig/stateupgrader/v0/kafka/kafka_topic.go
```diff
@@ -209,6 +209,7 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 					Description:      "unclean.leader.election.enable value",
 					Optional:         true,
 					DiffSuppressFunc: schemautil.EmptyObjectDiffSuppressFunc,
+					Deprecated:       "This field is deprecated and won't be available starting from the next major release.",
```
Serpentiel, line 212: I've had an internal talk with @eliax1996, perhaps we would need additional input from them in this GitHub thread on whether this is getting removed or will it still be accepted by the API?

if it's getting removed, we will have to remove it as well, as we have adjusted our development strategy previously to remove fields which no longer work with the API itself

Serpentiel (changes requested):
couple of threads to talk thru

internal/schemautil/userconfig/stateupgrader/v0/kafka/kafka_topic.go
eliax1996: We are discussing on the strategy [here](https://aiven.atlassian.net/jira/software/c/projects/HH/boards/61?modal=detail&selectedIssue=HH-1012&assignee=642a0356d774ab7297295922) we want to get rid of this property as soon as possible because it's hard to understand and it's creating a lot of escalations. We will enable our customers only after chatting with us and we will do that only internally by the prune administrative shell.

TLDR: long term customers won't be able to modify this property directly, right now we are planning to remove it from Terraform, the K8s operator, the CLI and the website. We have to discuss how to move on that change also over the `.sh` kafka management scripts and on the MM

eliax1996: I have a meeting today just to confirm the strategy, I will come back today on that topic! Let me confirm that we can drop it, in that case we can close the mr and delete directly the field

Serpentiel: thanks a lot, please let us know if this is getting removed at the end of this talk, and we'll get another PR which removes it instead of deprecating it ready then :)

eliax1996: confirmed, this is the plan. Official message [here](https://aiven.atlassian.net/browse/HH-1012?focusedCommentId=159153)

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ nav_order: 1
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
-- Fix not being able to be set `ip_filter` and similar array fields in user config options to an empty array
+- `aiven_kafka_topic.unclean_leader_election_enable` field is
```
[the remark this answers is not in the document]
ivan-savciuc: auto rebase issue I guess (did that using GitHub)

docs/resources/kafka_topic.md
```diff
@@ -85,7 +84,7 @@ Optional:
 - `segment_index_bytes` (String) segment.index.bytes value
 - `segment_jitter_ms` (String) segment.jitter.ms value
 - `segment_ms` (String) segment.ms value
-- `unclean_leader_election_enable` (Boolean) unclean.leader.election.enable value
+- `unclean_leader_election_enable` (Boolean, Deprecated) unclean.leader.election.enable value; This field is deprecated and no longer functional.
```
eliax1996, line 87: shouldn't we directly delete the field instead of deprecating it?

internal/sdkprovider/service/kafka/kafka_topic.go
```diff
@@ -213,9 +213,10 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 				},
 				"unclean_leader_election_enable": {
 					Type:             schema.TypeBool,
-					Description:      "unclean.leader.election.enable value",
+					Description:      "unclean.leader.election.enable value; This field is deprecated and no longer functional.",
```
eliax1996, line 216: same doubt here

internal/schemautil/userconfig/stateupgrader/v0/kafka/kafka_topic.go
```diff
@@ -206,9 +206,10 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 				},
 				"unclean_leader_election_enable": {
 					Type:             schema.TypeString,
-					Description:      "unclean.leader.election.enable value",
+					Description:      "unclean.leader.election.enable value; This field is deprecated and no longer functional.",
```
eliax1996, line 209: and here

eliax1996 (changes requested):
seems fine! My only doubt is if we need to mark it as deprecated or remove it directly. But from what I saw if I'm not mistaken the field isn't inserted in the json request, in that case the implementation is totally right and you can mark as resolved my points 😄 

PS it would be a good idea to insert some sort of integration tests just to check that the examples are working. We can work on that together if I can help you, I was imagining something were we launch almond just with the schema validation without really running the side effects just to check that at least the schema validation is sync between terraform and aiven-core, let me know if the idea seems useful!

Thanks for the pr

ivan-savciuc:
> seems fine! My only doubt is if we need to mark it as deprecated or remove it directly. But from what I saw if I'm not mistaken the field isn't inserted in the json request, in that case the implementation is totally right and you can mark as resolved my points smile
> 
> PS it would be a good idea to insert some sort of integration tests just to check that the examples are working. We can work on that together if I can help you, I was imagining something were we launch almond just with the schema validation without really running the side effects just to check that at least the schema validation is sync between terraform and aiven-core, let me know if the idea seems useful!
> 
> Thanks for the pr

@eliax1996 we are deprecating this field, but we're also completely removing it from the TF API request/response and all the assassinated logic. We prefer doing it this way instead of removing it from resource because it will cause trouble for the customers that are using it; if case we had dropped it, they would have to edit their TF manifest imminently because TF clients would stop all the operations saying this field does not exist. Not to say that it is going too far from the HashiCorp TF provider's best practices. 

So the plan is to deprecate it, saying in the deprecation message that this field is no longer functional. And in the next major release, this field will be gone.

docs/resources/kafka_topic.md
ivan-savciuc: Please check my comment bellow

eliax1996:
Good to know, if the requests are without the field everything it’s fine! Thanks for the explanation

eliax1996 (approved):

Serpentiel (approved):
lgtm

PR #1206, opened 2023-06-14
chore(aiven_account_team_project): un-deprecate

Serpentiel:

un-deprecates `aiven_account_team_project`

## Why this way

we should deprecate it when there's an alternative


ivan-savciuc (approved):
lgtm

PR #1217, opened 2023-06-22
chore: use client's retries

byashimov:

- Uses client's retries for corner cases
- Fixes KafkaConector create (retries 404 on POST)
- Fixes VPC test


Serpentiel (approved):
lgtm

PR #1218, opened 2023-06-22
fix(organization): wrong ID used

Serpentiel:

fixes wrong ID used for `organization` resources

## Why this way

the `aiven_organization` resource should have `organization_id` used as its ID, while still being able to create nested units, which require organization's account ID

this ID is more important to us since it's used in other organization-related resources (which aren't implemented yet), thus abolishing the need in organization's account ID

### DIFF
--- a/internal/sdkprovider/service/organization/organization.go
+++ b/internal/sdkprovider/service/organization/organization.go
@@ -62,7 +62,7 @@ func resourceOrganizationCreate(ctx context.Context, d *schema.ResourceData, m i
 		return diag.FromErr(err)
 	}
 
-	d.SetId(r.Account.Id)
+	d.SetId(r.Account.OrganizationId)
 
 	return resourceOrganizationRead(ctx, d, m)
 }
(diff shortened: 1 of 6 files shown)


internal/sdkprovider/service/organization/util.go
```diff
@@ -0,0 +1,28 @@
...
+		r, err := client.Organizations.Get(id)
+
+		if err != nil {
+			return "", err
+		}
+
+		id = r.AccountID
+	}
+
+	return id, nil
```
byashimov, line 27: can it use just [HasPrefix](https://pkg.go.dev/strings#HasPrefix) instead of refexp?

internal/sdkprovider/service/organization/organization.go
```diff
@@ -94,23 +99,32 @@ func resourceOrganizationRead(_ context.Context, d *schema.ResourceData, m inter
 func resourceOrganizationUpdate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
 	client := m.(*aiven.Client)
 
-	r, err := client.Accounts.Update(d.Id(), aiven.Account{
+	id, err := normalizeID(client, d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	r, err := client.Accounts.Update(id, aiven.Account{
```
byashimov, line 107: This one probably need to be wrapped with `ResourceReadHandleNotFound`.

byashimov (approved):
lgtm

PR #1222, opened 2023-06-26
test(vpc): fix free tier error

Serpentiel:

fixes free tier error in VPC tests

## Why this way

make VPC tests better


byashimov (approved):
lgtm

PR #1234, opened 2023-06-30
fix(kafka_schema): compatibility_level import shows non empty plan

byashimov:

Kafka schema `compatibility_level` is handled in a special way. 
If it is not set by user it is not stored in state.
However, import command get's empty resource with empty compatibility_level.
Then when `plan` is called terraform shows the diff: now it reads compatibility_level value but the state doesn't have it.

### DIFF
--- a/internal/schemautil/schemautil.go
+++ b/internal/schemautil/schemautil.go
@@ -1,11 +1,13 @@
 package schemautil
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aiven/aiven-go-client"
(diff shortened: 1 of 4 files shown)


Serpentiel (commented):
could you please add this new function to `ResourceReadHandleNotFound`? thank you!

internal/schemautil/schemautil.go
```diff
@@ -336,3 +338,30 @@ func unmarshalUserConfig(src interface{}) ([]map[string]interface{}, error) {
...
+// The terraform import command can only import one resource at a time
+// https://developer.hashicorp.com/terraform/cli/import/usage
+var (
+	importingResourceID         string
+	importingResourceIDSentinel int32
+)
+
+// ImportStatePassthroughContext same as schema.ImportStatePassthroughContext
+// but stores importing resource id to now if Read() handler has been called on `terraform import` command
+func ImportStatePassthroughContext(ctx context.Context, d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
```
ivan-savciuc, line 352: It looks to me like an overkill, I think we can simplify this to something like: 
```
StateContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
	if err := d.Set("compatibility_level", "importing"); err != nil {
		return nil, aiven.ErrInvalidHost
	}
	return schema.ImportStatePassthroughContext(ctx, d, m)
},
``` 

WDYT?

ivan-savciuc (commented):
Can we also extend this PR with the import test?

ivan-savciuc:
Also please re-base a squash before merging

byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/5474771005/jobs/9969971968#step:6:154)

byashimov:
> could you please add this new function to `ResourceReadHandleNotFound`? thank you!

Let's have a ticket for that. Just adding won't help, we need to replace all `ImportStatePassthroughContext` functions, which will produce enormous diff.

Serpentiel (approved):
lgtm

PR #1242, opened 2023-07-05
chore(project): rename owner_entity_id to parent_id

Serpentiel:

rename `owner_entity_id` to `parent_id`

n.b. we are safe to do that because it wasn't released yet

## Why this way

it was mistakenly called this way


byashimov (approved):
lgtm

PR #1243, opened 2023-07-06
chore(go): use same go 1.18 everywhere

byashimov:

Linter fails to provide output because of version difference.


byashimov:
Linter fails on 1.18

PR #1251, opened 2023-07-10
chore(opensearch): rename Opensearch to OpenSearch

ivan-savciuc:
rename Opensearch to OpenSearch


docs/resources/kafka_connector.md
```diff
@@ -20,7 +20,7 @@ resource "aiven_kafka_connector" "kafka-os-con1" {
 
   config = {
     "topics"              = aiven_kafka_topic.kafka-topic1.topic_name
-    "connector.class" : "io.aiven.kafka.connect.opensearch.OpensearchSinkConnector"
+    "connector.class" : "io.aiven.kafka.connect.opensearch.OpenSearchSinkConnector"
```
Serpentiel, line 23: this is going to break, this field is case sensitive

examples/resources/aiven_kafka_connector/resource.tf
```diff
@@ -5,7 +5,7 @@ resource "aiven_kafka_connector" "kafka-os-con1" {
 
   config = {
     "topics"              = aiven_kafka_topic.kafka-topic1.topic_name
-    "connector.class" : "io.aiven.kafka.connect.opensearch.OpensearchSinkConnector"
+    "connector.class" : "io.aiven.kafka.connect.opensearch.OpenSearchSinkConnector"
```
Serpentiel, line 8: same problem here

Serpentiel (changes requested):

PR #1259, opened 2023-07-14
feat(vpc): gcp privatelink

Serpentiel:

adds GCP private service connect (Privatelink) support

## Why this way

to make it happen :)


ivan-savciuc (approved):
LGTM

PR #1277, opened 2023-07-27
fix(frameworkprovider): failing build

Serpentiel:

fix failing build of the framework provider

## Why this way

we need to have it working


internal/sdkprovider/provider/provider.go
```diff
@@ -36,10 +36,11 @@ func Provider(version string) *schema.Provider {
 		Schema: map[string]*schema.Schema{
 			"api_token": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
```
Serpentiel, line 39: because `DefaultFunc` exists, this becomes `true` in runtime, I have changed it for clarity

PR #1281, opened 2023-08-07
Change tf plan feedback for service args `plan` and `disk_usage`

jclarysse:
Make the feedback of command `terraform plan` more explicit for service resource arguments `plan` and `disk_usage`.

[PM-3406]


Serpentiel (approved):
lgtm

PR #1282, opened 2023-08-08
feat(org): add user group member support

ivan-savciuc:

Adding organization user group member support.

I cannot add acceptance tests for this feature because the `user_id` of the newly invited member will be known only after an invitation is accepted. And we cannot simulate this during the test. Moreover, if we invite already existing Aiven users to a different organization, `user_id` is still unavailable, and we have to accept an invitation.


internal/sdkprovider/service/organization/organization_user_group_member.go
```diff
@@ -0,0 +1,130 @@
...
+
+	orgID, grID, userID, err := schemautil.SplitResourceID3(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	err = client.OrganizationUserGroupMembers.Modify(
+		orgID, grID,
+		aiven.OrganizationUserGroupMemberRequest{
+			Operation: "remove_members",
```
byashimov, line 121: Probably unrelated, but I think this constant should by provided by the go client.

internal/sdkprovider/service/organization/organization_user_group_member.go
```diff
@@ -0,0 +1,130 @@
...
+func resourceOrganizationUserGroupMemberDelete(_ context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	client := m.(*aiven.Client)
+
+	orgID, grID, userID, err := schemautil.SplitResourceID3(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	err = client.OrganizationUserGroupMembers.Modify(
+		orgID, grID,
```
byashimov, line 119: Could you turn it into single column please?

internal/sdkprovider/service/organization/organization_user_group_member.go
Serpentiel: good point, I agree

internal/sdkprovider/service/organization/organization_user_group_member.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func resourceOrganizationUserGroupMemberCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	client := m.(*aiven.Client)
+
+	orgID := d.Get("organization_id").(string)
+	grID := d.Get("group_id").(string)
+	userID := d.Get("user_id").(string)
+	err := client.OrganizationUserGroupMembers.Modify(
+		orgID, grID,
```
byashimov, line 62: Please make it one column.

docs/data-sources/organization_user_group.md
```diff
@@ -24,5 +24,6 @@ The Organization User Groupe data source provides information about the existing
 
 - `create_time` (String) Time of creation
 - `description` (String) The organization user group description. This property cannot be changed, doing so forces recreation of the resource.
+- `group_id` (String) Organization user group ID
```
ngilles-aiven, line 27: I'm not intimately familiar with Terraform internals, so I'm not sure what the difference between is between this `group_id` and the `id` below?

docs/resources/organization_user_group.md
```diff
@@ -28,6 +28,7 @@ The Organization User Group resource allows the creation and management of an Ai
 ### Read-Only
 
 - `create_time` (String) Time of creation
+- `group_id` (String) Organization user group ID
```
ngilles-aiven, line 31: Same as above.

ngilles-aiven (commented):
Something didn't sit quite right after going over this, but I couldn't quite put my finger on it, but it looks like it's because user management within the Aiven platform being not a great fit Terraform as fully declarative system.

I also am not fully sure why it seems to be more complicated in the case of organization users (and groups) than it was with teams, and it seems it comes down to managing things with user ids instead of email? (as invites are done with emails).

Maybe this warrants more discussion on the overall approach or at least the need to extra api/functionality?

docs/data-sources/organization_user_group.md
Serpentiel, line 1: can we please have the `Groupe` typo replaced with `Group` in this file? thank you!

Serpentiel (changes requested):

ivan-savciuc:
@ngilles-aiven and @Serpentiel sorry for not been active on this branch, it was blocked by the tests inactive for the organization package but now I'm going to revive this branch.

PR #1292, opened 2023-08-10
fix(kafka_connector): set a minite timeout for create retries

byashimov:

Sets a minute for KafkaConnector create retries.


Serpentiel (approved):
lgtm

PR #1300, opened 2023-08-18
fix(kafka): fix kafka schema update

ivan-savciuc:

 fix kafka schema update


byashimov (approved):
lgtm

PR #1316, opened 2023-08-24
fix(frameworkprovider): remove defaultfunc workaround

Serpentiel:

partially reverts workaround introduced in #1311

## Why this way

we need to use `DefaultFunc` properly


Serpentiel:
suppressed by #1317

PR #1325, opened 2023-08-28
refactor(kafka): split kafka package to run tests faster

byashimov:

Splits `kafka` package to run tests faster.


Serpentiel (approved):
lgtm

PR #1330, opened 2023-08-30
fix(service): ip filters normalization

ivan-savciuc:

Fix ip filters normalization


Serpentiel (approved):
lgtm

PR #1331, opened 2023-08-30
docs(clickhouse): highlight how to grant privilege on all tables

kmichel-aiven:

It's not immediately obvious that the table field should be omitted, instead of specifying a "*" table.
Add a phrase to highlight that trap.

## Why this way

This can reduce user frustration.


Serpentiel (dismissed):
lgtm, remove the changelog entry and run `make docs`

Serpentiel (approved):
lgtm

PR #1341, opened 2023-09-06
feat(opensearch): security plugin

Serpentiel:

adds Open Search Security Plugin support

## Why this way

to support the new feature


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ nav_order: 1
 - Fix service IP filters normalization
 - Fix improper omitting in `ToAPI`
 - Fix Kafka Topic perfomance
+- OpenSearch Security Plugin support
```
byashimov, line 16: nit: Add `aiven_opensearch_security_plugin_config` resource (OpenSearch Security Plugin support)

internal/sdkprovider/service/opensearch/opensearch_security_plugin_config.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+// aivenOpenSearchSecurityPluginConfigSchema holds the schema for the OpenSearch Security Plugin Config resource.
+var aivenOpenSearchSecurityPluginConfigSchema = map[string]*schema.Schema{
+	"project":      schemautil.CommonSchemaProjectReference,
+	"service_name": schemautil.CommonSchemaServiceNameReference,
+	"admin_password": {
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "The password for the os-sec-admin user.",
```
byashimov, line 21: Should be marked sensitive.

internal/sdkprovider/service/opensearch/opensearch_user.go
```diff
@@ -50,3 +57,84 @@ func ResourceOpenSearchUser() *schema.Resource {
...
+	if err := detourSecurityPluginEnabledCheck(d, m); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return schemautil.ResourceServiceUserCreate(ctx, d, m)
+}
+
+// resourceOpenSearchUserRead reads a OpenSearch User into the Terraform state.
+func resourceOpenSearchUserRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	_ = schemautil.ResourceServiceUserRead(ctx, d, m)
```
byashimov, line 91: Why do you ignore the error? Or should it go after "enabled" check?

byashimov (changes requested):

internal/sdkprovider/service/opensearch/opensearch_user.go
Serpentiel: it is ignored because of the way OpenSearch Security Plugin works

Serpentiel: because of the way the OpenSearch Security Plugin works, once it's enabled, there will be an error when attempting to send this `GET` request to retrieve the error, and we want to prevent any errors from happening as we discussed earlier with the team; I will, however, update it now to add a validation for a specific error that should be skipped

byashimov (approved):
Lgtm

PR #1342, opened 2023-09-06
feat(plugin): add new user config generator

byashimov:

- Adds user config generator for Terraform plugin framework
- Migrates ServiceIntegration to the framework


internal/plugin/service/serviceintegration/service_integration_data_source.go
```diff
@@ -0,0 +1,140 @@
...
+		},
+	}
+}
+
+// Read reads datasource
+// All functions adapted for resourceModel, so we use it as donor
+// Copies state from datasource to resource, then back, when things are done
+func (s *serviceIntegrationDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var o dataSourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &o)...)
```
Serpentiel, line 112: use `util.PlanStateToModel`

internal/plugin/service/serviceintegration/service_integration_data_source.go
```diff
@@ -0,0 +1,140 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	err = copier.Copy(&o, &res)
+	if err != nil {
+		resp.Diagnostics.AddError("dto copy error", err.Error())
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, o)...)
```
Serpentiel, line 139: use `util.ModelToPlanState`

internal/plugin/service/serviceintegration/service_integration_resource.go
```diff
@@ -0,0 +1,343 @@
...
+			"kafka_mirrormaker_user_config":               kafkamirrormaker.NewResourceSchema(),
+			"logs_user_config":                            logs.NewResourceSchema(),
+			"metrics_user_config":                         metrics.NewResourceSchema(),
+		},
+	})
+}
+
+func (s *serviceIntegrationResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var o resourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &o)...)
```
Serpentiel, line 153: use `util.PlanStateToModel`

internal/plugin/service/serviceintegration/service_integration_resource.go
```diff
@@ -0,0 +1,343 @@
...
+	if err != nil {
+		diags.AddError(errmsg.SummaryErrorReadingResource, err.Error())
+		return
+	}
+
+	loadFromDTO(ctx, diags, o, dto)
+	if diags.HasError() {
+		return
+	}
+	diags.Append(state.Set(ctx, o)...)
```
Serpentiel, line 278: use `util.ModelToPlanState`

internal/plugin/service/serviceintegration/userconfig.go
```diff
@@ -0,0 +1,134 @@
...
+	externalAwsCloudwatchMetricsType = "external_aws_cloudwatch_metrics"
+	kafkaConnectType                 = "kafka_connect"
+	kafkaLogsType                    = "kafka_logs"
+	kafkaMirrormakerType             = "kafka_mirrormaker"
+	logsType                         = "logs"
+	metricsType                      = "metrics"
+	readReplicaType                  = "read_replica"
+)
+
+var integrationTypes = []string{
```
Serpentiel, line 35: change to:

```go
func integrationTypes() []string {
    return []string{"..."}
}
```

to avoid mutability

internal/schemautil/plugin.go
```diff
@@ -0,0 +1,153 @@
...
+	b, err := json.Marshal(&src)
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(b, dst)
+}
+
+// ValueStringPointer checks for "unknown"
+// Returns nil instead of zero value
+func ValueStringPointer(v types.String) *string {
```
Serpentiel, line 121: consider this approach instead:

```go
type Valuer interface {
    IsUnknown() bool
    IsNull() bool
}

type StringPointer interface {
    Valuer
    ValueStringPointer() *string
}

type BoolPointer interface {
    Valuer
    ValueBoolPointer() *bool
}

type Int64Pointer interface {
    Valuer
    ValueInt64Pointer() *int64
}

type Float64Pointer interface {
    Valuer
    ValueFloat64Pointer() *float64
}

func ValuePointer[T any](v T, getter func(T) interface{}) interface{} {
    valuer, ok := v.(Valuer)
    if !ok || valuer.IsUnknown() || valuer.IsNull() {
        return nil
    }
    return getter(v)
}

func ValueStringPointer(v StringPointer) *string {
    return ValuePointer(v, func(v StringPointer) interface{} {
        return v.ValueStringPointer()
    }).(*string)
}

func ValueBoolPointer(v BoolPointer) *bool {
    return ValuePointer(v, func(v BoolPointer) interface{} {
        return v.ValueBoolPointer()
    }).(*bool)
}

func ValueInt64Pointer(v Int64Pointer) *int64 {
    return ValuePointer(v, func(v Int64Pointer) interface{} {
        return v.ValueInt64Pointer()
    }).(*int64)
}

func ValueFloat64Pointer(v Float64Pointer) *float64 {
    return ValuePointer(v, func(v Float64Pointer) interface{} {
        return v.ValueFloat64Pointer()
    }).(*float64)
}
```

ucgenerator/main.go
```diff
@@ -0,0 +1,598 @@
...
+		}
+	}
+	return s
+}
+
+func getValidator(name string, v any) *jen.Statement {
+	return jen.Qual(importListvalidator, name).Call(jen.Lit(v))
+}
+
+func valSizeAtLeast(n int) *jen.Statement {
```
Serpentiel, line 580: these functions could be moved to separate `util.go` file

ucgenerator/models.go
```diff
@@ -0,0 +1,142 @@
...
+	// Root item and properties should have max constraint
+	if o.Type == objectTypeObject {
+		if o.isRoot || o.parent != nil && o.parent.Type == objectTypeObject {
+			o.MaxItems = toPtr(1)
+		}
+	}
+}
+
+// toCamelCase some fields has dots within, makes cleaner camelCase
+func toCamelCase(s string) string {
```
Serpentiel, line 136: could be moved to `util.go`

Serpentiel (commented):
- in all getter functions, drop `get` from the name, in Go it's a bad practice
- drop `DTO` prefixes everywhere, in Go this is an anti-pattern

Serpentiel (changes requested):
above review

docs/resources/service_integration.md
```diff
@@ -72,15 +72,15 @@ Required:
 Optional:
 
 - `auto_offset_reset` (String) Action to take when there is no initial offset in offset store or the desired offset is out of range. The default value is `earliest`.
-- `columns` (Block List, Max: 100) Table columns. (see [below for nested schema](#nestedblock--clickhouse_kafka_user_config--tables--columns))
```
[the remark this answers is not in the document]
byashimov: Validators now are functions, not attributes. I believe that's why it can't get docs now for these things.

byashimov:
> * in all getter functions, drop `get` from the name, in Go it's a bad practice
> * drop `DTO` prefixes everywhere, in Go this is an anti-pattern

There are structs for dto, terraform objects, getters, setters etc. With the same name. I need to distinguish them somehow.
The `get` prefix is convention for getter _method_, [not function](https://go.dev/doc/effective_go#Getters).

internal/plugin/service/serviceintegration/service_integration_data_source.go
byashimov: > use `util.ModelToPlanState`

`ModelToPlanState` does same thing without any extra profit. It returns bool value, that I don't need at all. I think `ModelToPlanState` is a bad idea.

docs/resources/service_integration.md
```diff
@@ -33,17 +33,17 @@ resource "aiven_service_integration" "my_integration_metrics" {
 
 ### Optional
 
-- `clickhouse_kafka_user_config` (Block List, Max: 1) ClickhouseKafka user configurable settings (see [below for nested schema](#nestedblock--clickhouse_kafka_user_config))
-- `clickhouse_postgresql_user_config` (Block List, Max: 1) ClickhousePostgresql user configurable settings (see [below for nested schema](#nestedblock--clickhouse_postgresql_user_config))
-- `datadog_user_config` (Block List, Max: 1) Datadog user configurable settings (see [below for nested schema](#nestedblock--datadog_user_config))
+- `clickhouse_kafka_user_config` (Block List) Integration user config (see [below for nested schema](#nestedblock--clickhouse_kafka_user_config))
```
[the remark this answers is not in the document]
byashimov: We get new doc because of validators. They now not printed out. Or I might be missing something.

internal/plugin/service/serviceintegration/service_integration_data_source.go
byashimov: That's useless function. It does nothing but makes you read it and match variable names to understand what it does.

Serpentiel: it saves the amount of lines that is necessary to implement a `Read` method, and it also has a comprehensive comment on it, so one wouldn't need to jump in it to understand its usage, if you believe that the naming is not good, feel free to rename it in this PR, otherwise I see no reason why shouldn't it be reused

internal/plugin/service/serviceintegration/service_integration_data_source.go
Serpentiel: Cf. `organization.go` to understand the purpose of the boolean return value

Serpentiel:
> There are structs for dto, terraform objects, getters, setters etc. With the same name. I need to distinguish them somehow.

`tfoTables` should be called this way instead: `<name>ResourceModel` or `<name>DataSourceModel`, Cf. [`organization_resource.go`](https://github.com/aiven/terraform-provider-aiven/blob/main/internal/plugin/service/organization/organization_resource.go#L41C1-L55C2)

> The get prefix is convention for getter method, [not function](https://go.dev/doc/effective_go#Getters).

how is this any different? please research online why `get` prefix is not used, it's for a reason, e.g. here are two good posts:

- https://stackoverflow.com/questions/45189126/go-getter-methods-vs-fields-proper-naming
- https://github.com/99designs/gqlgen/issues/2359

Serpentiel:
> We are still determining how long v5 will take to release

the changes that are proposed in this PR are safe to release with v4.x

internal/plugin/service/serviceintegration/service_integration_data_source.go
```diff
@@ -0,0 +1,140 @@
...
+	resp.Diagnostics.Append(req.Config.Get(ctx, &o)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var res resourceModel
+	err := copier.Copy(&res, &o)
+	if err != nil {
+		resp.Diagnostics.AddError("data config copy error", err.Error())
+	}
```
byashimov, line 121: Please, take a look at this: with the new plugin framework we now work with different types for resources and datasources. That makes us either use interfaces per field or duplicate all service functions. Or, as I did here, I use resource as a "donor". All functions are adopted for "resource" then I just _copy_ values to the datasource.

internal/plugin/service/serviceintegration/service_integration_data_source.go
```diff
@@ -0,0 +1,140 @@
...
+// Copies state from datasource to resource, then back, when things are done
+func (s *serviceIntegrationDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var o dataSourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &o)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var res resourceModel
+	err := copier.Copy(&res, &o)
```
Serpentiel, line 118: does this approach have any performance drawbacks?

internal/plugin/service/serviceintegration/service_integration_data_source.go
Serpentiel: does this approach have any performance drawbacks?

byashimov: I don't think that we can notice it even.

byashimov:
CodeQL fails because of a known [bug](https://github.com/jeremylong/DependencyCheck/issues/5907).

ucgenerator/main.go
```diff
@@ -0,0 +1,598 @@
...
+		if p.isNestedBlock() {
+			blocks[key] = getSchemaAttributes(p, pkg)
+		} else {
+			// For scalars
+			var value *jen.Statement
+			switch p.Type {
+			case objectTypeObject:
+				// Schemaless map
+				//value = jen.Qual(importTypes, "StringType")
+				panic("schemaless objects are not supported")
```
ivan-savciuc, line 382: Can we add error handling here and in a few other places where we use panic?  It would be more predictable if we did not through panic in the middle of the business logic.

ucgenerator/main.go
```diff
@@ -0,0 +1,598 @@
+package main
```
ivan-savciuc, line 1: Should we perhaps add a build tag?

ivan-savciuc (changes requested):
Tested and it works really well, I do have some minor comments thought

ucgenerator/main.go
byashimov: > Can we add error handling here and in a few other places where we use panic? It would be more predictable if we did not through panic in the middle of the business logic.

Hey. I totally agree with this. But this doesn't run in runtime and it makes much easier to read the code. This error will finish the program as the panic does.

ucgenerator/main.go
byashimov: Turns out, go walks through imports. If module not imported, it's not included.

internal/plugin/service/serviceintegration/service_integration_data_source.go
Serpentiel: could you consider adding a `gobench` testing the following:

- a data source which doesn't performs any copying with this library
- a data source which does

both should be similar

wdyt about adding it?

byashimov:
Rebased on master, fixed conflicts.

Serpentiel (approved):
lgtm

ivan-savciuc (approved):
LGTM

PR #1347, opened 2023-09-07
fix(kafkatopic): cache invalidation

ivan-savciuc:

Fix cache invalidation for the cases when we try to remove the last element from the list


Serpentiel (approved):
lgtm

PR #1348, opened 2023-09-07
test(connectionpool): fix misuse of wrong user resource

Serpentiel:

fixes misuse of `aiven_opensearch_user` resource instead of `aiven_pg_user` in a `connectionpool` test, that is `pg`-related

## Why this way

the test was failing


byashimov:
It's not a bug, because PR doesn't fix provider.

byashimov (approved):
lgtm

Serpentiel:
bug label is also to use with tests

PR #1360, opened 2023-09-18
fix(kafkatopic): replication for startup-2 plan

ivan-savciuc:

Fixing 409 error:
```
 Error: 409: {"errors":[{"message":"Cluster only has 2 broker(s), cannot set replication factor to 3","status":409}],"message":"Cluster only has 2 broker(s), cannot set replication factor to 3"} - 
```


byashimov (approved):
lgtm

Serpentiel (approved):

Serpentiel (approved):

PR #1361, opened 2023-09-18
fix(kafkatopic): acc tests but reverting performance fix

ivan-savciuc:

Fixes acceptance but performance patch get reverted


ivan-savciuc:
tests: 
https://github.com/aiven/terraform-provider-aiven/actions/runs/6222613991

PR #1364, opened 2023-09-19
feat(pg): allow to modify user replication settings

### LINKED ISSUE #1362: aiven_pg_user resource should not replace user when enabling replication
## What happened?

Enabling replication using the `aiven_pg_user` resource forces replacement of the PostgreSQL user. When the user is replaced, it will loose all ownership and privileges over database objects such as tables and functions and the default `avnadmin` will be owner of previous DB objects. If enabling replication using the web console, it works as expected. The user will keep current ownership and privileges.

Enable replication without user being force replaced and loose ownership + privileges.

```
Terraform v1.5.7
on linux_amd64
+ provider registry.terraform.io/aiven/aiven v4.8.2
```

```
  # aiven_pg_user.this["test"] must be replaced
-/+ resource "aiven_pg_user" "this" {
      + access_cert          = (sensitive value)
      + access_key           = (sensitive value)
      ~ id                   = "***/mypostgres/test" -> (known after apply)
      ~ pg_allow_replication = false -> true # forces replacement
      ~ type                 = "normal" -> (known after apply)
        # (4 unchanged attributes hidden)
    }
```

Serpentiel:

allows to modify `pg_user` replication settings

resolves #1362 

## Why this way

this is long ago supported by the API

### DIFF
--- a/internal/sdkprovider/service/pg/pg_user.go
+++ b/internal/sdkprovider/service/pg/pg_user.go
@@ -28,16 +28,12 @@ var aivenPGUserSchema = map[string]*schema.Schema{
 		Sensitive:        true,
 		Computed:         true,
 		DiffSuppressFunc: schemautil.EmptyObjectDiffSuppressFunc,
-		Description:      "The password of the PG User ( not applicable for all services ).",
+		Description:      "The password of the PG User (not applicable for all services).",
 	},
 	"pg_allow_replication": {
 		Type:        schema.TypeBool,
 		Optional:    true,
-		ForceNew:    true,
-		Description: userconfig.Desc("Defines whether replication is allowed.").ForceNew().Build(),
-		Elem: &schema.Schema{
-			Type: schema.TypeBool,
-		},
+		Description: "Defines whether replication is allowed.",
 	},
 
 	// computed fields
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aiven/terraform-provider-aiven
 
-go 1.19
+go 1.21.1
```
Serpentiel, line 3: `terratest` requires `go` v1.21.1, and fails otherwise

internal/sdkprovider/service/pg/pg_user.go
```diff
@@ -128,6 +124,23 @@ func resourcePGUserUpdate(ctx context.Context, d *schema.ResourceData, m interfa
 		return diag.FromErr(err)
 	}
 
+	if d.HasChange("pg_allow_replication") {
```
ivan-savciuc, line 127: Can we somehow combine previous `client.ServiceUsers.Update(projectName, serviceName, username,
		aiven.ModifyServiceUserRequest{...` with this logic? It is strange to see two update calls to the same endpoint one after another. 

So the question is - can `Operation: set-access-control` also update passwords

ivan-savciuc (approved):
LGTM, however I have a question

internal/sdkprovider/service/pg/pg_user.go
Serpentiel: no, it's not possible

PR #1365, opened 2023-09-21
test(kafka): fix mirrormaker invalid offset syncs topic location

Serpentiel:

fixes mirrormaker invalid offset syncs topic location test

## Why this way

the error message was updated in backend, I've made it a bit more fail-proof as well in this PR


ivan-savciuc (approved):
LGTM

PR #1367, opened 2023-09-25
feat(org): add user group member support

ivan-savciuc:

Adding organization user group member support.

I cannot add acceptance tests for this feature because the `user_id` of the newly invited member will be known only after an invitation is accepted. And we cannot simulate this during the test. Moreover, if we invite already existing Aiven users to a different organization, `user_id` is still unavailable, and we have to accept an invitation.

This is revived version of this PR: https://github.com/aiven/terraform-provider-aiven/pull/1282

@ngilles-aiven wrote this feedback: 

> Something didn't sit quite right after going over this, but I couldn't quite put my finger on it, but it looks like it's because user management within the Aiven platform being not a great fit Terraform as fully declarative system.

> I also am not fully sure why it seems to be more complicated in the case of organization users (and groups) than it was with teams, and it seems it comes down to managing things with user ids instead of email? (as invites are done with emails).

> Maybe this warrants more discussion on the overall approach or at least the need to extra api/functionality?

The difference between `aiven_account_team_member` and new `aiven_organization_user_group_member` is following: 
- The organization version requires a user ID for creation, and the user ID is only known after the user has accepted an invitation; that makes our life much more problematic when testing it since we cannot close the loop using only Terraform.
- The account version accepts the user's email address and, moreover, accepts it even if the user hasn't accepted an email invitation. 

Ideally we want to keep using email address for the instead of user ID if possible, cc @ngilles-aiven @rsalevsky

### DIFF
--- a/internal/sdkprovider/service/organization/organization_user_group_member.go
+++ b/internal/sdkprovider/service/organization/organization_user_group_member.go
@@ -1,0 +1,126 @@
+package organization
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/aiven/aiven-go-client"
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
+)
+
+var aivenOrganizationUserGroupMemberSchema = map[string]*schema.Schema{
+	"organization_id": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: userconfig.Desc("The unique organization ID").ForceNew().Build(),
+	},
+	"group_id": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: userconfig.Desc("The unique organization user group ID").ForceNew().Build(),
+	},
+	"user_id": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: userconfig.Desc("The organization user group user ID").ForceNew().Build(),
+	},
+	"state": {
+		Type:        schema.TypeString,
+		Computed:    true,
+		Description: "State of the organization user group member",
+	},
+}
+
+func ResourceOrganizationUserGroupMember() *schema.Resource {
+	return &schema.Resource{
+		Description:   "The Organization User Group Member resource allows the creation and management of an Aiven Organization Group Members.",
+		CreateContext: resourceOrganizationUserGroupMemberCreate,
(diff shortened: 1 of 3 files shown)


Serpentiel (changes requested):
looks like current state of things allows us to implement this via Terraform Plugin Framework, could you please redo it with its usage?

Serpentiel:
> I cannot add acceptance tests for this feature because the user_id of the newly invited member will be known only after an invitation is accepted. And we cannot simulate this during the test.

does calling [this](https://api.aiven.io/doc/#tag/Organization/operation/UserGroupMemberList) API endpoint not returns a list of users even with those that hadn't accepted the invitation yet?

internal/sdkprovider/service/organization/organization_user_group_member.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/aiven/aiven-go-client"
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
+	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
+)
+
+var aivenOrganizationUserGroupMemberSchema = map[string]*schema.Schema{
+	"organization_id": {
```
rsalevsky, line 15: Isn't the organization_id given by the group?

Serpentiel: it is, why?

ivan-savciuc:
The decision was made to skip this feature implementation

mwoods-familiaris:
> The decision was made to skip this feature implementation

Does this mean there will not be a way to administer user/group membership via Terraform? We have created users and groups via Terraform, and were hoping that we might be able to handle group membership that way as well. Is this not going to be possible? If so...is the recommendation to stick with the older "teams" functionality for RBAC instead?

Serpentiel:
@ivan-savciuc could you please give your comment?

ivan-savciuc:
@mwoods-familiaris, unfortunately, it will not be possible to administer user/group membership via Terraform. The reason is API implementation. An organization user has no ID until an invitation that was sent via email hasn't been accepted. Therefore, we cannot organize users into groups until all users we manage have accepted invitations, which breaks Terraform flow to multiple stages.  

So, unlike account teams and projects, similar functionality will not be available for organizations. If this is important for your organization, please contact our customer support with such a request. 

We need to change an API to unblock this functionality for Terraform and other IaC tools, and an internal decision was made to wait and get feedback if this is something our customers want. 

Lastly, the account teams, users, projects, etc., will be deprecated and eventually deleted from Terraform in the upcoming major releases. All the customers should migrate to organizations that do not support the same level of functionality in Terraform for the moment of writing.

PR #1369, opened 2023-09-26
feat(kafka): allow protobuf schema

### LINKED ISSUE #1366: Add support for protobuf schema type for aiven_kafka_schema
## What is currently missing?

Protobuf schema support for Kafka schemas

aiven_kafka_schema should support protobuf schema types since it's supported by the Aiven platform and the REST API
https://api.aiven.io/doc/#tag/Service:_Kafka/operation/ServiceSchemaRegistrySubjectVersionPost

Serpentiel:

allows usage of `protobuf` schema in Kafka resources

resolves #1366

## Why this way

we would want to have this feature in place

### DIFF
--- a/internal/sdkprovider/service/kafkaschema/kafka_schema.go
+++ b/internal/sdkprovider/service/kafkaschema/kafka_schema.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 
 	"github.com/aiven/aiven-go-client/v2"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -15,6 +16,9 @@ import (
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
 )
+
+// newlineRegExp is a regular expression that matches a newline.
+var newlineRegExp = regexp.MustCompile(`\r?\n`)
 
 var aivenKafkaSchemaSchema = map[string]*schema.Schema{
 	"project":      schemautil.CommonSchemaProjectReference,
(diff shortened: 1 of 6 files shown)


internal/sdkprovider/service/kafkaschema/kafka_schema.go
```diff
@@ -16,6 +17,29 @@ import (
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil/userconfig"
 )
 
+// protobufHeaderRegExp is a regular expression that matches a valid Protobuf header.
+var protobufHeaderRegExp = regexp.MustCompile("syntax(.*|)=(.*|)proto3(.*|);")
```
ivan-savciuc, line 21: I think we will be better off with something like https://github.com/bufbuild/protovalidate-go or https://pkg.go.dev/github.com/deelawn/protoc-gen-validation. Since the protobuf format may evolve in the future, Aiven, in the long run, will also support various versions simultaneously. We want to delegate validation to dedicated projects that focus solely on that.

internal/sdkprovider/service/kafkaschema/kafka_schema.go
```diff
@@ -79,13 +105,40 @@ func diffSuppressJSONObject(_, old, new string, _ *schema.ResourceData) bool {
...
+
 // normalizeJSONString returns normalized JSON string
 func normalizeJSONString(v interface{}) string {
 	jsonString, _ := structure.NormalizeJsonString(v)
 
 	return jsonString
 }
 
+// normalizeProtobufString returns normalized Protobuf string.
+func normalizeProtobufString(i any) string {
```
ivan-savciuc, line 125: Same point here, I would love to do normalization using a go protobuf library for that.

ivan-savciuc (dismissed):
LGTM, the only proposition of my would be to use a go-protobuf library for validation and normalization

internal/sdkprovider/service/kafkaschema/kafka_schema.go
Serpentiel: currently it was implemented the same way as it is on front end, of course in the long run it will be viable to replace it

internal/sdkprovider/service/kafkaschema/kafka_schema.go
Serpentiel: it requires the end users to install `protoc`, which wouldn't be too good

internal/sdkprovider/service/kafkaschema/kafka_schema_test.go
```diff
@@ -269,6 +295,62 @@ data "aiven_kafka_schema" "schema" {
...
+EOT
+}
+
+data "aiven_kafka_schema" "schema" {
+  project      = aiven_kafka_schema.foo.project
+  service_name = aiven_kafka_schema.foo.service_name
+  subject_name = aiven_kafka_schema.foo.subject_name
+
+  depends_on = [aiven_kafka_schema.foo]
+}`, os.Getenv("AIVEN_PROJECT_NAME"), name, name)
```
byashimov, line 351: Is it possible to extend existing test with json schema (to not create an extra service)?

Serpentiel: should be, I'll update the PR

internal/sdkprovider/service/kafkaschema/kafka_schema.go
byashimov: I don't see any reason having such a validation. It basically checks that string contains `syntax = proto3;` substring. Which is not a valid syntax btw.
And I might be missing something `(.*|)` means "anything-or-nothing or nothing", right? The `plan` command won't show any "real" validation issues.

byashimov (changes requested):

Serpentiel:
> I don't see any reason having such a validation. It basically checks that string contains syntax = proto3; substring. Which is not a valid syntax btw.

so what do you suggest changing it with then? I checked, it works just fine as long as the string looks like `syntax...=...proto3...;`

we need some kind of validation in this place to be able to combine it together with the existing JSON validation mechanic

Serpentiel:
@byashimov I've updated the PR with the fix

byashimov (approved):
lgtm

PR #1371, opened 2023-09-28
fix(project_user): invite error handling

ivan-savciuc:

V3 fix for project user when user invitation has yet to be accepted, but we try to refresh data and send another invitation, resulting in 409 error from Aiven.

### DIFF
--- a/internal/service/kafka/kafka_topic_cache_test.go
+++ b/internal/service/kafka/kafka_topic_cache_test.go
@@ -35,6 +35,8 @@ func TestGetTopicCache(t *testing.T) {
 			&kafkaTopicCache{
 				internal: make(map[string]map[string]aiven.KafkaTopic),
 				inQueue:  make(map[string][]string),
+				missing:  make(map[string][]string),
+				v1list:   make(map[string][]string),
 			},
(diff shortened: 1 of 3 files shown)


internal/service/kafka/kafka_topic_cache_test.go
```diff
@@ -35,6 +35,8 @@ func TestGetTopicCache(t *testing.T) {
 			&kafkaTopicCache{
 				internal: make(map[string]map[string]aiven.KafkaTopic),
 				inQueue:  make(map[string][]string),
+				missing:  make(map[string][]string),
+				v1list:   make(map[string][]string),
 			},
```
byashimov, line 40: I think, this might be added by an accident.

ivan-savciuc: I just discovered that unit tests are failing in v3 and this is a fix for that

ivan-savciuc: I can add it in a separate commit or PR if you prefer it this way.

byashimov (approved):
lgtm

PR #1382, opened 2023-10-05
feat(plugin): use set instead of list

byashimov:

Uses set instead of list type.


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/6419074431/job/17428162032)

Serpentiel (approved):
lgtm

PR #1390, opened 2023-10-13
chore: deprecating account and project user

ivan-savciuc:

Deprecating account teams and team member resources alongside project users, as discussed internally. 

:warning: The deprecation message does not contain a particular resource as the recommended replacement because we do not expect a 1-to-1 transition from account to organization, and not all organization-related functionality will be transferred to the Terraform provider due to an internal investigation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ nav_order: 1
 
 # Changelog
 
+- Deprecating `project_user`, `account_team` and `account_team_member resources`
 ## [5.0.0] - YYYY-MM-DD
 
 - Migrate `aiven_service_integration` to the Plugin Framework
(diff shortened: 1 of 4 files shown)


Serpentiel (dismissed):
the message mentions "Organization User Groups" but doesn't provide a clear reference to it—this can confuse users; we should offer a more explicit path or documentation for users to follow when transitioning

to ensure a smooth transition for users, might we consider delaying this PR until we have a clear replacement or at least a documented alternative?

ivan-savciuc:
> the message mentions "Organization User Groups" but doesn't provide a clear reference to it—this can confuse users; we should offer a more explicit path or documentation for users to follow when transitioning
> 
> to ensure a smooth transition for users, might we consider delaying this PR until we have a clear replacement or at least a documented alternative?

@Serpentiel I explicitly mentioned this in the PR description 
image

Serpentiel (approved):
lgtm

PR #1391, opened 2023-10-13
fix(go.mod): fix broken dependency

ivan-savciuc:

fix go.mod automerge problem


byashimov (approved):
lgtm

PR #1393, opened 2023-10-13
fix(topic): performance for read and creation

ivan-savciuc:

Fix Kafka Topic create and read performance

### DIFF
--- a/internal/sdkprovider/service/kafkatopic/kafka_topic.go
+++ b/internal/sdkprovider/service/kafkatopic/kafka_topic.go
@@ -269,16 +269,9 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 	// aiven.KafkaTopics.Create() function may return 501 on create
 	// Second call might say that topic already exists
 	// So to be sure, better check it before create
-	_, err := getTopic(ctx, m, d.Timeout(schema.TimeoutRead), project, serviceName, topicName)
-
-	// No error means topic exists
-	if err == nil {
+	client := m.(*aiven.Client)
+	if isTopicExists(ctx, client, project, serviceName, topicName) {
 		return diag.Errorf("Topic conflict, already exists: %s", topicName)
-	}
-
-	// If this is not "does not exist", then something happened
-	if !aiven.IsNotFound(err) {
-		return diag.FromErr(err)
 	}
 
 	createRequest := aiven.CreateKafkaTopicRequest{
@@ -289,7 +282,7 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 		Tags:        getTags(d),
 	}
 
-	err = m.(*aiven.Client).KafkaTopics.Create(
+	err := client.KafkaTopics.Create(
 		ctx,
 		project,
 		serviceName,
@@ -301,8 +294,7 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 
 	d.SetId(schemautil.BuildResourceID(project, serviceName, topicName))
 
-	// Invalidates cache for the topic
-	DeleteTopicFromCache(project, serviceName, topicName)
+	getTopicCache().AddToQueue(project, serviceName, topicName)
 
 	// We do not call a Kafka Topic read here to speed up the performance.
 	// However, in the case of Kafka Topic resource getting a computed field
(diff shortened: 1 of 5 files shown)


byashimov:
@ivan-savciuc hey is this for v4 or v3, or both?

byashimov (dismissed):
lgtm

internal/sdkprovider/service/kafkatopic/kafka_topic.go
```diff
@@ -269,18 +269,11 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 	// aiven.KafkaTopics.Create() function may return 501 on create
 	// Second call might say that topic already exists
 	// So to be sure, better check it before create
-	_, err := getTopic(ctx, m, d.Timeout(schema.TimeoutRead), project, serviceName, topicName)
-
-	// No error means topic exists
-	if err == nil {
+	client := m.(*aiven.Client)
+	if isTopicExists(ctx, client, project, serviceName, topicName) {
```
Serpentiel, line 273: consider adding a comment above this block to briefly explain its purpose for clarity

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
```diff
@@ -201,3 +201,11 @@ func DeleteTopicFromCache(projectName, serviceName, topicName string) {
 	}
 	t.Unlock()
 }
+
+// GetFullQueue retrieves a topics queue
+func (t *kafkaTopicCache) GetFullQueue(projectName, serviceName string) []string {
```
Serpentiel, line 206: ensure the method name accurately represents its purpose; If the function retrieves a specific type of queue or has a unique behavior, be more explicit in the name

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
```diff
@@ -201,3 +201,11 @@ func DeleteTopicFromCache(projectName, serviceName, topicName string) {
...
+	t.RLock()
+	defer t.RUnlock()
+
+	return t.inQueue[projectName+serviceName]
```
Serpentiel, line 210: consider handling a scenario where the key might not exist in the `inQueue` map; directly accessing a non-existent key will return a nil slice, which might be unexpected for the function caller

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"log"
+	"sync"
+
+	"github.com/aiven/aiven-go-client/v2"
+	"golang.org/x/exp/slices"
+)
+
+var onceForService sync.Map
```
Serpentiel, line 12: consider renaming the variable to be more descriptive, like `onceCheckTopicForService` or something similar

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+
+// isTopicExists checks if topic exists
+func isTopicExists(ctx context.Context, client *aiven.Client, project, serviceName, topic string) bool {
+	c := getTopicCache()
+
+	var err error
+	once, _ := onceForService.LoadOrStore(project+"/"+serviceName, new(sync.Once))
+	once.(*sync.Once).Do(func() {
```
Serpentiel, line 20: add comments for better clarity, especially on the significance of using `sync.Once`

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+		var list []*aiven.KafkaListTopic
+		list, err = client.KafkaTopics.List(ctx, project, serviceName)
+		if err != nil {
+			return
+		}
+
+		c.SetV1List(project, serviceName, list)
+	})
+
+	if err != nil {
```
Serpentiel, line 30: this approach can be confusing: if possible, consider handling the error within the `once.Do` block, or refactoring the approach to make the flow clearer

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+		log.Printf("[ERROR] cannot check kafka topic existence: %s", err)
+		return false
+	}
+
+	if slices.Contains(c.GetV1List(project, serviceName), topic) || slices.Contains(c.GetFullQueue(project, serviceName), topic) {
```
Serpentiel, line 35: for better performance, if the topic is found in the first `c.GetV1List` check, there's no need to check in the second `c.GetFullQueue` call: you can return `true` immediately

Serpentiel (changes requested):

ivan-savciuc:
Converting this patch for v4 instead of v5 (current master), will open another RP

PR #1395, opened 2023-10-17
test(opensearch): fix weak password error

Serpentiel:

fixes a weak password error in our [acceptance tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/6540687245/job/17761032810)

## Why this way

to make it work :)


ivan-savciuc (approved):
LGTM

PR #1405, opened 2023-10-22
test(account): fix failing team tests

Serpentiel:

fixes failing team test

## Why this way

to make them work

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -69,6 +69,8 @@ jobs:
         env:
           AIVEN_TOKEN: ${{ secrets.AIVEN_TOKEN }}
           AIVEN_PROJECT_NAME: ${{ secrets.AIVEN_PROJECT_NAME }}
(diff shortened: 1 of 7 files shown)


ivan-savciuc (changes requested):
Replacing the account as an organization under the hood is not a good idea. We should request a change of permissions for our CI projects to test the same API and underlying logic that our customers do, those who are stuck with the accounts, and we have most such customers right now.

Serpentiel:
this never replaced anything with an organization, its simply an env variable name, because the entity was renamed from `account` to `organization`, but as you can see, under the hood it uses `aiven_account` data source

Serpentiel:
acceptance tests (ignore everything except `account`): https://github.com/aiven/terraform-provider-aiven/actions/runs/6609605992/job/17950010317

Serpentiel:
@mhoffm-aiven thanks for helping to set things up from the CI side of things 🥂

ivan-savciuc (approved):
LGTM

PR #1406, opened 2023-10-23
test(flink): fix application version flakiness

Serpentiel:

fixes Flink application version flakiness

## Why this way

to make it work :)

### DIFF
--- a/internal/sdkprovider/service/flink/flink_application_version.go
+++ b/internal/sdkprovider/service/flink/flink_application_version.go
@@ -2,9 +2,12 @@ package flink
 
 import (
 	"context"
+	"regexp"
+	"time"
 
 	"github.com/aiven/aiven-go-client/v2"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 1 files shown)


internal/sdkprovider/service/flink/flink_application_version.go
```diff
@@ -186,13 +189,40 @@ func resourceFlinkApplicationVersionCreate(ctx context.Context, d *schema.Resour
...
+				Statement: d.Get("statement").(string),
+				Sources:   sources,
+				Sinks:     sinks,
+			},
+		)
+		if err != nil {
+			return &retry.RetryError{
+				Err: err,
+				Retryable: regexp.MustCompile(
+					"Integration not found: " +
```
ivan-savciuc, line 212: Perhaps we can simplify this like `regexp.MustCompile("Integration not found:")`, whatever goes after service integration is unimportant. 

For example, if someone alters an error message on the back-end, like adding a dot in the end or something else, that would impact us here.

ivan-savciuc (commented):
LGTM, a small suggestion for catching the right error

internal/sdkprovider/service/flink/flink_application_version.go
Serpentiel: at least we'll know then if something changed around this API :)

ivan-savciuc (approved):
LGTM

PR #1407, opened 2023-10-23
test(project): fix missing organization link

Serpentiel:

fix project test missing organization link in second step

## Why this way

to make the test work


ivan-savciuc (approved):
LGTM

PR #1408, opened 2023-10-23
fix(mysql): use correct read context fn

### LINKED ISSUE #1401: Handle missing aiven_mysql_user resource gracefully
## What happened?

When user is removed manually from the console running `terraform plan` at that point returns `Error: user xx/xx/xx not found`.

It should handle the situation gracefully and create the resource again. It works this way with `aiven_pg_user` as it calls the [ResourceReadHandleNotFound](https://github.com/aiven/terraform-provider-aiven/blob/main/internal/schemautil/service.go#L832) function 

To reproduce create `aiven_mysql_user` resource and delete the user manually from the console. Then execute `terraform plan` again and error should pop up.

Serpentiel:

uses correct read context fn in mysql user resource

resolves #1401

## Why this way

to fix it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,11 @@ nav_order: 1
 
 # Changelog
 
-- Deprecating `project_user`, `account_team` and `account_team_member resources`
 ## [5.0.0] - YYYY-MM-DD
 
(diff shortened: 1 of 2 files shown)


ivan-savciuc (approved):
LGTM

PR #1410, opened 2023-10-24
Kafka Topic performance improvements for v4

ivan-savciuc:

Fix Kafka Topic create and read performance

### DIFF
--- a/internal/sdkprovider/service/kafkatopic/kafka_topic.go
+++ b/internal/sdkprovider/service/kafkatopic/kafka_topic.go
@@ -265,20 +265,13 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 	topicName := d.Get("topic_name").(string)
 	partitions := d.Get("partitions").(int)
 	replication := d.Get("replication").(int)
+	client := m.(*aiven.Client)
 
 	// aiven.KafkaTopics.Create() function may return 501 on create
 	// Second call might say that topic already exists
 	// So to be sure, better check it before create
-	_, err := getTopic(ctx, m, d.Timeout(schema.TimeoutRead), project, serviceName, topicName)
-
-	// No error means topic exists
-	if err == nil {
+	if isTopicExists(ctx, client, project, serviceName, topicName) {
 		return diag.Errorf("Topic conflict, already exists: %s", topicName)
-	}
-
-	// If this is not "does not exist", then something happened
-	if !aiven.IsNotFound(err) {
-		return diag.FromErr(err)
 	}
 
 	createRequest := aiven.CreateKafkaTopicRequest{
@@ -289,7 +282,7 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 		Tags:        getTags(d),
 	}
 
-	err = m.(*aiven.Client).KafkaTopics.Create(
+	err := client.KafkaTopics.Create(
 		ctx,
 		project,
 		serviceName,
@@ -300,9 +293,7 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 	}
 
 	d.SetId(schemautil.BuildResourceID(project, serviceName, topicName))
-
-	// Invalidates cache for the topic
-	DeleteTopicFromCache(project, serviceName, topicName)
+	getTopicCache().AddToQueue(project, serviceName, topicName)
 
 	// We do not call a Kafka Topic read here to speed up the performance.
 	// However, in the case of Kafka Topic resource getting a computed field
--- a/internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
+++ b/internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
@@ -142,8 +142,8 @@ func (t *kafkaTopicCache) GetMissing(projectName, serviceName string) []string {
 	return t.missing[projectName+serviceName]
 }
 
-// GetQueue retrieves a topics queue, retrieves up to 100 first elements
-func (t *kafkaTopicCache) GetQueue(projectName, serviceName string) []string {
+// GetQueueTop100 retrieves a topics queue, retrieves up to 100 first elements
+func (t *kafkaTopicCache) GetQueueTop100(projectName, serviceName string) []string {
 	t.RLock()
 	defer t.RUnlock()
 
@@ -202,4 +202,12 @@ func DeleteTopicFromCache(projectName, serviceName, topicName string) {
 		delete(t.internal[key], topicName)
 	}
 	t.Unlock()
+}
+
+// GetFullQueue retrieves a topics queue
+func (t *kafkaTopicCache) GetFullQueue(projectName, serviceName string) []string {
+	t.RLock()
+	defer t.RUnlock()
+
+	return t.inQueue[projectName+serviceName]
 }
--- a/internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
+++ b/internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
@@ -1,0 +1,42 @@
+package kafkatopic
+
+import (
+	"context"
+	"log"
+	"sync"
+
+	"github.com/aiven/aiven-go-client/v2"
+	"golang.org/x/exp/slices"
+)
+
+var onceCheckTopicForService sync.Map
+
+// isTopicExists checks if topic exists
+func isTopicExists(ctx context.Context, client *aiven.Client, project, serviceName, topic string) bool {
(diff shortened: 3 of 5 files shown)


internal/sdkprovider/service/kafkatopic/kafka_topic.go
```diff
@@ -265,22 +265,15 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
...
+	client := m.(*aiven.Client)
 
 	// aiven.KafkaTopics.Create() function may return 501 on create
 	// Second call might say that topic already exists
 	// So to be sure, better check it before create
-	_, err := getTopic(ctx, m, d.Timeout(schema.TimeoutRead), project, serviceName, topicName)
-
-	// No error means topic exists
-	if err == nil {
+	if isTopicExists(ctx, client, project, serviceName, topicName) {
```
Serpentiel, line 273: nit: consider adding a comment above this block to briefly explain its purpose for clarity

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
```diff
@@ -203,3 +203,11 @@ func DeleteTopicFromCache(projectName, serviceName, topicName string) {
 	}
 	t.Unlock()
 }
+
+// GetFullQueue retrieves a topics queue
+func (t *kafkaTopicCache) GetQueue(projectName, serviceName string) []string {
```
Serpentiel, line 208: ensure the method name accurately represents its purpose; If the function retrieves a specific type of queue or has a unique behavior, be more explicit in the name

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
```diff
@@ -203,3 +203,11 @@ func DeleteTopicFromCache(projectName, serviceName, topicName string) {
...
+	t.RLock()
+	defer t.RUnlock()
+
+	return t.inQueue[projectName+serviceName]
```
Serpentiel, line 212: consider handling a scenario where the key might not exist in the `inQueue` map; directly accessing a non-existent key will return a nil slice, which might be unexpected for the function caller

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"log"
+	"sync"
+
+	"github.com/aiven/aiven-go-client/v2"
+	"golang.org/x/exp/slices"
+)
+
+var onceForService sync.Map
```
Serpentiel, line 12: consider renaming the variable to be more descriptive, like `onceCheckTopicForService` or something similar

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+
+// isTopicExists checks if topic exists
+func isTopicExists(ctx context.Context, client *aiven.Client, project, serviceName, topic string) bool {
+	c := getTopicCache()
+
+	var err error
+	once, _ := onceForService.LoadOrStore(project+"/"+serviceName, new(sync.Once))
+	once.(*sync.Once).Do(func() {
```
Serpentiel, line 20: add comments for better clarity, especially on the significance of using `sync.Once`

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+		var list []*aiven.KafkaListTopic
+		list, err = client.KafkaTopics.List(ctx, project, serviceName)
+		if err != nil {
+			return
+		}
+
+		c.SetV1List(project, serviceName, list)
+	})
+
+	if err != nil {
```
Serpentiel, line 30: this approach can be confusing: if possible, consider handling the error within the `once.Do` block, or refactoring the approach to make the flow clearer

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
```diff
@@ -0,0 +1,40 @@
...
+		log.Printf("[ERROR] cannot check kafka topic existence: %s", err)
+		return false
+	}
+
+	if slices.Contains(c.GetV1List(project, serviceName), topic) || slices.Contains(c.GetQueue(project, serviceName), topic) {
```
Serpentiel, line 35: for better performance, if the topic is found in the first `c.GetV1List` check, there's no need to check in the second `c.GetQueue` call: you can return true immediately

Serpentiel (dismissed):

internal/sdkprovider/service/kafkatopic/kafka_topic.go
ivan-savciuc: We have a big comment here already 
```
// aiven.KafkaTopics.Create() function may return 501 on create
	// Second call might say that topic already exists
	// So to be sure, better check it before create
```

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
ivan-savciuc: We have only one queue and two methods:
1. retrieves the full queue - this one `GetQueue`
2. retrieves only top 100 elements `GetQueueTop100` 

And I have comments for each of the functions describing all of that

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
ivan-savciuc: ok will do

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
ivan-savciuc: No need `nil` slice is perfectly expect here, looks at the `GetQueueTop100`

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
ivan-savciuc: Please provide an example we cannot do that here since we need to handle a possible error from the `once.Do` in the main scope.

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
ivan-savciuc: It does not make any sense to me, `slices.Contains` takes nano seconds and since we have `OR ||` operation, interpreter first check the `c.GetV1List` and if it is `true` second checks will be spiked by interpreter.

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
ivan-savciuc: ok

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
Serpentiel: please see this code example: https://github.com/aiven/terraform-provider-aiven/pull/1410#discussion_r1369798811

this approach allows to reduce to complexity by moving the second `slices.Contains` call to the `return` statement

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
Serpentiel: nit: make it more obvious by naming it `GetFullQueue`

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
ivan-savciuc: I do not think that it makes core more understandable and we still treat error outside the do.Once scope.

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
ivan-savciuc: Ok, now I see your point will change that.

internal/sdkprovider/service/kafkatopic/kafka_topic_exists.go
Serpentiel: it makes the code less complex and easier to understand in comparison to the previous version

internal/sdkprovider/service/kafkatopic/kafka_topic_cache.go
ivan-savciuc: ok

Serpentiel (approved):
lgtm

Serpentiel:
@ivan-savciuc don't forget to port this to v5

PR #1413, opened 2023-10-25
ci(acctests): run for v4 branch

Serpentiel:

- ci(checkout): set fetch-depth zero globally
- ci(acctests): run for v4 branch

## Why this way

- we want to have CI in place for v4 and v3 branches
- we want the CI to work properly


ivan-savciuc (approved):
LGTM

PR #1416, opened 2023-10-25
chore(changelog): drop v4 entry

Serpentiel:

drops v4 changelog entry: https://github.com/aiven/terraform-provider-aiven/blob/v4/CHANGELOG.md

## Why this way

it will already be released in v4, so we don't need to duplicate the changelog entry here


ivan-savciuc (approved):
LGTM

PR #1420, opened 2023-10-30
chore(userconfig): improve naming and docs

Serpentiel:

improves naming and docs in `userconfig` package

## Why this way

to make it clearer for understanding


ivan-savciuc (approved):
I this this is a good change, improves readability of the code :+1:

PR #1425, opened 2023-11-02
fix(schemautil): unmarshalling empty userconfig crash

Serpentiel:

- fixes unmarshalling empty userconfig crash
- improves debugging experience

## Why this way

- to have userconfig unmarshalling working
- to be able to debug in a more efficient way


byashimov (approved):
lgtm

PR #1433, opened 2023-11-07
docs: document other env vars for make test-acc

sigmaris:

Improves documentation of the environment variables that are used or influence the running of the acceptance tests.

## Why this way

Fully running the acceptance tests uses more than just a project, there are tests which interact with accounts and orgs, and it took some questioning on Slack and searching the source code to figure out how to properly run these tests locally and against a dev environment, so we should probably document how to get them to run here to save future engineers time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,11 +31,15 @@ Run the tests with the command below:
 make test
 ```
 
+Acceptance tests interact with the real Aiven API and require an API token and existing project and organization on Aiven to create resources in. A non-default API URL can be used by setting the `AIVEN_WEB_URL` environment variable to e.g. `https://your.custom.api:443`.
+
 Run the acceptance tests with the commands below:
 
 ```bash
 export AIVEN_TOKEN="your-token"
 export AIVEN_PROJECT_NAME="your-project-name"
+export AIVEN_ORGANIZATION_NAME="your-existing-org-name"
+export AIVEN_ACCOUNT_NAME="$AIVEN_ORGANIZATION_NAME"
 
 # run all acceptance tests
 make test-acc


CONTRIBUTING.md
```diff
@@ -31,11 +31,15 @@ Run the tests with the command below:
...
 ```
 
+Acceptance tests interact with the real Aiven API and require an API token and existing project, account and organization on Aiven to create resources in. A non-default API URL can be used by setting the `AIVEN_WEB_URL` environment variable to e.g. `https://your.custom.api:443`.
+
 Run the acceptance tests with the commands below:
 
 ```bash
 export AIVEN_TOKEN="your-token"
 export AIVEN_PROJECT_NAME="your-project-name"
+export AIVEN_ACCOUNT_NAME="your-existing-account-name"
```
Serpentiel, line 41: `AIVEN_ORGANIZATION_NAME` and `AIVEN_ACCOUNT_NAME` should match each other, `AIVEN_ACCOUNT_NAME` should be defined as:

```bash
export AIVEN_ACCOUNT_NAME=$AIVEN_ORGANIZATION_NAME
```

see https://github.com/aiven/terraform-provider-aiven/blob/main/.github/workflows/acceptance-tests.yml#L74-L75

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ nav_order: 1
 
 - Migrate `aiven_service_integration` to the Plugin Framework
 - Deprecating `project_user`, `account_team` and `account_team_member` resources
+- Improve contributing documentation on running acceptance tests.
```
Serpentiel, line 13: no changelog for docs

Serpentiel (changes requested):

CHANGELOG.md
sigmaris: I skipped this in the first version of the PR, but there's a "Required" check on PRs named "Changelog Enforcer" that fails if there's no changelog entry, even though only the docs were changed, so it looks like "changelog even for docs" is this repo's policy.

sigmaris:
The semgrep action also seems broken, though unrelated to this change.

Serpentiel:
> The semgrep action also seems broken, though unrelated to this change.

I'm currently working on it :)

CHANGELOG.md
Serpentiel: there's a `no changelog` label, I've added it for your PR now, and this check is now skipped

Serpentiel (approved):
lgtm

PR #1435, opened 2023-11-08
ci(lint): fix semgrep

Serpentiel:

fixes `semgrep` in the `lint` workflow

## Why this way

it was broken ☹️


byashimov (approved):
lgtm

PR #1436, opened 2023-11-08
ci(acctest): disable v4 workflow

Serpentiel:

disables v4 `acctest` workflow

## Why this way

it's no longer needed


byashimov (approved):
lgtm

PR #1437, opened 2023-11-08
test(examples): fix semgrep failing on math/rand import

Serpentiel:

fixes `semgrep` failing on `math/rand` import

## Why this way

we don't need no failures :)


byashimov (approved):
lgtm

PR #1442, opened 2023-11-13
fix(apiconvert): never skip basic auth username/password

### LINKED ISSUE #1411: Unable to update external schema registry auth credentials
## What happened?

An error occurs applying the following change:
```diff
 resource "aiven_service_integration_endpoint" "external_schema_registry" {
   project = "example-project"  

   endpoint_name = "schema-registry-proxy"
   endpoint_type   = "external_schema_registry"

   external_schema_registry_user_config {
     url = "https://schema-registry.example.com:8081"

     authentication      = "basic"
     basic_auth_username = "username"
-    basic_auth_password = "password"
+    basic_auth_password = "new-password"
   }
 }
```

The error is:
`Error: 400: {"errors":[{"message":"Invalid 'user_config': Invalid input: \"basic_auth_username\" is required with authentication=\"basic\"","status":400}],"message":"Invalid 'user_config': Invalid input: \"basic_auth_username\" is required with authentication=\"basic\""}`

A similar error is shown when updating only the `basic_auth_username` field. Updating both `basic_auth_username` and `basic_auth_password` works correctly.

The change should apply successfully when only changing one of the  `basic_auth_username` or `basic_auth_password` fields.

This seems to be a similar issue to #1142, but slightly different as these fields are only required conditionally.

Serpentiel:

prevents skipping basic auth username/password in certain service integrations

resolves #1411

## Why this way

API always wants us to send these details, regardless of if they're changed or not

### DIFF
--- a/internal/schemautil/userconfig/apiconvert/toapi.go
+++ b/internal/schemautil/userconfig/apiconvert/toapi.go
@@ -236,7 +236,7 @@ func itemToAPI(
 		}
 	}
 
-	if omitValue && isRequired {
+	if omitValue && isRequired || key == "basic_auth_username" || key == "basic_auth_password" {
 		omitValue = false
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ nav_order: 1
 
 - Deprecating `project_user`, `account_team` and `account_team_member` resources
 - Fix unmarshalling empty userconfig crash
+- Never skip basic auth username/password in service integrations user config options when sending them to the API
 
 ## [4.9.3] - 2023-10-27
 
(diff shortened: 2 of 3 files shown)


internal/schemautil/userconfig/apiconvert/toapi.go
```diff
@@ -236,7 +236,7 @@ func itemToAPI(
 		}
 	}
 
-	if omitValue && isRequired {
+	if omitValue && isRequired || key == "basic_auth_username" || key == "basic_auth_password" {
```
Serpentiel, line 239: @byashimov as I anticipated, it's a one line fix, at least in the current version of the generator

Serpentiel:
acceptance tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/6852128028

byashimov (approved):
lgtm

PR #1447, opened 2023-11-15
feat(mirrormaker): add emit_backward_heartbeats_enabled field support

Serpentiel:

adds `emit_backward_heartbeats_enabled` field support

## Why this way

to have it in place


ivan-savciuc (approved):
LGTM

PR #1451, opened 2023-11-16
feat: make sure emails are lowercase and valid

### LINKED ISSUE #1450: User invite with capitals in email would cause an infinite loop
## What happened?
when user invite contains capitals in email it goes into an infinite loop

```
variable "users" {
  type        = map(string)
  description = "Map of users"
  default     = {}
}
```
```
users = {
    "A@gmail.com" = "read_only"
    "B@gmail.com" = "read_only"
}
```

```iven_project_user.users["A@gmail.com"]: Creating...
aiven_project_user.users["B@gmail.com"]: Creating...
aiven_project_user.users["B@gmail.com"]: Still creating... [10s elapsed]
aiven_project_user.users["A@gmail.com"]: Still creating... [10s elapsed]
aiven_project_user.users["A@gmail.com"]: Still creating... [20s elapsed]
aiven_project_user.users["B@gmail.com"]: Still creating... [20s elapsed]
aiven_project_user.users["B@gmail.com"]: Still creating... [30s elapsed]
aiven_project_user.users["A@gmail.com"]: Still creating... [30s elapsed]
aiven_project_user.users["A@gmail.com"]: Still creating... [40s elapsed]
aiven_project_user.users["B@gmail.com"]: Still creating... [40s elapsed]
aiven_project_user.users["A@gmail.com"]: Still creating... [50s elapsed]
aiven_project_user.users["B@gmail.com"]: Still creating... [50s elapsed]
aiven_project_user.users["B@gmail.com"]: Still creating... [1m0s elapsed]
aiven_project_user.users["A@gmail.com"]: Still creating... [1m0s elapsed]
```
Apparently it already created the user invites in lower case, but it kept on retrying to create them again...  according to RFC emails should be case insensitive.

email invites should be case insensitive, they should be created and terraform complete the job/exit successfully

These users whould be created in lower case.  This appears to be the API that it is calling:
https://api.aiven.io/v1/project/project-name/invite?limit=999
Tested in `3.13.3`
@ivan-savciuc I created a bug here to keep track of this issue, thanks!

Serpentiel:

makes sure emails are lowercase and valid

uppercase emails are not accepted by our backend

resolves #1450

## Why this way

it should be this way

### DIFF
--- a/internal/schemautil/schemautil.go
+++ b/internal/schemautil/schemautil.go
@@ -2,6 +2,7 @@ package schemautil
 
 import (
 	"fmt"
+	"net/mail"
 	"net/url"
 	"regexp"
 	"strconv"
@@ -219,6 +220,22 @@ func ValidateHumanByteSizeString(v interface{}, k string) (ws []string, errors [
 	if _, err := units.RAMInBytes(v.(string)); err != nil {
 		return ws, append(errors, fmt.Errorf("%q: invalid human readable byte size", k))
 	}
+	return
+}
+
+// ValidateEmailAddress is a ValidateFunc that ensures a string is a valid email address
+func ValidateEmailAddress(v any, k string) (ws []string, errors []error) {
+	addr, err := mail.ParseAddress(v.(string))
+	if err != nil {
+		errors = append(errors, err)
+
+		return
+	}
+
+	if strings.ToLower(addr.Address) != addr.Address {
+		errors = append(errors, fmt.Errorf("%q: invalid email address", k))
+	}
+
 	return
 }
 
--- a/internal/sdkprovider/service/account/account_team_member.go
+++ b/internal/sdkprovider/service/account/account_team_member.go
@@ -26,10 +26,14 @@ var aivenAccountTeamMemberSchema = map[string]*schema.Schema{
 		Description: userconfig.Desc("An account team id").ForceNew().Build(),
 	},
 	"user_email": {
-		Type:        schema.TypeString,
-		Required:    true,
-		ForceNew:    true,
-		Description: userconfig.Desc("Is a user email address that first will be invited, and after accepting an invitation, he or she becomes a member of a team.").ForceNew().Build(),
+		Type:     schema.TypeString,
+		Required: true,
+		ForceNew: true,
+		Description: userconfig.Desc(
+			"Is a user email address that first will be invited, and after accepting an invitation, he " +
+				"or she becomes a member of a team. Should be lowercase.",
+		).ForceNew().Build(),
+		ValidateFunc: schemautil.ValidateEmailAddress,
 	},
 	"invited_by_user_email": {
 		Type:        schema.TypeString,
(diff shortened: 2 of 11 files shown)


byashimov:
1. The original issue refers to v3
2. Why this infinite loop even happens? Does it ignore an error in create handler or something?

ivan-savciuc:
It is also relevant to v4

internal/sdkprovider/service/account/account_team_member.go
```diff
@@ -76,7 +77,7 @@ func resourceAccountTeamMemberCreate(ctx context.Context, d *schema.ResourceData
 	client := m.(*aiven.Client)
 	accountID := d.Get("account_id").(string)
 	teamID := d.Get("team_id").(string)
-	userEmail := d.Get("user_email").(string)
+	userEmail := strings.ToLower(d.Get("user_email").(string))
```
ivan-savciuc, line 80: I think we should instead add validation to the `user_email` since we convert email to lowercase without clear communication on behalf of the user. 

Instead we should error during the `plan` operation

ivan-savciuc (changes requested):
Consider adding validation to the field instead of force converting to the lowercase string!

Serpentiel:
> Why this infinite loop even happens?

because the emails are lowercase according to the RFC, and when an uppercase email is provided, our backend replies with a lowercased version of it, and `Read` infinitely waits for a different email, e.g:

- in Terraform you set it to be `A@xyz.xyz`
- backend replies with `a@xyz.xyz`
- Terraform infinitely waits for `A@xyz.xyz`

internal/sdkprovider/service/account/account_team_member.go
```diff
@@ -26,10 +26,14 @@ var aivenAccountTeamMemberSchema = map[string]*schema.Schema{
 		Description: userconfig.Desc("An account team id").ForceNew().Build(),
 	},
 	"user_email": {
-		Type:        schema.TypeString,
-		Required:    true,
-		ForceNew:    true,
-		Description: userconfig.Desc("Is a user email address that first will be invited, and after accepting an invitation, he or she becomes a member of a team.").ForceNew().Build(),
+		Type:     schema.TypeString,
```
ivan-savciuc, line 29: Perhaps we can use https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/validation/strings.go#L93C6-L93C17 `validation.StringMatch()`?

internal/schemautil/schemautil.go
```diff
@@ -222,6 +222,15 @@ func ValidateHumanByteSizeString(v interface{}, k string) (ws []string, errors [
 	return
 }
 
+// ValidateEmailAddress is a ValidateFunc that ensures a string is a valid email address
+func ValidateEmailAddress(v any, k string) (ws []string, errors []error) {
+	if ok, _ := regexp.MatchString(`^[a-z_\-.]+@([\w-]+\.)+[\w-]{2,4}$`, v.(string)); !ok {
```
ivan-savciuc, line 227: Incorrect regexp: `test+123abs@gmail.com` a valid email but will trigger an error.

internal/sdkprovider/service/account/account_team_member.go
Serpentiel: it's better to have it as a separate function to avoid duplicating this function call all around the code in multiple services

internal/schemautil/schemautil.go
Serpentiel: took a different approach

ivan-savciuc (approved):
LGTM

PR #1454, opened 2023-11-17
docs: change accounts example to orgs and units

staceysalamon-aiven:

Accounts are being replaced by organizations and units in Aiven platform. The accounts example has been updated to show users how to create an organization, units, and projects.


examples/account/README.md
```diff
@@ -1,118 +1,154 @@
-# Aiven Project, Account, and Teams Example
+# Aiven organizations, units, and projects example
```
Serpentiel, line 1: this has to be created separately from `examples/account`, e.g. as `examples/organization`, let's not delete the old docs while the resources are still partially supported

Serpentiel (changes requested):

Serpentiel:
@staceysalamon-aiven any news in regards to this PR? If not, I propose to close it.

PR #1457, opened 2023-11-24
feat(service): add connection_info field to components

### LINKED ISSUE #1444: Provide Kafka SASL connect uri on aiven_kafka
## What is currently missing?

There doesn't seem to be any output from `aiven_kafka` that provides the SASL connect uri.
Currently, you have to do a workaround like:

```
locals {
  sasl_uri = [for component in aiven_kafka.kafka.components :
  "${component.host}:${component.port}" if component.kafka_authentication_method == "sasl"]
}
```

Serpentiel:

adds `connection_info` field to components of all services

resolves #1444 

## Why this way

to provide with a convenient way to retrieve `host:port` string and ease the process of usage of those fields


ivan-savciuc (approved):
LGTM

PR #1459, opened 2023-11-27
chore(service): rename connection_info field to connection_uri

Serpentiel:

renames services' `connection_info` field to `connection_uri`

## Why this way

better naming


byashimov (approved):
lgtm

PR #1461, opened 2023-12-04
fix(kafka_topic): remove expect non empty diff in test

byashimov:

Should not expect non-empty plan:

> When a test is ran, Terraform runs plan, apply, refresh, and then final plan for each TestStep in the TestCase. If the last plan results in a non-empty plan, Terraform will exit with an error. This enables developers to ensure that configurations apply cleanly. In the case of introducing regression tests or otherwise testing specific error behavior, TestStep offers a boolean field [ExpectNonEmptyPlan](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep.ExpectNonEmptyPlan) as well ExpectError regex field to specify ways the test framework can handle expected failures. If these properties are omitted and either a non-empty plan occurs or an error encountered, Terraform will fail the test.

### DIFF
--- a/internal/sdkprovider/service/kafkatopic/kafka_topic_test.go
+++ b/internal/sdkprovider/service/kafkatopic/kafka_topic_test.go
@@ -401,8 +401,7 @@ func TestAccAivenKafkaTopic_recreate_missing(t *testing.T) {
 			},
 			{
 				// Step 3: recreates the topic
-				ExpectNonEmptyPlan: true,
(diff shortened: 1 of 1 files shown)


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/7087858067)

Serpentiel (approved):
lgtm

PR #1465, opened 2023-12-05
feat(sweeper): add billing group support

Serpentiel:

adds billing group support for sweeper

## Why this way

to avoid congesting the organizations with irrelevant billing groups


byashimov (approved):
lgtm

PR #1471, opened 2023-12-07
fix(account_team_member): no err on deletion if it does not exist

Serpentiel:

no err on `aiven_account_team_member` deletion if it does not exist

## Why this way

to avoid edge cases, e.g. invitations

### DIFF
--- a/internal/sdkprovider/service/account/account_team_member.go
+++ b/internal/sdkprovider/service/account/account_team_member.go
@@ -205,21 +205,19 @@ func resourceAccountTeamMemberDelete(ctx context.Context, d *schema.ResourceData
 	}
 
 	// delete account team member
-	found := false
 	for _, m := range r.Members {
 		if m.UserEmail == userEmail {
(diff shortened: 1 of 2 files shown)


internal/sdkprovider/service/account/account_team_member.go
```diff
@@ -205,21 +205,15 @@ func resourceAccountTeamMemberDelete(ctx context.Context, d *schema.ResourceData
...
 			}
-			found = true
 			break
 		}
 	}
 
-	if !found {
-		return diag.Errorf("user with email %q is not a part of the team %q", userEmail, teamID)
-	}
-
```
byashimov, line 223: Please leave a comment to explain why we do this.

byashimov (changes requested):

internal/sdkprovider/service/account/account_team_member.go
Serpentiel: done!

byashimov (approved):
lgtm

PR #1473, opened 2023-12-12
fix(mirrormaker_replication_flow): race on create/update/delete opera…

byashimov:

Fixes race issues on write operations.


byashimov:
Success [tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/7192941984)

byashimov:
No additional test has been added, cause the test case needs three kafka to run.

PR #1475, opened 2023-12-13
fix(mirrormaker_replication_flow): race on create/update/delete operations

byashimov:

A workaround to fix API race issues.


Serpentiel (approved):
lgtm

PR #1476, opened 2023-12-13
feat(service): expose tech_emails

jeff-held-aiven:
Allow users to set `tech_emails` member of service.

### DIFF
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -49,6 +49,16 @@ const (
 	ServiceTypeClickhouse       = "clickhouse"
 )
 
+var TechEmailsResourceSchema = &schema.Resource{
+	Schema: map[string]*schema.Schema{
+		"email": {
+			Type:        schema.TypeString,
+			Description: "An email address to contact for technical issues",
+			Required:    true,
+		},
+	},
+}
+
 func ServiceCommonSchema() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"project": CommonSchemaProjectReference,
@@ -267,6 +277,12 @@ func ServiceCommonSchema() map[string]*schema.Schema {
 				},
 			},
 		},
+		"tech_emails": {
+			Type:        schema.TypeSet,
+			Elem:        TechEmailsResourceSchema,
+			Optional:    true,
+			Description: "Defines the email addresses that will receive alerts about upcoming maintenance updates or warnings about service instability.",
+		},
 	}
 }
 
@@ -373,6 +389,10 @@ func ResourceServiceRead(ctx context.Context, d *schema.ResourceData, m interfac
 		return diag.Errorf("unable to set tag's in schema: %s", err)
 	}
 
+	if err := d.Set("tech_emails", getTechnicalEmailsForTerraform(d, "tech_emails", s)); err != nil {
+		return diag.Errorf("unable to set tech_emails in schema: %s", err)
+	}
+
 	return nil
 }
 
@@ -429,6 +449,7 @@ func resourceServiceCreate(ctx context.Context, d *schema.ResourceData, m interf
 			DiskSpaceMB:           diskSpace,
 			UserConfig:            cuc,
 			StaticIPs:             FlattenToString(d.Get("static_ips").(*schema.Set).List()),
+			TechnicalEmails:       getContactEmailListForAPI(d, "tech_emails"),
 		},
 	)
 	if err != nil {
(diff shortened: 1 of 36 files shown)


internal/sdkprovider/service/project/project.go
```diff
@@ -79,7 +79,7 @@ var aivenProjectSchema = map[string]*schema.Schema{
 		Type:        schema.TypeSet,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
-		Description: "Defines the email addresses that will receive alerts about upcoming maintenance updates or warnings about service instability. It is  good practice to keep this up-to-date to be aware of any potential issues with your project.",
+		Description: "Defines the email addresses that will receive alerts about upcoming maintenance updates or warnings about service instability. It is good practice to keep this up-to-date to be aware of any potential issues with your project.",
```
Serpentiel, line 82: typo, should be _it is a good practice_ (missing _a_)

Serpentiel (changes requested):

internal/schemautil/service.go
```diff
@@ -572,6 +594,20 @@ func getDefaultDiskSpaceIfNotSet(ctx context.Context, d *schema.ResourceData, cl
 	return diskSpace, nil
 }
 
+func getTechnicalEmailsForTerraform(d *schema.ResourceData, field string, s *aiven.Service) *schema.Set {
+	_, existsBefore := d.GetOk(field)
+	if !existsBefore && len(s.TechnicalEmails) == 0 {
```
byashimov, line 599: nit: technically "before" is `s.TechnicalEmails`. It is "now" or more conventional is "ok"

internal/schemautil/service.go
```diff
@@ -864,3 +904,17 @@ func DatasourceServiceRead(ctx context.Context, d *schema.ResourceData, m interf
 
 	return diag.Errorf("common %s/%s not found", projectName, serviceName)
 }
+
+func getContactEmailListForAPI(d *schema.ResourceData, field string) *[]aiven.ContactEmail {
+	var results []aiven.ContactEmail
+	valuesInterface, ok := d.GetOk(field)
+	if ok && valuesInterface != nil {
```
byashimov, line 911: What was the case with nil value?

internal/schemautil/service.go
```diff
@@ -864,3 +904,17 @@ func DatasourceServiceRead(ctx context.Context, d *schema.ResourceData, m interf
...
+		for _, emailInterface := range valuesInterface.(*schema.Set).List() {
+			results = append(results, aiven.ContactEmail{Email: emailInterface.(map[string]interface{})["email"].(string)})
+		}
+	}
+	if results == nil {
+		return &[]aiven.ContactEmail{}
+	}
+	return &results
```
byashimov, line 919: You can safely remove 916-918. `results` is the same as that part.

internal/sdkprovider/service/redis/redis_test.go
```diff
@@ -27,6 +27,32 @@ func TestAccAiven_redis(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet(resourceName, "service_uri"),
+				),
+			},
+			{
+				Config: testAccRedisRemoveEmailsResource(rName),
+				Check: resource.ComposeTestCheckFunc(
+					acc.TestAccCheckAivenServiceCommonAttributes("data.aiven_redis.common"),
+					testAccCheckAivenServiceRedisAttributes("data.aiven_redis.common"),
+					resource.TestCheckResourceAttr(resourceName, "tag.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "tech_emails.#", "0"),
```
byashimov, line 55: Nice test! I believe you can remove duplicate checks from 56 to 69.

internal/schemautil/service.go
byashimov: Ah, you see, `*[]` doesn't make sense. Cause you always return a pointer to a slice. Hence it sends `null`, because it has a type, but not a length. https://go.dev/play/p/FJLtc1sE-xf
You need to return `nil` to make your `if` check make sense.
I believe that the API just ignores `null` and that's why it works.

So you can do this:
```go
results := make([]aiven.ContactEmail, 0)
valuesInterface, ok := d.GetOk(field)
if ok {
    ...
}
return &results
```

It will always send an array. Works exactly like if you replace `*[]email` with `[]email` in the client.

byashimov (changes requested):
minor changes are required, otherwise lgtm

byashimov (approved):
lgtm

byashimov (approved):
lgtm

PR #1484, opened 2023-12-18
feat(org_user): deprecating resource, new datasource

ivan-savciuc:

This PR is first in the series that is bringing organization user group members support, we are deprecating `aiven_organization_user` resource and we instead users will be using a corresponding data source. 

This is required as we have the security requirement to block user sign-up via API only allow sign-up via Console. In the near future we will add support to provision users from the IdP.

Implications: it is going to be impossible to test organization users in acceptance tests.


Serpentiel (changes requested):

internal/sdkprovider/service/organization/organization_user_data_source_test.go
```diff
@@ -1,34 +0,0 @@
...
-import (
-	"testing"
-
-	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
-	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-
-	acc "github.com/aiven/terraform-provider-aiven/internal/acctest"
-)
-
-func TestAccAivenOrganizationUserDataSource_basic(t *testing.T) {
```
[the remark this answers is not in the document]
ivan-savciuc: Unfortunately not, I describe that in the PR from above. In order to that I one has not create an organization user via web console, and then use this data-source to import user information to the Terraform.

docs/data-sources/organization_user.md
```diff
@@ -17,12 +17,11 @@ The Organization User data source provides information about the existing Aiven
...
-- `user_email` (String) This is a user email address that first will be invited, and after accepting an invitation, they become a member of the organization. Should be lowercase. This property cannot be changed, doing so forces recreation of the resource.
+- `organization_id` (String) The unique organization ID.
+- `user_id` (String) The unique organization user ID
 
 ### Read-Only
 
-- `accepted` (Boolean) This is a boolean flag that determines whether an invitation was accepted or not by the user. `false` value means that the invitation was sent to the user but not yet accepted. `true` means that the user accepted the invitation and now a member of an organization.
 - `create_time` (String) Time of creation
 - `id` (String) The ID of this resource.
-- `invited_by` (String) The email address of the user who sent an invitation to the user.
```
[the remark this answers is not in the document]
ivan-savciuc: Message from our back-end team: 

> One big change we should introduce first is to the users ressource. The users ressource should be read-only meaning we will remove support for inviting users to Aiven via TF provider. This is required as we have the security requirement to block user sign-up via API only allow sign-up via Console. In the near future we will add support to provision users from the IdP.

This is the reason why we had to do that.

internal/sdkprovider/service/organization/organization_user_data_source.go
```diff
@@ -6,25 +6,96 @@ import (
...
+		}
+	}
+
+	if found == 0 {
+		return diag.Errorf("organization user %s not found in organization %s", userEmail, organizationID)
+	}
+
+	if found > 1 {
+		return diag.Errorf("multiple organization users %s found in organization %s", userEmail, organizationID)
+	}
```
byashimov, line 96: nit: I would redesign this a bit just not to make grow the part with the loop with more "set"s when new fields added. Both id and email are [mandatory](https://api.aiven.io/doc/#tag/Users/operation/OrganizationUserList), so you can omit empty checks.

```go
        var found int
	var userInfo *aiven.OrganizationUserInfo
	for _, user := range rm.Users {
		if user.UserID == userId || user.UserInfo.UserEmail == userEmail {
			userInfo = &user.UserInfo
          		found++
		}
	}

	if found != 1 {
		return diag.Errorf("expected exactly one user with email %q or id=%q, found %d", userEmail, userId, found)
	}

	if err := d.Set("organization_id", organizationID); err != nil {
		return diag.FromErr(err)
	}
	if err := d.Set("user_email", userInfo.UserEmail); err != nil {
		return diag.FromErr(err)
	}
	if err := d.Set("create_time", user.JoinTime.String()); err != nil {
		return diag.FromErr(err)
	}
	if err := d.Set("user_id", user.UserID); err != nil {
		return diag.FromErr(err)
	}
        // this part can grow now
```

internal/sdkprovider/service/organization/organization_user.go
```diff
@@ -66,100 +72,55 @@ eliminate the member from the organization if one has accepted an invitation pre
...
+			if err := d.Set("user_email", userInfo.UserEmail); err != nil {
 				return diag.FromErr(err)
 			}
-			if err := d.Set("invited_by", invite.InvitedBy); err != nil {
+			if err := d.Set("create_time", user.JoinTime.String()); err != nil {
 				return diag.FromErr(err)
 			}
-			if err := d.Set("create_time", invite.CreateTime.String()); err != nil {
+			if err := d.Set("user_id", user.UserID); err != nil {
 				return diag.FromErr(err)
```
byashimov, line 122: Do you expect create_time and invited_by to disappear?

internal/sdkprovider/service/organization/organization_user_data_source.go
ivan-savciuc: Updated code to something similar

internal/sdkprovider/service/organization/organization_user.go
ivan-savciuc: `create_time` will stay and `invited_by` is getting deprecated, we no longer can get data for this field since we no longer can invite people or read the invitations.

byashimov (dismissed):
lgtm

PR #1485, opened 2023-12-18
ci(test): proper project name prefix var substitution

Serpentiel:

- proper project name prefix var substitution.
- additionally adds support for `sweep.yml` (forgot to do this earlier)

## Why this way

resolves CI error https://github.com/aiven/terraform-provider-aiven/actions/runs/7247944868/job/19742908213


ivan-savciuc (approved):
LGTM

PR #1491, opened 2023-12-21
fix(kafka_topic): migration from v3

ivan-savciuc:

Fix Kafka Topic migration from v3 to v4


Serpentiel (approved):
lgtm

PR #1494, opened 2023-12-27
fix(service): v3 to v4 migration cloud_name field

ivan-savciuc:

fix migration issues from v3 to v4 related to cloud_name diff


Serpentiel (approved):
lgtm

PR #1504, opened 2024-01-03
docs(kafkatopic): return dot instead of underscore

Serpentiel:

return dot instead of underscore in docs of kafkatopic

## Why this way

it's the proper way


ivan-savciuc (approved):
LGTM

PR #1520, opened 2024-01-10
refactor(acctest): skip option support

Serpentiel:

skip option support for `acctest`

## Why this way

makes code cleaner


ivan-savciuc (approved):
LGTM

PR #1542, opened 2024-01-18
ci: add sweep check

ivan-savciuc:

Add sweeper check CI job

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -65,3 +65,16 @@ jobs:
 
       - run: make docs
       - run: git diff --exit-code
+  
+  sweep_check:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


.github/workflows/sweep-check.yaml
```diff
@@ -0,0 +1,33 @@
+name: Sweepers Check
```
Serpentiel, line 1: please integrate this into an existing `lint.yaml` workflow

Serpentiel (changes requested):

.github/workflows/sweep-check.yaml
ivan-savciuc: ok

byashimov (approved):
lgtm

PR #1544, opened 2024-01-19
fix(org): org user datasource

ivan-savciuc:

Fix `aiven_organization_user` datasource get method


byashimov (approved):
lgtm

PR #1546, opened 2024-01-22
fix(org): org user and group member model fill

ivan-savciuc:

Fix org user and group model fill

### DIFF
--- a/internal/plugin/service/organization/organization_application_user.go
+++ b/internal/plugin/service/organization/organization_application_user.go
@@ -153,11 +153,13 @@ func (r *organizationApplicationUser) fillModel(
 	}
 
 	if appUser == nil {
-		return fmt.Errorf(errmsg.AivenResourceNotFound, r.TypeName(), model.ID.ValueString())
+		return fmt.Errorf(
+			errmsg.AivenResourceNotFound,
+			r.TypeName(),
+			model.ID.ValueString())
 	}
 
 	model.Name = types.StringValue(appUser.Name)
-
 	model.Email = types.StringValue(appUser.UserEmail)
 
 	return err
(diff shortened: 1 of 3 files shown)


internal/plugin/service/organization/organization_application_user.go
```diff
@@ -143,21 +143,23 @@ func (r *organizationApplicationUser) fillModel(
...
 		if u.UserID == model.UserID.ValueString() {
-			appUser = &u
+			appUser = u
 			break
 		}
 	}
 
-	if appUser == nil {
-		return fmt.Errorf(errmsg.AivenResourceNotFound, r.TypeName(), model.ID.ValueString())
+	if appUser.UserID == "" {
```
Serpentiel, line 155: looks like this isn't needed, could we please resort back to the original way it's done via pointer?

internal/plugin/service/organization/organization_user_group_member.go
```diff
@@ -173,7 +173,9 @@ func (r *organizationUserGroupMembersResource) fillModel(
 		)
 	}
 
-	model.LastActivityTime = types.StringValue(member.LastActivityTime.String())
+	if member.LastActivityTime != nil {
```
Serpentiel, line 176: please do `model.LastActivityTime = util.ValueOrDefault(member.LastActivityTime.String(), types.StringNull())`

internal/plugin/service/organization/organization_user_group_member.go
```diff
@@ -152,16 +152,16 @@ func (r *organizationUserGroupMembersResource) fillModel(
...
 	for _, m := range list.Members {
 		if m.UserID == model.UserID.ValueString() {
-			member = &m
+			member = m
 			break
 		}
 	}
 
-	if member == nil {
+	if member.UserID == "" {
```
Serpentiel, line 164: this isn't clear why we are switching from the usage of the pointer here as well, could you please elaborate?

Serpentiel (changes requested):

Serpentiel (approved):
lgtm

PR #1550, opened 2024-01-23
feat(flink): add pre deletion check

ivan-savciuc:

Add Flink service deletion check, it is impossible to delete a Flink service if we have active Deployments with Jobs that are managed outside of Terraform via Web UI. Expectation is that users clean up all the running Jobs via Web UI and remove all deployments prior to deletion of Flink service.


Serpentiel (approved):
lgtm

PR #1554, opened 2024-01-24
feat: dragonfly support

Serpentiel:

adds dragonfly support in disabled state until it's available

## Why this way

currently dragonfly is not GA; this PR adds full support for the new service, we need to enable it once it's available

### DIFF
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -47,6 +47,7 @@ const (
 	ServiceTypeM3Aggregator     = "m3aggregator"
 	ServiceTypeFlink            = "flink"
 	ServiceTypeClickhouse       = "clickhouse"
+	ServiceTypeDragonfly        = "dragonfly"
 )
 
 var TechEmailsResourceSchema = &schema.Resource{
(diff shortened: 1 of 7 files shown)


internal/schemautil/service.go
```diff
@@ -821,12 +822,10 @@ func copyConnectionInfoFromAPIResponseToTerraform(
 	props := make(map[string]interface{})
 
 	switch serviceType {
-	case "cassandra":
```
Serpentiel, line 824: the removed ones don't do anything, Go syntax for `switch` is this:

```go
switch x {
case "1", "foo", "bar":
    // ...
case "xyz":
    // ...
default:
    // ...
 }
 ```

File: internal/schemautil/service.go [hunk 2]
```diff
@@ -855,15 +852,8 @@ func copyConnectionInfoFromAPIResponseToTerraform(
...
 		props["replica_uri"] = connectionInfo.PostgresReplicaURI
 		props["max_connections"] = metadata.(map[string]interface{})["max_connections"]
-	case "clickhouse":
-	case "redis":
 	case "flink":
 		props["host_ports"] = connectionInfo.FlinkHostPorts
-	case "kafka_mirrormaker":
-	case "m3db":
-	case "m3aggregator":
-	default:
```
Serpentiel, line 865: we don't need `default` here to cause a `panic`, this is an extra place of an unnecessary maintenance—instead, if there's an irregular service that requires additional connection options, we will know about it via failing acceptance tests

Serpentiel: It's not possible to run into the scenario where you get a `panic` unless there's a new resource type added to the provider, thus we should default for doing nothing instead of `panic`, because we already add the resource in the `provider.go`, so there's no way to have the unexpected resource type


> Re: acceptance tests: could you explain how they should catch irregular service types?

this specific code converts connection information array passed from the API (e.g. `hostname`, `host`, `port`, etc.) into the Terraform state format

generally, those don't have any extra fields except for a few specific service types, that are `opensearch`, `elasticsearch`, `influxdb`, `kafka`, `pg`, `flink`

if some service starts returning any additional unexpected fields in future, we will know about it via failing acceptance tests

the `panic` that was removed isn't going to prevent us from knowing about it

Issue #858, opened 2022-08-05
consider adding more linters


Serpentiel:
## What is currently missing?

some important linters are missing which could:

- prevent appearance of future bugs
- fix existing bugs
- improve overall performance and security of the provider

## How could this be improved?

add the following linters:

- `durationcheck` _(check for two durations multiplied together)_
- `errorlint` _(errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.)_
- `forbidigo` _(Forbids identifiers)_
- `forcetypeassert` _(finds forced type assertions)_
- `goconst` _(Finds repeated strings that could be replaced by a constant)_
- `gocyclo` _(Computes and checks the cyclomatic complexity of functions)_
- `gomnd` _(An analyzer to detect magic numbers.)_
- `gosec` _(Inspects source code for security problems)_
- `prealloc` _(Finds slice declarations that could potentially be pre-allocated)_
- `typecheck` _(Like the front-end of a Go compiler, parses and type-checks Go code)_

## Is this a feature you would work on yourself?

* [ ] I plan to open a pull request for this feature.

rominf:
It looks like `forcetypeassert` is not needed: [it's not developed anymore](https://github.com/gostaticanalysis/forcetypeassert/graphs/code-frequency), and the same functionality [is provided](https://github.com/gostaticanalysis/forcetypeassert/issues/15) by `errcheck`.

PR #1561, opened 2024-01-26
ci(lint): add errorlint

rominf:

Add https://github.com/polyfloyd/go-errorlint to find code that will cause problems with Go's error wrapping scheme.

This is partial implementation of #858.

## Why this way

Implementation is based on recommendations from https://pkg.go.dev/errors.


Serpentiel (approved):
lgtm

PR #1562, opened 2024-01-26
docs: update app users description

staceysalamon-aiven:

Add more info to the application users description about this feature and that it is in the LA stage.


byashimov (approved):
Thanks!

PR #1564, opened 2024-01-26
docs: add import and usage example for `aiven_flink_application_deployment`, fix and extend flink example

rriski:

* Adds usage and import example to the `aiven_flink_application_deployment` resource
* Fixes the flink example by removing extra quotes and extends the example by adding `aiven_flink_application`, `aiven_flink_application_version` and `aiven_flink_application_deployment` to the example

Resolves: #xxxxx.

## Why this way

Improves docs.


Serpentiel (approved):
lgtm

PR #1567, opened 2024-01-29
fix(org): removal of org group member

ivan-savciuc:

Fix org group member deletion


Serpentiel (approved):
lgtm

PR #1574, opened 2024-02-01
docs: add organizations example

staceysalamon-aiven:

Add an example for creating an organization that contains units and projects. Also removes the accounts example as accounts have been superseded by orgs in Aiven.


Serpentiel (approved):
lgtm

PR #1576, opened 2024-02-03
ci(lint): add prealloc

rominf:

Add https://github.com/alexkohler/prealloc to find slice declarations
that could potentially be pre-allocated and address linter errors.

This is partial implementation of #858.

## Why this way

`make` + assign pattern was used in codebase already and it works faster than `append`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,6 +17,7 @@ linters:
     - unused
     - stylecheck
     - errorlint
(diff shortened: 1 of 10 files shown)


rominf:
Feel free to run acceptance tests if needed.

Serpentiel (approved):
lgtm

PR #1578, opened 2024-02-05
ci(checkout): add missing fetch-depth

Serpentiel:

adds missing `fetch-depth` to workflows

## Why this way

it's better to have it than the other way around


byashimov (approved):
lgtm

PR #1580, opened 2024-02-06
docs: create new get started example

staceysalamon-aiven:

Create a new example for getting started with Aiven Provider for Terraform to replace the outdated example using teams.


examples/get-started/README.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+The output is similar to the following:
+
+```sh
+Terraform v1.6.2
++ provider registry.terraform.io/aiven/aiven v4.9.2
+```
+
+2. Clone this repository.
```
Serpentiel, line 29: cloning the repository is only needed for development purposes, for regular usage of the Terraform provider it's not needed

Serpentiel (changes requested):

examples/get-started/README.md
staceysalamon-aiven: This step was added here and in other examples in anticipation of creating a dedicated examples repo. That way everything can just be moved without editing the steps - does that work or is it better to still generalize the instructions?

Serpentiel: Ah, I see the point now. We should definitely consider a dedicated examples repo in future. For now, let's get forward with it.

Serpentiel (approved):
lgtm

Issue #1596, opened 2024-02-15
Fail terraform plan that would result in the deletion of aiven_kafka_topic that has termination_protection enabled


apjm:
## What happened?

With provider version = "4.9.3"

For an aiven_kafka_topic resource with termination_protection = true, remove the config for the topic. Run `terraform plan` to produce a valid plan including the destruction of the topic, eg.

```
# module.module_id.aiven_kafka_topic.this["x"] will be destroyed
  # (because key ["x"] is not in for_each map)
  - resource "aiven_kafka_topic" "this" {
      - id                     = "topic_id" -> null
      - partitions             = 3 -> null
      - project                = "project_id" -> null
      - replication            = 3 -> null
      - service_name           = "service_id" -> null
      - termination_protection = true -> null
      - topic_name             = "x" -> null

      - config {
        ...
        }

    }
```

Running tf apply for the same plan errors with

`Error: cannot delete kafka topic when termination_protection is enabled`

## What did you expect to happen?

Terraform plan should error if it would include destroy for a topic that has termination protection enabled.

## What else do we need to know?

Terraform version 1.6.5

byashimov:
@Serpentiel I think it is totally valid, yet we can't implement it. Not sure if even possible on the plugin framework.

Serpentiel:
@byashimov well, yes, but what we can't do—we can't do 🙁

PR #1598, opened 2024-02-16
fix: termination_protection diff `true -> null`

byashimov:

When `termination_protection = true` is removed the diff shows `true -> null`, and sends zero-value, which is `false` [service.go](https://github.com/aiven/terraform-provider-aiven/blob/d6b602b6e1f136ead0e9a385e23c9e59f99f5b0d/internal/schemautil/service.go#L547-L547).
This PR outputs `true -> false` transition, because explicit is better than implicit.


byashimov:
[Tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/7929019268).

Serpentiel:
Discovered in #1596

Serpentiel (approved):
lgtm

PR #1601, opened 2024-02-18
docs: add teams deprecation info to relevant resources

staceysalamon-aiven:

Add a note to the account team resources letting users know that teams are being replaced by groups and link to the [migration guide](https://github.com/aiven/aiven-docs/pull/67).


internal/sdkprovider/service/account/account_team.go
```diff
@@ -41,7 +40,7 @@ var aivenAccountTeamSchema = map[string]*schema.Schema{
 
 func ResourceAccountTeam() *schema.Resource {
 	return &schema.Resource{
-		Description:   "The Account Team resource allows the creation and management of an Account Team.",
+		Description:   `Creates and manages a team.`,
```
Serpentiel, line 43: Could you please set the property `Deprecated` there and fill it with information as well? And in other resources, too.

Serpentiel (changes requested):

Serpentiel (approved):
lgtm

PR #1609, opened 2024-02-21
fix(aiven_kafka_schema): update schema version

### LINKED ISSUE #1608: Version attribute for aiven_kafka_schema isn't incremented on apply
## What happened?

We see that the `version` attribute for `aiven_kafka_schema` isn't getting incremented on apply, even if a new version is created. We observe this using  an `output`.

For example (all apply's below change the schema in some way): 

```
Apply 1 -> version = 1
Apply 2 -> version = 1
Apply 3 -> version = 2
...
```

So it appears that the provider observes the previous version bump only on a subsequent plan/apply.

Version should have been incremented right after apply.

Terraform version `1.5.5`, Aiven terraform provider version `4.14.0`

byashimov:

Fixes version issues for `aiven_kafka_schema`:

1. When a `aiven_kafka_schema` is updated it doesn't necessarily get the latest _version_.
   Instead, if it just reverts to a previous state it get's an existing _version_.
   Takes schema ID from "update" response and fetches the valid _version_.
2. Version list is not updated instantly, there is a tiny lag. Polls until finds the given schema ID from the "update" response. This case is flaky, hence it can't be covered in tests.

Resolves #1608.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ nav_order: 1
 <!-- TODO: uncomment when dragonfly is supported -->
 <!-- - Dragonfly support -->
 
+- Fix `aiven_kafka_schema` version update
+
 ## [4.14.0] - 2024-02-20
 
 - Refactor deprecated `resource.StateChangeConf`/`resource.StateRefreshFunc` usage to their equivalent with `retry`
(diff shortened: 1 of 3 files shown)


byashimov:
[Tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/7988998938), flink fails, as expected.

Serpentiel (approved):
lgtm

PR #1611, opened 2024-02-22
fix(flink): poll application deployment deletion

byashimov:
Tests run fast and fail to delete Flink Application Deployment in transitional statuses.
The state machine for this type of resource is quite complicated. Instead of trying to replicate the transition flow this PR proposes different approach: it calls deletion over and over until succeeds or exceeds context timeout.


internal/sdkprovider/service/flink/flink_application_deployment.go
```diff
@@ -135,42 +134,30 @@ func resourceFlinkApplicationDeploymentDelete(
...
-		MinTimeout: 1 * time.Second,
-	}
 
-	_, err = conf.WaitForStateContext(ctx)
-	if err != nil {
-		return diag.Errorf("error waiting for Flink Application Deployment to become canceled: %s", err)
-	}
+			// Must be canceled before deleted
+			_, err = client.FlinkApplicationDeployments.Cancel(ctx, project, serviceName, applicationID, deploymentID)
+			if err == nil {
```
Serpentiel, line 153: This will run forever when the states are:

- `INITIALIZING`
- `CANCELLING`
- `SAVING_AND_STOP_REQUESTED`
- `SAVING_AND_STOP`
- `FAILED`
- `CANCELED` (quite obvious, because it’s already cancelled)
- `FINISHED`
- `DELETE_REQUESTED`
- `DELETING`

internal/sdkprovider/service/flink/flink_application_deployment.go
```diff
@@ -135,42 +134,30 @@ func resourceFlinkApplicationDeploymentDelete(
...
+				continue
+			}
 
-	_, err = client.FlinkApplicationDeployments.Delete(ctx, project, serviceName, applicationID, deploymentID)
```
Serpentiel, line 168: We need to call `Delete` to delete it.

Serpentiel (changes requested):
I'm not sure if this approach is correct. This will do nothing if the job is in these states, instead of deleting it:

- `INITIALIZING`
- `FAILING`
- `CANCELLING`
- `SAVING_AND_STOP_REQUESTED`
- `SAVING_AND_STOP`
- `DELETING`

Terraform will report that the job was deleted successfully, when it wasn't.

Additionally, please update the test with the correct SQL, like in https://github.com/aiven/terraform-provider-aiven/pull/1606/files#diff-71cae1e8db8b50bfe944747bd7d61f9128e4a137878d4b00e613497bdc415f51.

internal/sdkprovider/service/flink/flink_application_deployment.go
```diff
@@ -135,42 +134,30 @@ func resourceFlinkApplicationDeploymentDelete(
...
-			if err != nil {
-				return nil, "", err
+	// Flink Application Deployment has a quite complicated state machine
+	// https://api.aiven.io/doc/#tag/Service:_Flink/operation/ServiceFlinkDeleteApplicationDeployment
+	// Retries until succeeds or exceeds the timeout
+	for {
+		select {
+		case <-ctx.Done():
+			// The context itself already comes with delete timeout
+			return diag.Errorf("can't delete Flink Application Deployment %q: %s", ctx.Err())
```
Serpentiel, line 144: Incorrect `diag.Errorf` call, there's more placeholders than arguments.

Serpentiel (approved):
lgtm

PR #1614, opened 2024-02-23
test(flink_application_version): fix incorrect sql

Serpentiel:
## Merging strategy

- [x] Merge
- [ ] Squash

- fixes incorrect sql in `aiven_flink_application_version` resource's test, which resulted the job being in infinite `RESTARTING` state right after its creation
- fixes heredoc formatting in all docs and tests

## Why this way

- sql should be correct in the test to reduce the load the test generates on the service
  - maintainers should later decide whether this is incorrect to report success on creation of such job or not
- heredoc formatting was broken


byashimov (approved):
lgtm

PR #1615, opened 2024-02-23
refactor(userconfig): generate types list, rename integrations

byashimov:

1. renames `integration` userconfigs to `serviceintegration`. The `integration` variable is quite common in existing code, better rename the module to exclude collisions
2. generates types list to automatically add those to schemas, like integrations and integration endpoints


Serpentiel (commented):
lgtm, but I have one question

Serpentiel (approved):
lgtm

PR #1623, opened 2024-02-27
refactor(aiven_service_integration_endpoint): new user configs

### LINKED ISSUE #1621: Rsyslog max_message_size being read as 0 instead of 8192
## What happened?

We have a strange issue when bumping versions from `v4.9.2` to `v4.14.0` whereby the Terraform diff suggests we're going to change the Rsyslog `max_message_size` on a service integration endpoint from 0 to 8192

```hcl
  # module.aiven.aiven_service_integration_endpoint.rsyslog["kafka-test"] will be updated in-place
  ~ resource "aiven_service_integration_endpoint" "rsyslog" {
        id              = "test/d75669d0-ed18-48bc-b9e8-89ce439e6101"
        # (4 unchanged attributes hidden)

      ~ rsyslog_user_config {
          ~ max_message_size = 0 -> 8192
            # (5 unchanged attributes hidden)
        }
    }
```

..the Aiven console, however, shows the `max_message_size` to already be set to 8192 (which is our desired value).

If we bump back down to version `v4.9.2` then Terraform shows no diff.

On version `v4.14.0` there should be no Terraform diff.

I've taken a cursory glance at the [v4.9.2..v4.14.0](https://github.com/aiven/terraform-provider-aiven/compare/v4.9.2...v4.14.0) changes and can't see any related change that could have caused it.

We could of course apply the change, but thought it best to check with you first as the diff is unexpected

byashimov:

Replaces old user config generator for `aiven_service_integration_endpoint` type.
Resolves #1621.


byashimov:
Please double check the list of types.

byashimov:
Updates [tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/8063432772).

Serpentiel (approved):
lgtm, good stuff 👍

PR #1624, opened 2024-02-27
docs: improve docs for user group related resources

staceysalamon-aiven:

Updates some info in docs for user group and related resources, including adding example usage and import syntax. Other minor changes to descriptions , fields, etc., in these and other resources. The beta message and forces recreation notices were also slightly reworded.


docs/data-sources/organization.md
```diff
@@ -3,18 +3,18 @@
...
-Retrieves information about an organization from Aiven.
+Gets information about an organization.
 
 ## Example Usage
 
 ```terraform
-data "aiven_organization" "organization1" {
-  name = "<ORGANIZATION_NAME>"
+data "aiven_organization" "main" {
+  name = "Example organization"
```
Serpentiel, line 17: I think we use placeholders in other datasources, maybe a more standardized approach should be used, or this should be changed elsewhere too?

Serpentiel (approved):
lgtm, but a small comment

docs/data-sources/organization.md
staceysalamon-aiven: You're right and I considered sticking with that, but after some research and consideration about what might be easier to work with and read I decided to go with this. I have an epic to go through all of the resources and data sources to standardize the descriptions, example usage, import syntax, etc. :)

Serpentiel: Ah, good. Thank you for the clarification!

PR #1635, opened 2024-03-06
fix: require explicit flag to purge ip_filter

byashimov:


Serpentiel:
I think I have a better solution. I'll check it.

Serpentiel (approved):
lgtm, let's get this merged first, until we figure things out

PR #1639, opened 2024-03-07
docs: add LA note and update app users

staceysalamon-aiven:

- Updates application user documentation with this notice and standardized/improved other descriptions.
- Adds example usage and import syntax.


internal/plugin/service/organization/organization_application_user_token.go
```diff
@@ -108,24 +108,25 @@ func (r *organizationApplicationUserToken) Schema(
...
-				Description: "Identifier of the organization the application user token belongs to.",
+				Description: "The ID of the organization the application user belongs to.",
 				Required:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
 			},
 			"user_id": schema.StringAttribute{
-				Description: "Identifier of the application user the token belongs to.",
+				Description: "The ID of the application user the token belongs to.",
```
Serpentiel, line 129: I think we should word it like: "The ID of the application user the token should belong to.", because the token here wasn't created yet at this stage, and we are creating the token, but it doesn't belongs to it before it was created.

internal/plugin/service/organization/organization_application_user_data_source.go
```diff
@@ -67,22 +67,22 @@ func (r *organizationApplicationUserDataSource) Schema(
...
 				Required:    true,
 			},
 			"name": schema.StringAttribute{
-				Description: "Name of the organization application user.",
+				Description: "Name of the application user.",
 				Computed:    true,
 			},
 			"email": schema.StringAttribute{
-				Description: "Email of the organization application user.",
+				Description: "The auto-generated email address of the application user.",
```
Serpentiel, line 85: I took most of the descriptions from the Aiven Public API docs. Perhaps, this should also be adjusted in them, too?

internal/plugin/util/pluginhelpers.go
```diff
@@ -34,6 +34,16 @@ set the ` + "`PROVIDER_AIVEN_ENABLE_BETA`" + ` environment variable to true.
 `
 }
 
+// LimitedAvailabilityDescription is a helper function that returns a description for features that are in LA.
+func LimitedAvailabilityDescription(description string) string {
```
Serpentiel, line 38: WDYT if we combine `BetaDescription` and `LimitedAvailabilityDescription` in a builder pattern, like [here](https://github.com/aiven/terraform-provider-aiven/blob/ee5c8524ecfb14279d48658fc39a5088bf0f534e/internal/schemautil/userconfig/desc.go)? Please let me know if this is too advanced for you to implement, I'll be willing to assist 🙂

Serpentiel (commented):
lgtm in general, but there are some questions

Serpentiel:
Here's the example to give you an idea:

```go
type EntityType int

const (
	Resource EntityType = iota
	DataSource
)

func (et EntityType) String() string {
	return [...]string{"resource", "datasource"}[et]
}

type AvailabilityType int

const (
	Beta AvailabilityType = iota + 1
	Limited
)

type DescriptionBuilder struct {
	entityType         EntityType
	mainText            string
	availabilityType AvailabilityType
}

func (db *DescriptionBuilder) EntityType(entityType EntityType) *DescriptionBuilder {
	db.entityType = entityType
	return db
}

func (db *DescriptionBuilder) MainText(mainText string) *DescriptionBuilder {
	db.mainText = mainText
	return db
}

func (db *DescriptionBuilder) AvailabilityType(availabilityType AvailabilityType) *DescriptionBuilder {
	db.availabilityType = availabilityType
	return db
}

func (db *DescriptionBuilder) Build() string {
	var builder strings.Builder

	if db.mainText != "" {
		builder.WriteString(db.mainText)
		if db.availabilityType != 0 {
			builder.WriteString(" ")
		}
	}

	commonPart := `
**This %[1]s is in the %[2]s stage and may change without notice.** %[3]s
the ` + "`PROVIDER_AIVEN_ENABLE_BETA`" + ` environment variable to use the %[1]s.`

	switch db.availabilityType {
	case Beta:
		builder.WriteString(fmt.Sprintf(
			commonPart,
			db.entityType.String(),
			"beta",
			"Set",
		))
	case Limited:
		builder.WriteString(fmt.Sprintf(
			commonPart,
			db.entityType.String(),
			"limited availability",
			"\nTo enable this feature, contact the [sales team](mailto:sales@aiven.io). After it's enabled, set",
		))
	}

	return builder.String()
}
```

internal/plugin/service/organization/organization_application_user_data_source.go
staceysalamon-aiven: Good point. The API docs are also being (slowly) updated/improved, so we'll get to this eventually...

Serpentiel (dismissed):
lgtm

Serpentiel (approved):
lgtm

PR #1641, opened 2024-03-07
chore(userconfig/converters): improve ip_filter safeguard

Serpentiel:

improves the `ip_filter` safeguard, adds some docs about it

## Why this way

it's simply better


SmuliS (approved):
LGTM!

PR #1644, opened 2024-03-11
fix(userconfig/converters): missing negation in ip_filter safeguard

Serpentiel:

there's a missing negation in the `ip_filter` safeguard, this PR brings it

## Why this way

it was missing 🤷


SmuliS (approved):
LGTM, and thanks for the quick fix. Glad we did a small manual test before releasing this.

Issue #1107, opened 2023-03-22
[Kafka] ip_filter_object unnecessary output


mverle50:
Topic discussed in #1088 with @Serpentiel 

I'm just wondering if it's normal behavior that when you change 1 single IP it changes the order of all those that follow ?

Like this:
Plan result to remove ip "aaa.aaa.aaa.a/23":

   ```
 ~ kafka_user_config {
            (8 unchanged attributes hidden)

          ~ ip_filter_object {
              ~ network = "aaa.aaa.aaa.a/23" -> "bbb.bbb.bbb.b/23"
            }
          ~ ip_filter_object {
              ~ network = "bbb.bbb.bbb.b/23" -> "ccc.ccc.c.c/17"
            }
          ~ ip_filter_object {
              ~ network = "ccc.ccc.c.c/17" -> "ddd.ddd.dd.d/23"
            }
          ~ ip_filter_object {
              ~ network = "ddd.ddd.dd.d/23" -> "eee.eee.ee.e/23"
            }
          ~ ip_filter_object {
              + description = "test"
              ~ network     = "eee.eee.ee.e/23" -> "fff.fff.ff.fff/32"
            }
          ~ ip_filter_object {
              - description = "test" -> null
              ~ network     = "fff.fff.ff.fff/32" -> "ggg.ggg.gg.g/23"
            }
          ~ ip_filter_object {
              ~ network = "ggg.ggg.gg.g/23" -> "hhh.hhh.hh.h/23"
            }
          ~ ip_filter_object {
              ~ network = "hhh.hhh.hh.h/23" -> "iii.iii.iii.i/24"
            }
          ~ ip_filter_object {
              ~ network = "iii.iii.iii.i/24" -> "jjj.jjj.jjj.j/18"
            }
          ~ ip_filter_object {
              ~ network = "jjj.jjj.jjj.j/18" -> "kkk.kkk.kkk.k/23"
            }
          ~ ip_filter_object {
              ~ network = "kkk.kkk.kkk.k/23" -> "lll.ll.lll.l/23"
            }
          - ip_filter_object {
              - network = "lll.ll.lll.l/23" -> null
            }

            (40 unchanged blocks hidden)
        }

        (1 unchanged block hidden)
    }

```

Serpentiel:
hey, @mverle50! 👋 

could you please let us know if this is still a valid issue?

thank you!

mverle50:
Hello @Serpentiel 
[...]

PR #1647, opened 2024-03-12
fix(ip_filter_object): use set type

### LINKED ISSUE #1638: ip_filter_objects reordering without change
## What happened?

We have multiple ip_filter_objects setup like this:
```
    ip_filter_object {
      network = "x.x.x.x/23"
      description = "a ipv4"
    }
    ip_filter_object {
      network     = "x:x:x::/48"
      description = "a ipv6"
    }
    ip_filter_object {
      network     = "x.x.x.x/16"
      description = "b"
    }
    ip_filter_object {
      network     = "x.x.x.x/16"
      description = "c"
    }
    ip_filter_object {
      network = "${var.d}/32"
      description = "d"
    }
```
Most of them are static defined as a hardcoded string. However, we also have one static ip (var.d) that is defined through another resource. Since it's a public static ip address, this one is not changing.

Without changing anything in the terraform configuration, the order if ip_filter_objects gets randomly reshuffled. The content stays the same, only the order changes.
```
          ~ ip_filter_object {
              ~ description = "d" -> "a ipv4"
              ~ network     = "x.x.x.x/32" -> "x.x.x.x/23"
            }
          ~ ip_filter_object {
              ~ description = "c" -> "a ipv6"
              ~ network     = "x.x.x.x/16" -> "x:x:x::/48"
            }
          ~ ip_filter_object {
              ~ description = "a ipv6" -> "c"
              ~ network     = "x:x:x::/48" -> "x.x.x.x/16"
            }
          ~ ip_filter_object {
              ~ description = "a ipv4" -> "d"
              ~ network     = "x.x.x.x/23" -> "x.x.x.x/32"
            }
```

And then after a while, again without touching the terraform files:
```
          ~ ip_filter_object {
              ~ description = "d" -> "a ipv4"
              ~ network     = "x.x.x.x/32" -> "x.x.x.x/23"
            }
          ~ ip_filter_object {
              ~ description = "c" -> "a ipv6"
              ~ network     = "x.x.x.x/16" -> "x:x:x::/48"
            }
          ~ ip_filter_object {
              ~ description = "a ipv6" -> "c"
              ~ network     = "x:x:x::/48" -> "x.x.x.x/16"
            }
          ~ ip_filter_object {
              ~ description = "a ipv4" -> "d"
              ~ network     = "x.x.x.x/23" -> "x.x.x.x/32"
            }
```

ip_filter_objects to stay in the same order when nothing changes

Version 4.14.0

byashimov:

Set type is the only way to fix `ip_filter_object` field.

- renames `hasValue` to `isNull` in `converters.go`, as it missleading
- uses `TypeSet` for `ip_filter_object`

Resolves #1638, #1107.

### DIFF
--- a/internal/sdkprovider/userconfig/converters/converters.go
+++ b/internal/sdkprovider/userconfig/converters/converters.go
@@ -23,7 +23,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-const userConfigSuffix = "_user_config"
+const (
+	userConfigSuffix   = "_user_config"
+	AllowIPFilterPurge = "AIVEN_ALLOW_IP_FILTER_PURGE"
+)
 
 // Expand expands schema.ResourceData into a DTO map.
 // It takes schema.Schema to know how to turn a TF item into json.
@@ -48,15 +51,16 @@ func Expand(kind string, s *schema.Schema, d *schema.ResourceData) (map[string]a
 		return nil, err
 	}
 
+	// Renames ip_filter_object/string to ip_filter
 	renameAliases(dto)
 
-	// TODO: Move to a validation part of the schema.
 	if v, ok := dto["ip_filter"].([]any); ok && len(v) == 0 {
-		if _, ok := os.LookupEnv("AIVEN_ALLOW_IP_FILTER_PURGE"); !ok {
+		if _, ok := os.LookupEnv(AllowIPFilterPurge); !ok {
 			return nil, fmt.Errorf(
-				"ip_filter list is empty, but AIVEN_ALLOW_IP_FILTER_PURGE is not set. Please set " +
-					"AIVEN_ALLOW_IP_FILTER_PURGE to confirm that you want to remove all IP filters, which is going " +
+				"ip_filter list is empty, but %[1]s is not set. Please set "+
+					"%[1]s to confirm that you want to remove all IP filters, which is going "+
 					"to block all traffic to the service",
+				AllowIPFilterPurge,
 			)
 		}
 	}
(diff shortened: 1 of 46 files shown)


byashimov:
tests
- https://github.com/aiven/terraform-provider-aiven/actions/runs/8266608999
- https://github.com/aiven/terraform-provider-aiven/actions/runs/8265421038

internal/sdkprovider/userconfig/diff/diff.go
```diff
@@ -64,3 +73,25 @@ func suppressIPFilterSet(k, old, new string, d *schema.ResourceData) bool {
...
+
+	// user_config field
+	path = path[1:]
+
+	// Drills down the field
+	for _, v := range path {
+		if v == "0" {
+			value = value.AsValueSlice()[0]
+		}
+		if value.Type().HasAttribute(v) {
```
byashimov, line 90: this must be early exit, going to fix now

internal/sdkprovider/userconfig/converters/converters.go
```diff
@@ -23,7 +23,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-const userConfigSuffix = "_user_config"
+const (
+	userConfigSuffix   = "_user_config"
+	AllowIPFilterPurge = "AIVEN_ALLOW_IP_FILTER_PURGE"
```
Serpentiel, line 28: Why do we make it an exported const? I don't know if it's a good idea if we only use it in the test, maybe the test should also define the const? WDYT?

Serpentiel (commented):
lgtm, but I have a question

internal/sdkprovider/userconfig/converters/converters.go
byashimov: `AllowIPFilterPurge` is used to format the error message too.

Serpentiel: Yes, I can see, but I mean why is it an exported const? We can make it a non-exported const instead, and introduce a separate one in the test, WDYT?

byashimov: The idea of constants is to be the source of truth.

byashimov:
@Serpentiel I have restarted the job, and that's reseted the approve. Could you please approve it again?

Serpentiel (approved):

PR #1648, opened 2024-03-12
feat(userconfig/desc): add entity and availability types

Serpentiel:

- adds entity and availability types to the description builder
- removes the `BetaDescription` helper that we used previously
- adjusts old resources/datasources for the new implementation

## Why this way

this could be useful for optimizing the resources and effort that we need to make the docs

### DIFF
--- a/internal/schemautil/userconfig/desc.go
+++ b/internal/schemautil/userconfig/desc.go
@@ -17,7 +17,7 @@ const (
 
 // String is a function that returns the string representation of the entity type.
 func (et EntityType) String() string {
-	return [...]string{"resource", "datasource"}[et]
+	return [...]string{"resource", "data source"}[et]
 }
 
 // AvailabilityType is a type that represents the availability type of an entity.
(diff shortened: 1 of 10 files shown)


internal/schemautil/userconfig/desc.go
```diff
@@ -5,23 +5,72 @@ import (
...
+const (
+	// Resource is a constant that represents the resource entity type.
+	Resource EntityType = iota
+	// DataSource is a constant that represents the data source entity type.
+	DataSource
+)
+
+// String is a function that returns the string representation of the entity type.
+func (et EntityType) String() string {
+	return [...]string{"resource", "datasource"}[et]
```
staceysalamon-aiven, line 20:
```suggestion
	return [...]string{"resource", "data source"}[et]
```

staceysalamon-aiven (approved):
Thanks so much for this! Just a very minor change to add a space to "data source".

internal/schemautil/userconfig/desc.go
Serpentiel: Done. Sorry about that, we call it `datasource` in the code :D

byashimov (approved):
lgtm

Issue #1632, opened 2024-03-05
Change to aiven_mirrormaker_replication_flow resource affects unspecified optional fields


nickgooding:
## What happened?

Updating attribute(s) for a `aiven_mirrormaker_replication_flow` resource appears to be able to affect unrelated configuration.

Example:
1. Given the following resource:
```hcl
resource "aiven_mirrormaker_replication_flow" "this" {
  project        = "example-project"
  service_name   = "mirrormaker"
  source_cluster = "source-cluster"
  target_cluster = "target-cluster"
  enable         = true

  topics = [
    "example-topic",
  ]
}
```

Aiven Console view before:
![Screenshot from 2024-03-05 14-33-58](https://github.com/aiven/terraform-provider-aiven/assets/73336146/d63665c4-648c-48cd-9e35-3d79c5e39d9b)

2. Make the following change and re-apply
```diff
--- a/main.tf
+++ b/main.tf
@@ -7,6 +7,6 @@ resource "aiven_mirrormaker_replication_flow" "this" {
 
   topics = [
    "example-topic",
+   "another-topic",
   ]
 }
```

Plan:
```
$ terraform apply
aiven_mirrormaker_replication_flow.this: Refreshing state... [id=example-project/mirrormaker/source-cluster/target-cluster]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # aiven_mirrormaker_replication_flow.this will be updated in-place
  ~ resource "aiven_mirrormaker_replication_flow" "this" {
        id                                  = "example-project/mirrormaker/source-cluster/target-cluster"
      ~ topics                              = [
            "example-topic",
          + "another-topic",
        ]
        # (11 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

aiven_mirrormaker_replication_flow.this: Modifying... [id=example-project/mirrormaker/source-cluster/target-cluster]
aiven_mirrormaker_replication_flow.this: Modifications complete after 1s [id=example-project/mirrormaker/source-cluster/target-cluster]
```

Aiven Console view after:
[...]

PR #1651, opened 2024-03-14
fix(mirrormaker_replication_flow): incorrect behavior of schema fields

Serpentiel:

Fix incorrect behavior of `aiven_mirrormaker_replication_flow` schema fields:
  - `sync_group_offsets_enabled`
  - `sync_group_offsets_interval_seconds`
  - `emit_heartbeats_enabled`
  - `emit_backward_heartbeats_enabled`
  - `offset_syncs_topic_location`
  - `replication_policy_class`

Resolves: #1632.

## Why this way

The current way it works is broken. Additionally, `offset_syncs_topic_location` and `replication_policy_class` should be required, because they can't be empty, and API then converts an empty string to their default values, `source` and `org.apache.kafka.connect.mirror.DefaultReplicationPolicy` respectively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ nav_order: 1
   prevent accidental purging of IP filters, which can lead to loss of access to services. To enable purging, set the
   environment variable to any value before running Terraform commands.
 - Use `TypeSet` for `ip_filter_object`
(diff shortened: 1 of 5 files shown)


byashimov (approved):
lgtm

byashimov (approved):
Lgtm

PR #1655, opened 2024-03-15
docs: improve networking resource documentation

staceysalamon-aiven:

- Standardizes descriptions and links to some conceptual docs
- Updates or adds example usage
- Updates import syntax


docs/data-sources/aws_privatelink.md
```diff
@@ -3,19 +3,19 @@
...
+Gets information about an AWS PrivateLink connection for an Aiven service.
 
 ## Example Usage
 
 ```terraform
-data "aiven_aws_privatelink" "foo" {
-  project      = data.aiven_project.foo.project
-  service_name = aiven_kafka.bar.service_name
+data "aiven_aws_privatelink" "main" {
+  project      = aiven_project.example_project.project
```
byashimov, line 17: If we use `data.aiven_project` everywhere I would stick to it. `data.aiven_project` is a datasource, so it is different from `aiven_project` (a resources). Datasources are not manageable, hence safe to use as read-only.

internal/sdkprovider/service/vpc/project_vpc.go
```diff
@@ -22,13 +22,13 @@ var aivenProjectVPCSchema = map[string]*schema.Schema{
 		ForceNew:    true,
 		Required:    true,
 		Type:        schema.TypeString,
-		Description: userconfig.Desc("Defines where the cloud provider and region where the service is hosted in. See the Service resource for additional information.").ForceNew().Build(),
+		Description: userconfig.Desc("The cloud provider and region where the service is hosted in the format `CLOUD_PROVIDER-REGION_NAME`. For example, `google-europe-west1` or `aws-us-east-2`. You can get this value from the service resource.").ForceNew().Build(),
```
byashimov, line 25: I think, "You can get this value from the service" is misleading. Usually, a service runs on a virtual private network, not vice-versa. So probably, the old sentence pointed to the resource description for an [additional information](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/kafka#cloud_name). Probably, we used to have "a service" resource one for all kafka, pg, etc, or it's been just copy-pasted from the API doc.

internal/sdkprovider/service/vpc/azure_vpc_peering_connection_data_source.go
```diff
@@ -13,7 +13,7 @@ import (
 func DatasourceAzureVPCPeeringConnection() *schema.Resource {
 	return &schema.Resource{
 		ReadContext: datasourceAzureVPCPeeringConnectionRead,
-		Description: "The Azure VPC Peering Connection data source provides information about the existing Aiven VPC Peering Connection.",
+		Description: "Gets information about about an AWS VPC peering connection.",
```
byashimov, line 16: It's Azure :)

byashimov (changes requested):

docs/data-sources/aws_privatelink.md
staceysalamon-aiven: Ah, okay, that makes sense. Would it be better to use data sources in general then; for example, for `aiven_kafka.example_kafka.service_name` here?

internal/sdkprovider/service/vpc/project_vpc.go
staceysalamon-aiven: Could be as there is a service endpoint. I've removed the sentence since I don't think it's helpful and we don't use this type of thing in other descriptions.

docs/data-sources/aws_privatelink.md
byashimov: Well, it just we use often project as a datasource. Services are rather managed by TF.

staceysalamon-aiven: Gotcha. I've updated here and will make sure the rest of the examples are aligned with this later. 👍

byashimov (approved):
lgtm

PR #1658, opened 2024-03-19
refactor: encapsulate user config checks

byashimov:

- replaces old converter in service integration update handler
- the check if resource has a user config is moved to the converter to encapsulate the logic further


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/8344225710)

Serpentiel (approved):
lgtm

PR #1671, opened 2024-03-22
docs: update dragonfly and add example

staceysalamon-aiven:

- Updated resource and data source descriptions
- Added example usage and import syntax
- Added example for creating a Dragonfly service


Serpentiel (approved):
lgtm

PR #1680, opened 2024-03-28
docs(service_integrations): update

staceysalamon-aiven:

Update the documentation for the service_integrations resource and data source.


Serpentiel (approved):
lgtm

internal/sdkprovider/userconfig/serviceintegration/metrics.go
```diff
@@ -39,7 +39,7 @@ func metricsUserConfig() *schema.Schema {
 							Type:        schema.TypeBool,
 						},
 						"gather_file_events_stats": {
-							Description: "gather metrics from PERFORMANCE_SCHEMA.FILE_SUMMARY_BY_EVENT_NAME.",
+							Description: "Gather metrics from PERFORMANCE_SCHEMA.FILE_SUMMARY_BY_EVENT_NAME.",
```
Serpentiel, line 42: This is generated from the schema that is returned by the API and it's not possible to change it manually because it will be overridden when the generator is ran one more time.

@byashimov could you please add the usage of [this](https://github.com/aiven/terraform-provider-aiven/blob/a40df2ae7c2e92366e0f0e656816d0b4bac687ab/internal/schemautil/userconfig/desc.go#L42) to the generator? When description was built in the previous generator, we used this to capitalize the first letter of the sentence if it's a lowercase letter.

Serpentiel (changes requested):

internal/sdkprovider/userconfig/serviceintegration/metrics.go
byashimov: https://github.com/aiven/terraform-provider-aiven/pull/1690/

Serpentiel:
let's merge after #1690

Serpentiel (approved):
дпеь

PR #1687, opened 2024-04-08
ci(incacctests): add sweeping functionality

Serpentiel:

- adds sweeping functionality to `incacctests` workflow
- some small fixes

## Why this way

to prevent dangling resources from appearing


byashimov (approved):
Lgtm

PR #1690, opened 2024-04-09
docs(userconfig): uppercase first description letter

byashimov:

Generates user config description with leading uppercase letter.

### DIFF
--- a/ucgenerator/main.go
+++ b/ucgenerator/main.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


ucgenerator/main.go
```diff
@@ -363,3 +364,14 @@ func isTypeSet(o *object) bool {
 	}
 	return false
 }
+
+// toTitleRegex finds words with a leading lowercase letter,
+// ignores dotted and underscored literals
+var toTitleRegex = regexp.MustCompile(`^[a-z]\w+[\s,$]`)
```
Serpentiel, line 370: We could've avoided the usage of `regexp` here, and all the checks, and instead do:

```go
strings.ToUpper(string(some_string[0]))[0]
```

WDYT?

byashimov: It ignores literals that look like a code snippet (that refer to a property field). Those literals have dots and underscores within, which normal words do not have.

Serpentiel (approved):
lgtm

PR #1697, opened 2024-04-19
fix(service): mark service connection info sensitive

byashimov:
Marks all connection info blocks as sensitive because terraform can't hide sensitive data within nested blocks. See https://github.com/hashicorp/terraform-plugin-sdk/issues/201


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/8752329673)

byashimov:
The whole block is hidden:

```$ terraform plan
~ kafka                   = (sensitive value)
```

PR #1700, opened 2024-04-22
chore: remove redundant service connection info

byashimov:
Removes redundant connection info fields that are have no fields and rendered as empty blocks.


Serpentiel (changes requested):
please modify `ResourceServiceCreateWrapper`, otherwise it will be failing in all of the services

Serpentiel (approved):
lgtm

PR #1714, opened 2024-04-30
feat: remove beta flag from app users and tokens

byashimov:
Removes beta flag from app users and tokens

### DIFF
--- a/internal/plugin/provider.go
+++ b/internal/plugin/provider.go
@@ -114,15 +114,13 @@ func (p *AivenProvider) Resources(context.Context) []func() resource.Resource {
 		organization.NewOrganizationResource,
 		organization.NewOrganizationUserGroupMembersResource,
 		organization.NewOrganizationGroupProjectResource,
+		organization.NewOrganizationApplicationUser,
+		organization.NewOrganizationApplicationUserToken,
 	}
 
 	// Add to a list of resources that are currently in beta.
 	if util.IsBeta() {
-		betaResources := []func() resource.Resource{
-			organization.NewOrganizationApplicationUser,
-			organization.NewOrganizationApplicationUserToken,
-		}
-
+		var betaResources []func() resource.Resource
 		resources = append(resources, betaResources...)
 	}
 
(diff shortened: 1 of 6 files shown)


internal/plugin/provider.go
```diff
@@ -114,16 +113,8 @@ func (p *AivenProvider) Resources(context.Context) []func() resource.Resource {
 		organization.NewOrganizationResource,
 		organization.NewOrganizationUserGroupMembersResource,
 		organization.NewOrganizationGroupProjectResource,
-	}
-
-	// Add to a list of resources that are currently in beta.
-	if util.IsBeta() {
-		betaResources := []func() resource.Resource{
```
Serpentiel, line 121: Please don't remove it and make it an empty slice instead. We should make it easy to add new beta resources when we need it in future.

byashimov: I would like to make it a different way instead.

Serpentiel: The problem is that the current implementation of the incremental acceptance testing solution relies on the ability to be able to statically parse the file.

The only way to make it is to generate a file that's similar to this at the end. I'm afraid, dynamic solutions like loops over description builder won't work, sadly 🙁

Serpentiel (approved):
lgtm

internal/plugin/provider.go
```diff
@@ -118,11 +118,7 @@ func (p *AivenProvider) Resources(context.Context) []func() resource.Resource {
 
 	// Add to a list of resources that are currently in beta.
 	if util.IsBeta() {
-		betaResources := []func() resource.Resource{
-			organization.NewOrganizationApplicationUser,
```
Serpentiel, line 122: Could you please also initialize them above? The test is currently failing because those resources weren't initialized.

```
=== RUN   TestAccOrganizationApplicationUserResourceDataSource
    organization_application_user_test.go:31: Step 1/5 error: Error running pre-apply plan: exit status 1
        
        Error: Invalid resource type
        
          on terraform_plugin_test.tf line 16, in resource "aiven_organization_application_user" "foo":
          16: resource "aiven_organization_application_user" "foo" {
        
        The provider hashicorp/aiven does not support resource type
        "aiven_organization_application_user".
--- FAIL: TestAccOrganizationApplicationUserResourceDataSource (5.89s)
```

Serpentiel (changes requested):

internal/plugin/provider.go
byashimov: That's why I want to refactor this behaviour.

Serpentiel (approved):
lgtm

PR #1715, opened 2024-05-06
docs: fix ClickHouse integration examples

staceysalamon-aiven:

Fix the plan name used in the ClickHouse integration examples.


rriski (changes requested):
Looks good, please also update https://github.com/aiven/terraform-provider-aiven/blob/0b0cc5e46916fa33aa988d689e44cdee049db7c9/examples/resources/aiven_clickhouse_role/resource.tf#L4

Or are you planning on going over roles docs separately?

staceysalamon-aiven:
> Looks good, please also update
> 
> https://github.com/aiven/terraform-provider-aiven/blob/0b0cc5e46916fa33aa988d689e44cdee049db7c9/examples/resources/aiven_clickhouse_role/resource.tf#L4
> 
> Or are you planning on going over roles docs separately?

Yeah, I'm working on the Hashicorp docs and the examples migration in two separate workstreams, but worth fixing these all in one go! Done now.

rriski (approved):

PR #1721, opened 2024-05-10
fix: convert ip_filter values into the target type

byashimov:

Converts `ip_filter` values into the target type.


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/9032068133)

Serpentiel (approved):
lgtm

PR #1722, opened 2024-05-10
docs(mirromaker): added valid options for config

richard-joerger-aiven:
offset_syncs_topic_location takes a string of "source" or "target" but the docs did not reflect that. The config also lacks a default and that was not captured. This adds the valid options and the lack of a default

offset_syncs_topic_location takes a string of "source" or "target" but the docs did not reflect that. The config also lacks a default and that was not captured. This adds the valid options and the lack of a default

## Why this way
Based on the other configs, this is the approach that is used to document.

### DIFF
--- a/docs/data-sources/mirrormaker_replication_flow.md
+++ b/docs/data-sources/mirrormaker_replication_flow.md
@@ -37,7 +37,7 @@ data "aiven_mirrormaker_replication_flow" "f1" {
 - `emit_heartbeats_enabled` (Boolean) Whether to emit heartbeats to the target cluster. The default value is `false`.
 - `enable` (Boolean) Enable of disable replication flows for a service.
 - `id` (String) The ID of this resource.
-- `offset_syncs_topic_location` (String) Offset syncs topic location.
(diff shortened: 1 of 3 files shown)


Serpentiel:
Hey, @richard-joerger-aiven! 👋

Thanks for your contribution. Could you please update your commit's message and PR title to align with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec?

Here's an example of how it should be like in order to be complaint with the spec:

> docs(mirrormaker): added valid options for config

Please feel free to reach out if there are any questions!

Serpentiel (changes requested):
1. Comment above.

2. Docs linter fails because the docs change is out of sync with the schema of the resource. Please mirror the proposed change to `internal/sdkprovider/service/kafka/mirrormaker_replication_flow.go`.

    - You can run `make docs` to automatically generate docs from the schema, but this is optional and you can instead keep the file that you've manually edited already.

Serpentiel (approved):
lgtm

PR #1726, opened 2024-05-14
feat: add missing connection info to various service types

Serpentiel:

Adds missing connection info to various service types.

## Why this way

It was missing.

### DIFF
--- a/internal/schemautil/service.go
+++ b/internal/schemautil/service.go
@@ -734,6 +734,9 @@ func FlattenServiceComponents(r *aiven.Service) []map[string]interface{} {
 	return components
 }
 
+// TODO: This uses an untyped map in the final resource's schema, which might be unclear to the end users.
+//
+//	We should change this in the next major version.
 func copyConnectionInfoFromAPIResponseToTerraform(
 	d *schema.ResourceData,
 	serviceType string,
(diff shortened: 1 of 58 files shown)


Serpentiel:
Tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/9080783827

internal/schemautil/service.go
```diff
@@ -742,43 +745,147 @@ func copyConnectionInfoFromAPIResponseToTerraform(
 ) error {
 	props := make(map[string]interface{})
 
+	if len(connectionInfo.Direct) > 0 {
+		props["direct"] = connectionInfo.Direct
+	}
+
+	if connectionInfo.LetsEncrypt {
+		props["letsencrypt"] = connectionInfo.LetsEncrypt
+	}
```
byashimov, line 754: 1. it won't store `false` value or ever go from `true` to `false` (if that is possible).
2. these two fields are missing from schemas

internal/schemautil/service.go
```diff
@@ -742,43 +745,147 @@ func copyConnectionInfoFromAPIResponseToTerraform(
...
 		props["receiver_remote_write_uri"] = connectionInfo.ReceiverRemoteWriteURI
 		props["store_uri"] = connectionInfo.StoreURI
+	case ServiceTypeMySQL:
+		props["uris"] = connectionInfo.MySQLURIs
+
+		params := make([]map[string]interface{}, len(connectionInfo.MySQLParams))
+		for i, p := range connectionInfo.MySQLParams {
+			port, err := strconv.ParseInt(p.Port, 10, 32)
+			if err != nil {
+				panic(err)
```
byashimov, line 834: Please, return the error.

byashimov (changes requested):
These changes require per service tests. We must prove these new fields get values, because `connection_info` is untyped map and doesn't have any documented schema.

internal/schemautil/service.go
Serpentiel: Fair enough. I'm afraid, though, the current implementation doesn't allows this with other fields, too, for example strings, and other types.

Serpentiel: I think I will remove those fields for now because more work needs to be done to figure out which services actually use them.

Serpentiel:
Tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/9092119734

Serpentiel:
> These changes require per service tests.

Sadly, automatically testing this isn't feasible. We can do manual tests, but I'm afraid implementing automatic tests for this would require enormous effort.

Serpentiel:
Tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/9092732969

Serpentiel:
Tests: https://github.com/aiven/terraform-provider-aiven/actions/runs/9101292668

internal/schemautil/service.go
```diff
@@ -743,42 +746,140 @@ func copyConnectionInfoFromAPIResponseToTerraform(
...
+
+		}
+
+		// Keep this instead of the above block.
+		params := make([]map[string]interface{}, len(connectionInfo.PostgresParams))
+		for i, p := range connectionInfo.PostgresParams {
+			port, err := strconv.ParseInt(p.Port, 10, 32)
+			if err != nil {
+				panic(err)
+			}
```
byashimov, line 793: Please return the error.

internal/sdkprovider/service/pg/pg_test.go
```diff
@@ -1025,30 +1025,62 @@ func testAccCheckAivenServicePGAttributes(n string) resource.TestCheckFunc {
...
 		}
 
-		if a["pg.0.sslmode"] != "require" {
+		if a["pg.0.params.0.sslmode"] != "require" {
 			return fmt.Errorf("expected to get a correct sslmode from Aiven")
 		}
 
-		if a["pg.0.user"] != "avnadmin" {
+		if a["pg.0.params.0.user"] != "avnadmin" {
 			return fmt.Errorf("expected to get a correct user from Aiven")
```
byashimov, line 1073: Let's keep the old validation for existing fields too, just use `||`.

internal/schemautil/service.go
```diff
@@ -743,42 +746,140 @@ func copyConnectionInfoFromAPIResponseToTerraform(
...
+				"user":          p.User,
+				"password":      p.Password,
+				"database_name": p.DatabaseName,
+			}
 		}
+
+		props["params"] = params
 		props["replica_uri"] = connectionInfo.PostgresReplicaURI
+		props["standby_uris"] = connectionInfo.PostgresStandbyURIs
+		props["syncing_uris"] = connectionInfo.PostgresSyncingURIs
```
byashimov, line 808: I can't find these two at Aiven, nor do they are mentioned in tests. Are they real?

byashimov (changes requested):

internal/schemautil/service.go
Serpentiel: `replica_uri` is only available when the service has a replica. The other ones appear conditionally, too. You can find references to these in both front end, and back end test pipelines.

byashimov (approved):

PR #1727, opened 2024-05-15
docs: add VPC error info to private_access field fields

staceysalamon-aiven:

Setting private_access for services results in an error when those services are in a project that is in a VPC. This change informs users to not set this field for these types of services.


Serpentiel (approved):
lgtm

PR #1691, opened 2024-04-09
fix(kafka_topic): send false values

byashimov:
## About this change—what it does

- Doesn't submit kafka topic config default values 
- Fix sending `false` values in kafka topic config

PR #1730, opened 2024-05-22
fix(aiven_organization_application_user_token): optional fields

byashimov:
Rewrites `aiven_organization_application_user_token` resource to SDKv2:

- replaces Number type (up to 512 bit values support) with int type
- this resource do not need null type support, hence the plugin framework makes it just complicated
- there are issues with token handler in the old client, uses generated client instead


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/9191876090)

byashimov:
There is a flaky test with deletion which will be fixed with the client update.

internal/acctest/acctest.go
```diff
@@ -135,6 +136,10 @@ const (
 	DefaultRandomSuffixLength = 10
 )
 
+func RandStr() string {
+	return acctest.RandStringFromCharSet(DefaultRandomSuffixLength, acctest.CharSetAlphaNum)
+}
+
```
rriski, line 142: Could reuse this in a few places:

* https://github.com/aiven/terraform-provider-aiven/blob/89689cce435c444126e6a6832dd32d109bbb6dbd/internal/plugin/service/organization/organization_application_user_test.go#L29
* https://github.com/aiven/terraform-provider-aiven/blob/75cde0c24279e482d3c1e45b69e9fd236ae90592/internal/plugin/service/organization/organization_group_project_test.go#L23
* https://github.com/aiven/terraform-provider-aiven/blob/b8dd02d4a97401a085c0e8a69265d7d278487033/internal/plugin/service/organization/organization_test.go#L22
* https://github.com/aiven/terraform-provider-aiven/blob/89689cce435c444126e6a6832dd32d109bbb6dbd/internal/plugin/service/organization/organization_user_group_member_test.go#L25

internal/sdkprovider/provider/provider.go
```diff
@@ -282,6 +283,12 @@ func Provider(version string) *schema.Provider {
 			return nil, diag.FromErr(err)
 		}
 
+		// fixme: temporary solution
+		err = common.CacheGenAivenClient(token, p.TerraformVersion, version)
+		if err != nil {
+			return nil, diag.FromErr(err)
+		}
```
rriski, line 290: Why do we need `CacheGenAivenClient`? Do you already have a plan to fix this `fixme` comment?

internal/schemautil/schemautil.go
```diff
@@ -40,21 +41,6 @@ func OptionalIntPointer(d *schema.ResourceData, key string) *int {
...
-	}
-
-	floatValue, ok := val.(float64)
-	if !ok {
-		return nil
-	}
-
-	return &floatValue
-}
-
```
rriski, line 57: 🔎 This cleanup was missed in https://github.com/aiven/terraform-provider-aiven/pull/1691, good to get it cleaned up now

rriski (approved):
LGTM, one question about the `CacheGenAivenClient` and the related `fixme` comment

internal/sdkprovider/provider/provider.go
byashimov: Due to the provider's interface limitations, I couldn't find a better way rather than create a Singleton. Eventually, it will take the place of the original client.

internal/acctest/acctest.go
byashimov: As a separate refactoring PR, yes. When we will have the capacity to do it.

byashimov:
@rriski rebased on main, fixed changelog conflict

rriski (approved):

PR #1746, opened 2024-06-04
feat(aiven_mirrormaker_replication_flow): add config_properties_exclude

rriski:
* Add `config_properties_exclude` attribute to `aiven_mirrormaker_replication_flow` resource
* Test `topic_blacklist` attribute in `aiven_mirrormaker_replication_flow` tests


internal/sdkprovider/service/kafka/mirrormaker_replication_flow.go
```diff
@@ -102,6 +103,15 @@ var aivenMirrorMakerReplicationFlowSchema = map[string]*schema.Schema{
 		Description:  "Offset syncs topic location. Possible values are `source` & `target`. There is no default value.",
 		ValidateFunc: validation.StringInSlice([]string{"source", "target"}, false),
 	},
+	"config_properties_exclude": {
```
keejon, line 106: Hm only issue i might see is that we have max 2048 chars string here in [API](https://github.com/aiven/aiven-core/blob/15e1e5933c0d568433d884e21baf6268d79835e4/aiven/web/schema.py#L3660). Could potentially be confusing if it fails due to this but i don't think a major issue.

keejon (dismissed):
LGTM 👍

ssuomi-aiven (approved):

docs/data-sources/mirrormaker_replication_flow.md
```diff
@@ -33,6 +33,7 @@ data "aiven_mirrormaker_replication_flow" "f1" {
 
 ### Read-Only
 
+- `config_properties_exclude` (List of String) List of topic configuration properties and/or regular expressions to not replicate. The properties that are not replicated by default are: `follower.replication.throttled.replicas`, `leader.replication.throttled.replicas`, `message.timestamp.difference.max.ms`, `message.timestamp.type`, `unclean.leader.election.enable`, and `min.insync.replicas`. Setting this overrides the defaults. For example, to exclude only 'min.insync.replicas' and 'unclean.leader.election.enable'  set this to: ["follower\\\\.replication\\\\.throttled\\\\.replicas", "leader\\\\.replication\\\\.throttled\\\\.replicas", "message\\\\.timestamp\\\\.difference\\\\.max\\\\.ms",  "message\\\\.timestamp\\\\.type"]
```
keejon, line 36: a little confusing to me. "to exclude only 'min.insync.replicas' and 'unclean.leader.election.enable" refers to exclude it from the defaults?

rriski: FYI @staceysalamon-aiven. Yes, that refers to excluding the properties from the defaults.

rriski: Updated to "For example, to enable replication for..."

ssuomi-aiven (approved):

PR #1757, opened 2024-06-11
fix(organization_user_group): make name field required

staceysalamon-aiven:

Marks the `name` field as required for the user group resource.


ssuomi-aiven:
LGTM but @rriski can you give another look before we merge?

rriski (approved):

PR #1765, opened 2024-06-17
fix(aiven_mirrormaker_replication_flow): doesn't purge config_properties_exclude

byashimov:
- fix `config_properties_exclude` purge
- turn it into TypeSet for better plan diff


byashimov:
The [tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/9593913713) will fail until #1768 is merged.

internal/schemautil/schemautil.go
```diff
@@ -382,88 +386,40 @@ func ResourceDataSet(d *schema.ResourceData, s map[string]*schema.Schema, dto an
...
-			if !ok {
-				return nil, fmt.Errorf("expected a map, got %T", v)
-			}
-
-			result[i], err = serializeMap(elem.Schema, m)
-			if err != nil {
-				return nil, err
-			}
-		}
-		return result, nil
```
byashimov, line 464: It turns out that the TF behaviour on top level fields is different compared to the nested fields. This code was copied from user config convertors. It might happen to be a case later, but for these resources TF does not accept TypeSet values.

rriski (approved):

PR #1204, opened 2023-06-12
fix(vpc): cidrs handled improperly

Serpentiel:
## About this change—what it does

- fixes improper handling of CIDRs in VPC resources
- partially reverts #1171

resolves #1186

## Why this way

misimplemented previously

PR #1781, opened 2024-07-01
fix(aiven_transit_gateway_vpc_attachment): remove peer_region deprecation

### LINKED ISSUE #1751: Persistent diff when removing the deprecated aiven_transit_gateway_vpc_attachment.peer_region field
## What happened?

To comply with the instruction that `aiven_transit_gateway_vpc_attachment.peer_region` is [deprecated](https://github.com/aiven/terraform-provider-aiven/blob/f7ea4104075c5dd31533a4469dc47f3b0a4ccd4b/internal/sdkprovider/service/vpc/transit_gateway_vpc_attachment.go#L47), we removed the field from our terraform files. However **every plan now shows the resource as changing** i.e. `peer_region = "[region]" -> null` in each plan, even after applying the change.

```hcl
resource "aiven_transit_gateway_vpc_attachment" "this" {
  for_each = local.vpc_attachments
  }

  vpc_id                  = each.value.vpc_id
  peer_cloud_account      = each.value.peer_cloud_account
  peer_vpc                = each.value.peer_vpc
  user_peer_network_cidrs = each.value.user_peer_network_cidrs
  # peer_region removed from configuration
}
```

```
  # aiven_transit_gateway_vpc_attachment.this["xxxxx"] will be updated in-place
  ~ resource "aiven_transit_gateway_vpc_attachment" "this" {
        id                      = "xxxxxxx"
      - peer_region             = "eu-west-1" -> null
    }
```

We expected that removing the deprecated field would lead to no change in terraform plan (other than removing the deprecation warning)

- Currently running `4.15.0` of the Aiven terraform provider on Linux
- The same issue occurs across all our attachments in multiple regions
- Running `terraform apply` does not resolve the planned "change", it continues to show in subsequent plans

byashimov:
The field has been mistakenly deprecated in #1204.
Marks `peer_region` as create only, removes deprecation.

Resolves #1751

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ nav_order: 1
 <!-- Always keep the following header in place: -->
 <!-- ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD -->
 
+## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
+
+- Fix `aiven_transit_gateway_vpc_attachment`: remove `peer_region` deprecation, mark the field as create only.
+
 ## [4.20.0] - 2024-06-26
 
 - Mark several sensitive user config fields as "sensitive"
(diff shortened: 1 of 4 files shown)


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/9738020167)

rriski (approved):

PR #1783, opened 2024-07-02
docs: improve documentation for application users, tokens, and groups

staceysalamon-aiven:

Adds more information on how application users and their tokens work and how they can be assigned to groups. Clarifies the roles available for projects.


byashimov (approved):
lgtm

PR #1787, opened 2024-07-12
Introduce support for Valkey

bogdanp05:
## Introduce support for Valkey

Add aiven Valkey and aiven Valkey user resources

## Why this way
Similar to the resources for Redis, with the difference that I used the generated go client for `aiven_valkey_user`
- added `aiven_valkey` resource
- added `aiven_valkey_user` resources
- updated changelog
- generated docs


byashimov (approved):
Nice job! Thanks!

PR #1801, opened 2024-08-02
Read only is a valid member type for project user

Kyrremann:

In the GUI and through the Terraform provider you can use `read only` as a valid member type, so the documentation should reflect that.

https://aiven.io/docs/platform/reference/project-member-privileges

## Why this way

It looks like the documentation is generated from this Go-file, so hopefully it will populate where it is supposed to be.


byashimov:
Hey. Thanks for the contribution! 
Unfortunately, it looks like this resources is not the only one. I have made a PR that solves the issue automatically https://github.com/aiven/terraform-provider-aiven/pull/1807/files

Kyrremann:
No worries, great catch making it automated.

PR #1807, opened 2024-08-12
fix: use code-generated enum choices

byashimov:
This way TF will get all new values automatically.


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/10348952466)

PR #1809, opened 2024-08-14
feat: mark user config sensitive fields

byashimov:
1. Uses `_secure` flag from api schemas
2. Improves regexp to find more secure fields (they are not marked in the spec yet).


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/10402041379)

rriski (approved):

PR #1814, opened 2024-08-20
fix: mark attributes of sensitive blocks as sensitive

byashimov:
Recursively mark attributes of sensitive blocks as sensitive due to an [issue in Terraform](https://github.com/hashicorp/terraform-plugin-sdk/issues/201)


byashimov:
Tests https://github.com/aiven/terraform-provider-aiven/actions/runs/10471136785

rriski (approved):

PR #1823, opened 2024-09-02
feat: use generated service get

byashimov:
Usese `ServiceGet` from generated client. Requests secrets.


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/10685993268)

byashimov:
For the history: there is another call in `WaitForServiceCreation` but it doesn't require secrets, it just waits for the service being ready.

rriski (approved):

PR #1830, opened 2024-09-05
chore: replace service get for the waiters

byashimov:
Uses `ServiceGet` from code-generated client.


internal/schemautil/wait.go
```diff
@@ -246,54 +242,50 @@ func grafanaReady(service *aiven.Service) bool {
...
+	// No backups for read replicas type of service
+	// See https://github.com/aiven/terraform-provider-aiven/pull/172
+	for _, i := range s.ServiceIntegrations {
+		if i.IntegrationType == "read_replica" && *i.DestService == s.ServiceName {
 			return true
 		}
 	}
 
-	return len(service.Backups) > 0
+	return len(s.Backups) > 0
```
byashimov, line 265: I wonder if I should add OpenSearch here because it used to have ES. But we have been running this code for years with no issues.

byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/10720191727)

rriski (approved):

PR #1834, opened 2024-09-10
refactor: unify clients configuration

byashimov:
There are multiple places where we use old client (plugin, sdkv2, tests), and multiple constructors to build it. That adds unnecessary complexity, and complicates the new client integration.

- Replaces multiple client constructors with a single approach
- Unifies old and new client configuration


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/10791218944)

PR #1838, opened 2024-09-12
feat: check user config for redacted creds

byashimov:
Checks GET response doesn't contain `<redacted>` substring.


internal/sdkprovider/service/serviceintegration/service_integration_endpoint.go
```diff
@@ -116,6 +116,12 @@ func resourceServiceIntegrationEndpointRead(ctx context.Context, d *schema.Resou
 		return schemautil.ResourceReadHandleNotFound(err, d)
 	}
 
+	// The GET with include_secrets=true must not return redacted creds
+	err = schemautil.ContainsRedactedCreds(endpoint.EndpointConfig)
+	if err != nil {
+		return err
+	}
+
```
rriski, line 124: This is good, though testing against ServiceGet response would give more confidence as that's a more common endpoint

rriski (approved):

PR #1843, opened 2024-09-13
feat: add support for topic description and ownership

roope-kar:

Add capability to set description and owner group per topic.

## Why this way

topic description and owner user group are now supported per topic in aiven API's. Both fields are optional. Especially the owner group field will provide basic governing functionality by being able to mark owner to a topic. 

Worth mentioning is that due to how the Aiven topic update endpoint works (fields are reset by sending the field as null to the endpoint) it is not possible to remove the field from config to unset it. In order to unset one of these fields, `terraform destroy` and `terraform apply` must be done in order to completely unset these fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ nav_order: 1
 - Use `ServiceGet` from the code-generated client
 - Use the code-generated client to manage `aiven_service_integration` and `aiven_service_integration_endpoint`
 - Use Go 1.23
+- Add capability to set description and owner group per topic.
 
 ## [4.24.0] - 2024-08-21
 
--- a/internal/sdkprovider/service/kafkatopic/kafka_topic.go
+++ b/internal/sdkprovider/service/kafkatopic/kafka_topic.go
@@ -186,6 +186,16 @@ var aivenKafkaTopicSchema = map[string]*schema.Schema{
 		Required:    true,
 		Description: "The replication factor for the topic.",
 	},
+	"topic_description": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The description of the topic",
+	},
+	"owner_user_group_id": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The user group that is the owner of the topic",
+	},
 	"termination_protection": {
 		Type:        schema.TypeBool,
 		Optional:    true,
@@ -308,6 +318,8 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
 	topicName := d.Get("topic_name").(string)
 	partitions := d.Get("partitions").(int)
 	replication := d.Get("replication").(int)
+	topicDescription := d.Get("topic_description").(string)
+	ownerUserGroupID := d.Get("owner_user_group_id").(string)
 
 	config, err := getKafkaTopicConfig(d)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


internal/sdkprovider/service/kafkatopic/kafka_topic.go
```diff
@@ -308,18 +318,22 @@ func resourceKafkaTopicCreate(ctx context.Context, d *schema.ResourceData, m int
...
-		TopicName:   topicName,
-		Config:      config,
-		Tags:        getTags(d),
+		Partitions:       &partitions,
+		Replication:      &replication,
+		TopicName:        topicName,
+		Config:           config,
+		Tags:             getTags(d),
+		TopicDescription: &topicDescription,
+		OwnerUserGroupId: &ownerUserGroupID,
```
byashimov, line 336: This is will send empty strings when there is no value set.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ nav_order: 1
 - Use the code-generated client to manage `aiven_service_integration` and `aiven_service_integration_endpoint`
 - Use Go 1.23
 
+## [4.25.0] - 2024-09-13
+
+- Add capability to set description and owner group per topic.
+
```
byashimov, line 25: Please just add the change to the upper list. We add the headline when we are ready to release it :)

internal/sdkprovider/service/kafkatopic/kafka_topic_test.go
```diff
@@ -173,7 +187,7 @@ data "aiven_kafka_topic" "topic" {
   topic_name   = aiven_kafka_topic.foo.topic_name
 
   depends_on = [aiven_kafka_topic.foo]
-}`, os.Getenv("AIVEN_PROJECT_NAME"), name, name)
+}`, os.Getenv("AIVEN_ORGANIZATION_NAME"), name, os.Getenv("AIVEN_PROJECT_NAME"), name, name, name)
 }
```
byashimov, line 191: Please add a new topic, because we also want keep the old behaviour.

byashimov (changes requested):

internal/sdkprovider/service/kafkatopic/kafka_topic.go
byashimov: I can fix this later, because I see similar issues there, which are not related to the PR.

roope-kar:
@byashimov Thank you for the review. I moved the changelog entry and moved the related tests into own test case.

internal/sdkprovider/service/kafkatopic/kafka_topic.go
roope-kar: I think i need to fix this to get acceptance tests pass, but i'll leave the other occasion to you then 👍🏼

internal/sdkprovider/service/kafkatopic/kafka_topic_test.go
```diff
@@ -122,6 +122,64 @@ func TestAccAivenKafkaTopic_custom_timeouts(t *testing.T) {
...
+  topic_name          = "test-acc-topic-%s"
+  topic_description   = "test-acc-topic-desc-%s"
+  owner_user_group_id = aiven_organization_user_group.foo.group_id
+  partitions          = 3
+  replication         = 2
+}
+
+
+`, os.Getenv("AIVEN_ORGANIZATION_NAME"), name, os.Getenv("AIVEN_PROJECT_NAME"), name, name, name)
+}
```
byashimov, line 181: Hey. Can you please add a new topic to an existing test? I mean spinning out a whole new kafka for a single topic is kinda too much. What do you think?

roope-kar: Ah, ok yeah makes sense. I understood wrong :)

roope-kar:
Rebased

byashimov (approved):

PR #1846, opened 2024-09-17
Rriski fix service tf import

### LINKED ISSUE #1820: Missing Attributes from Import of `avien_pg` Resource
## What happened?

We have an existing database service that has two read replicas. We are trying to import this service into terraform so that it can be managed from there from now on. The terraform definition of the replicas looks like the following.

```
resource "aiven_pg" "db_read_replica" {
  project                 = var.aiven_project
  cloud_name              = var.cloud_name
  service_name            = "${local.base_database_name}-replica"
  plan                    = var.replica_plan
  maintenance_window_dow  = "sunday"
  maintenance_window_time = "10:00:00"
  termination_protection  = true
  additional_disk_space   = var.additional_disk_space

  service_integrations {
    integration_type    = "read_replica"
    source_service_name = aiven_pg.stats_db.service_name
  }

  pg_user_config {
    service_to_fork_from = aiven_pg.stats_db.service_name
  }
}
```
While the report is successful attempting to do a plan afterwards leads to two separate issues. First the additional_disk_space attribute is not imported so that even though the existing database has addition storage it shows as an addition.

Second, and more problematic, is that the `service_integration` block causes the plan to fail with the following error. It appears that this data has not been imported and the provider thinks it needs to apply the integration as a change, which it can't do as it needs to be specified at service creation apparently.

```
│ Error: service_integrations field can only be set during creation of a service
│ 
│   with aiven_pg.db_read_replica,
│   on [replica.tf](http://replica.tf/) line 1, in resource "aiven_pg" "db_read_replica":
│    1: resource "aiven_pg" "stats_db_read_replica" {
│ 
``` 
Currently I have solved this by editing the state directly.

I expect the properties of the database and service integrations to be added to the state as part of the import so that subsequent plans work and reflect the current state of the system.

rriski:

Fix #1820. Previously, we didn't handle `additional_disk_space` and `service_integrations` properly during `terraform import`. This changeset also fixes importing `tech_emails` field of services.


internal/schemautil/service.go
```diff
@@ -651,14 +664,16 @@ func copyServicePropertiesFromAPIResponseToTerraform(
...
+		if additionalDiskSpace > 0 {
+			if err := d.Set("additional_disk_space", HumanReadableByteSize(additionalDiskSpace*units.MiB)); err != nil {
+				return err
+			}
+		} else {
+			if err := d.Set("disk_space", HumanReadableByteSize(diskSpace*units.MiB)); err != nil {
+				return err
+			}
 		}
 	}
```
byashimov, line 678: nit:
```go
if a := diskSpace - servicePlanParams.DiskSizeMBDefault; a > 0 {
    ...
} else if diskSpace > 0 {
    ...
}
```

internal/schemautil/service.go
```diff
@@ -954,15 +978,16 @@ func DatasourceServiceRead(ctx context.Context, d *schema.ResourceData, m interf
...
-		for _, emailInterface := range valuesInterface.(*schema.Set).List() {
-			results = append(results, aiven.ContactEmail{Email: emailInterface.(map[string]interface{})["email"].(string)})
+func getContactEmailListForAPI(d *schema.ResourceData) *[]aiven.ContactEmail {
+	if valuesInterface, ok := d.GetOk("tech_emails"); ok {
+		emails := valuesInterface.(*schema.Set).List()
+		results := make([]aiven.ContactEmail, len(emails))
+		for i, emailInterface := range emails {
+			results[i] = aiven.ContactEmail{Email: emailInterface.(map[string]interface{})["email"].(string)}
 		}
+		return &results
```
byashimov, line 988: nit
```
var emails []aiven.ContactEmail
err = Remarshal(valuesInterface.(*schema.Set).List(), &emails)
```

internal/schemautil/service.go
```diff
@@ -590,6 +589,20 @@ func getTechnicalEmailsForTerraform(d *schema.ResourceData, field string, s *ser
...
+	for _, integration := range integrations {
+		if integration.IntegrationType == "read_replica" {
+			integrationMap := map[string]interface{}{
+				"integration_type":    integration.IntegrationType,
+				"source_service_name": integration.SourceService,
+			}
+			readReplicaIntegrations = append(readReplicaIntegrations, integrationMap)
+		}
+	}
+	return readReplicaIntegrations, nil
```
byashimov, line 603: nit: This will swallow unexpected integration type. I would set enum validation for the `integration_type` field, and then the "if" can be removed.

rriski:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/11219607399)

byashimov (approved):

PR #1855, opened 2024-09-26
fix: add version to .goreleaser.yml

rriski:

* Adds `version` field to goreleaser config file
* Adds `yaml-language-server` as instructed in https://goreleaser.com/customization/

Fixes the following warning printed on the release GHA:

image

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,3 +1,6 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/goreleaser/goreleaser/v2.3.2/www/docs/static/schema.json
+version: 2
+
 builds:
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -1,3 +1,6 @@
+# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
```
Echie, line 1: While we're at it, should we pin the schema version as instructed here:https://goreleaser.com/customization/#pin-the-schema-version?

rriski: Done

Echie (approved):
LGTM 👍

PR #1856, opened 2024-09-30
feat: remove `aiven_thanos` from beta resources

rriski:

Promotes `aiven_thanos` to GA

## Why this way

Aiven Thanos is now GA


rriski:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/11107743460)

byashimov (approved):

PR #1867, opened 2024-10-14
feat: add disaster_recovery integration type

byashimov:
- Adds `disaster_recovery` service integration type support
- Turns `service_integrations` into type set, so the order doesn't matter for future types
- Removes custom `service_integrations` on `aiven_clickhouse` schema
- Improves documentation: mentions only supported integrations types for each service type
- Adds integration types validation


internal/schemautil/service.go
```diff
@@ -774,9 +811,9 @@ func copyServicePropertiesFromAPIResponseToTerraform(
...
 	}
 
-	// Handle read_replica service integrations
-	readReplicaIntegrations := getIntegrationsForTerraform(s.ServiceIntegrations, service.IntegrationTypeReadReplica)
-	if err := d.Set("service_integrations", readReplicaIntegrations); err != nil {
+	// Handle service integrations
+	integrations := flattenIntegrations(s.ServiceIntegrations, getBootstrapIntegrationTypes(serviceType)...)
+	if err := d.Set(serviceIntegrationsKey, integrations); err != nil {
 		return err
 	}
```
byashimov, line 818: It sets back only `read_replica` and `disaster_recovery`. Said that, we should validate `integration_type` user input [here](https://github.com/aiven/terraform-provider-aiven/pull/1867/files#diff-6531b5312c16e9c85a338320bf80d70d135f7057d0d748540534d5e76ed6bc36R83), and, basically disable `service_integrations` field for the most of the service types. Only `mysql` and `pg` types can define integrations there.

byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/12165224755)

docs/resources/clickhouse.md
```diff
@@ -142,8 +142,8 @@ Optional:
 
 Required:
 
-- `integration_type` (String) Type of the service integration. Supported integrations are `clickhouse_kafka` and `clickhouse_postgresql`.
```
byashimov, line 145: We have tests for these two as a regular service integrations. Not sure if they should be here.

internal/schemautil/service.go
byashimov: So before this change it never read things back https://github.com/aiven/terraform-provider-aiven/commit/f7a7a5ceb35bb2316950b5e7b4de20f211439faf#diff-6531b5312c16e9c85a338320bf80d70d135f7057d0d748540534d5e76ed6bc36R789
I wonder if I should keep the validation or remove it. It is probably a breaking change, but it makes 100% sense to me.

rriski (approved):

PR #1910, opened 2024-11-20
fix(dependencies): dependabot.yml

byashimov:
Fixes dependabot.yml (contains duplicated blocks).


byashimov:
My previous PR wasn't validated, because of this https://github.com/dependabot/dependabot-core/issues/4605

PR #1941, opened 2024-12-12
fix(userconfig): plan doesn't show zero values for new resources

byashimov:
Zero values are not displayed on `plan` because `d.HasChange(k)` returns `true` for new resources.
They just don't have changes.


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/12313351775)

PR #1948, opened 2024-12-17
fix: display project_vpc_id computed on plan

byashimov:
When `project_vpc_id` is not assigned, it is set automatically on the backend. The change adds the same functionality on the client side to display it in the plan.


internal/schemautil/service.go
```diff
@@ -136,10 +136,11 @@ func ServiceCommonSchema() map[string]*schema.Schema {
...
 		},
 		"project_vpc_id": {
-			Type:        schema.TypeString,
-			Optional:    true,
-			Computed:    true,
-			Description: "Specifies the VPC the service should run in. If the value is not set the service is not run inside a VPC. When set, the value should be given as a reference to set up dependencies correctly and the VPC must be in the same cloud and region as the service itself. Project can be freely moved to and from VPC after creation but doing so triggers migration to new servers so the operation can take significant amount of time to complete if the service has a lot of data.",
+			Type:             schema.TypeString,
+			Optional:         true,
+			Computed:         true,
+			Description:      "Specifies the VPC the service should run in. If the value is not set the service is not run inside a VPC. When set, the value should be given as a reference to set up dependencies correctly and the VPC must be in the same cloud and region as the service itself. Project can be freely moved to and from VPC after creation but doing so triggers migration to new servers so the operation can take significant amount of time to complete if the service has a lot of data.",
```
staceysalamon-aiven, line 142: How about: 

The ID of the the aiven_project_vpc to deploy the service in. If not set, the service is automatically deployed in a VPC when that project has only one VPC. To avoid this, set the value to `null`. You can move the service to a different VPC after creation. This migration can take some time to complete if the service has a lot of data.

staceysalamon-aiven: Looks like that ^ is not accurate based on internal discussions? So maybe it should be: 

The ID of the the aiven_project_vpc to deploy the service in. If not set, the service is automatically deployed in a VPC when that project has only one VPC. You can move the service to a different VPC after creation. This migration can take some time to complete if the service has a lot of data.

Let me know if other changes are needed. 🙏🏻

byashimov:
Using this fix instead https://github.com/aiven/terraform-provider-aiven/pull/2005

PR #1952, opened 2024-12-18
feat(kafka_quota): added support for kafka quota

vmyroslav:

- this PR adds support for `aiven_kafka_quota` resource


internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -0,0 +1,206 @@
...
+		Type:     schema.TypeInt,
+		Optional: true,
+		ForceNew: true,
+		Description: `
+Defines the bandwidth limit in bytes/sec for each group of clients sharing a quota.
+Every distinct client group is allocated a specific quota, as defined by the cluster, on a per-broker basis.
+Exceeding this limit results in client throttling.`,
+		ValidateFunc: validation.All(
+			validation.IntAtLeast(0),
+			validation.IntAtMost(1073741824),
```
byashimov, line 69: nit: `IntBetween` should do the same.

internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -0,0 +1,206 @@
...
+		ForceNew: true,
+		Description: `
+Represents a logical group of clients, assigned a unique name by the client application.
+Quotas can be applied based on user, client-id, or both.
+The most relevant quota is chosen for each connection.
+All connections within a quota group share the same quota.
+It is possible to set default quotas for each (user, client-id), user or client-id group by specifying 'default'`,
+		ValidateFunc: validation.All(
+			validation.StringLenBetween(1, 255),
+			validation.StringIsNotEmpty,
```
byashimov, line 43: Is `StringIsNotEmpty` necessary when there is `StringLenBetween(1..)` already?

examples/resources/aiven_kafka_quota/import.sh
```diff
@@ -0,0 +1 @@
+terraform import aiven_kafka_quota.example_quota PROJECT/SERVICE_NAME
```
byashimov, line 1: It should have client or user, or both. I would put all examples here.

internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -0,0 +1,217 @@
...
+		consumerByteRate  = d.Get("consumer_byte_rate").(int)
+		producerByteRate  = d.Get("producer_byte_rate").(int)
+		requestPercentage = d.Get("request_percentage").(int)
+	)
+
+	if consumerByteRate == 0 && producerByteRate == 0 && requestPercentage == 0 {
+		return fmt.Errorf("at least one quota parameter must be specified")
+	}
+
+	return nil
```
byashimov, line 216: 1. There is an attribute `AtLeastOneOf`, probably can be used with client id and the user https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/validators-predefined#sdkv2
2. Zero value can't be set, right?

internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -0,0 +1,217 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	return schemautil.ResourceDataSet(
+		aivenKafkaQuotaSchema,
+		d,
+		resp,
+		schemautil.RenameAlias("client_id", "client-id"),
```
byashimov, line 167: nit: looks like it is used everywhere, probably it is better to store it in a map like here [mirrormaker_replication_flow.go](https://github.com/aiven/terraform-provider-aiven/blob/9d58d73419cc1551640df9b51438249471ff0bc1/internal/sdkprovider/service/kafka/mirrormaker_replication_flow.go#L22-L22)

byashimov (changes requested):

examples/resources/aiven_kafka_quota/import.sh
vmyroslav: upd

byashimov (approved):

PR #1958, opened 2024-12-23
chore(release): v4.31.1

byashimov:
- Validate whether the `aiven_project.billing_group` field has changed before calling admin API


byashimov:
This change can't be really tested, as it requires running the client with less privileged token.

byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/12465458818)

vmyroslav (approved):

PR #1986, opened 2025-01-10
refactor: ResourceDataSet should not take the schema

byashimov:
Resources like service users can't share the code, when the schema is mandatory.


byashimov:
Terraform panics on setting invalid fields. Schema is mandatory.

PR #2005, opened 2025-01-21
fix: services always created in the default VPC

byashimov:
Fix: services always created in the default VPC of the same cloud, 
when `project_vpc_id` wasn't set.

1. Uses [updated client](https://github.com/aiven/go-client-codegen/pull/227) that can send `nil` value in `project_vpc_id` (the client updated in this [PR](https://github.com/aiven/terraform-provider-aiven/pull/2009))
1. Adds a test
 
**Doesn't fix:** resetting `project_vpc_id` was never possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ nav_order: 1
 
 - Add `disaster_recovery` service integration type support
 - Add `aiven_flink_jar_application`, `aiven_flink_jar_application_version` and `aiven_flink_jar_application_deployment` BETA resources
+- Fix: services always created in the default VPC of the same cloud, when `project_vpc_id` wasn't set
 - Change `aiven_account_team_project` field `team_type` (enum): add `organization:networking:read`, `organization:networking:write`
 - Change `aiven_organization_permission` resource field `permissions.permissions` (enum):
   add `organization:networking:read`, `organization:networking:write`
--- a/internal/sdkprovider/service/grafana/grafana_test.go
+++ b/internal/sdkprovider/service/grafana/grafana_test.go
@@ -1,6 +1,7 @@
 package grafana_test
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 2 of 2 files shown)


byashimov:
Must be tested thoroughly. I can think of these cases:

### Case 1: migrating from a non-bug version
1. Use [v4.26.0](https://github.com/aiven/terraform-provider-aiven/blob/v4.26.0/internal/schemautil/service.go#L419) (doesn't have the bug)
2. Create a service in VPC
3. Create a service without VPC
4. Checkout on the branch, build it, run `terraform apply`
5. Both services must retain their VPC ids

### Case 2: migrating from the bugged version

1. Use `v4.33.0` (the latest stable at the moment)
2. Create a VPC
3. THE BUG: Create a service in the same cloud but _without_ vpc id set
5. Make sure the service is in the VPC from step 2 (prove the bug)
6. Checkout on the branch, build it, run `terraform apply`
7. The service hasn't changed

byashimov:
The [test](https://github.com/aiven/terraform-provider-aiven/actions/runs/12895768230/job/35957416306#step:6:37).

CHANGELOG.md
```diff
@@ -10,6 +10,17 @@ nav_order: 1
...
+## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
+
+- Fix: services always created in the default VPC of the same cloud, when `project_vpc_id` wasn't set
+- Change `aiven_account_team_project` resource and datasource field `team_type` (enum):
+  add `organization:networking:read`, `organization:networking:write`
+- Change `aiven_kafka_native_acl` resource field `resource_type` (enum)
+- Change `aiven_organization_permission` resource field `permissions.permissions` (enum):
+  add `organization:networking:read`, `organization:networking:write`
+- Change `aiven_project_user` resource and datasource field `member_type` (enum): add `organization:networking:read`, `organization:networking:write`
+- Change `aiven_service_integration_endpoint` resource and datasource field `endpoint_type` (enum): add `external_azure_blob_storage`
```
byashimov, line 22: Turns out there is a validation. So technically, even `external_azure_blob_storage` has released its user config, user can't set the `endpoint_type` to is it, because it wasn't in the OpenAPI spec (not on the choices list).

byashimov:
The client was merged in this [PR](https://github.com/aiven/terraform-provider-aiven/pull/2009).

internal/sdkprovider/service/grafana/grafana_test.go
```diff
@@ -695,3 +696,81 @@ resource "aiven_grafana" "bar" {
 }
 `, networks, os.Getenv("AIVEN_PROJECT_NAME"), name)
 }
+
+// TestAccAiven_grafana_and_default_vpc
+// Tests that when `project_vpc_id` is not set, the service is NOT created in the default VPC.
+// The backend automatically creates a service in the default VPC,
+// so terraform must send nil in `project_vpc_id`.
+// Tests all service types because the logic is in the shared code, but grafana runs faster.
```
rriski, line 704: nit: I'm having a hard time following this comment

suggestion: 


```suggestion
// Verifies that when `project_vpc_id` is not set, the service is NOT created in the default VPC.
// The backend automatically assigns a service to the default VPC if `project_vpc_id` is not specified,
// so terraform must ensure `project_vpc_id` is explicitly set to nil.
// Although this behavior affects all service types due to shared backend logic, we use Grafana for testing as it has the fastest setup time.
```

rriski (approved):

rriski (approved):

PR #2016, opened 2025-01-27
fix: enable read_replica for various service types

byashimov:
Enable `read_replica` service integration for various service types.


byashimov:
Wrong branch name.

PR #2024, opened 2025-01-28
ci: remove caching action

byashimov:
Fixes `Cannot open: File exists` error in [acceptance tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/13000813509/job/36258930620#step:5:27).

Caching action is already used by [`actions/setup.go`](https://github.com/actions/setup-go/blob/5a083d0e9a84784eb32078397cf5459adecb4c40/package.json#L28).
So we don't need to use [`actions/cache`](https://github.com/orgs/community/discussions/58174#discussioncomment-7627691).

Additionally, sets `ACC_TEST_PARALLELISM=5` to reduce concurrency errors.

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -35,6 +35,8 @@ jobs:
   acceptance_tests:
     needs: setup_aiven_project_suffix
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


byashimov:
tests https://github.com/aiven/terraform-provider-aiven/actions/runs/13016506327/job/36306942147

rriski (approved):

PR #2034, opened 2025-02-05
refactor(project|billing_group): replaced client-v2 with avngen

vmyroslav:

- replaced client-v2 with avngen in `project` and `billing_group` resources


byashimov (changes requested):

vmyroslav:
These resources are already deprecated. I'll close this PR for now.

PR #2038, opened 2025-02-07
refactor: validate ForceNew fields in ResourceDataSet

byashimov:
Terraform doesn't read the config on `import` command.
If a `ForceNew` field is not set during the `ReadContext`,
terraform then compares "empty" value with the value from the config,
and outputs "replace" plan.

`ResourceDataSet` must set all the `ForceNew` field values.

- Adds the validation
- Improves types conversions
- Makes `ResourceDataSet` more similar to `ResourceDataGet`

### DIFF
--- a/internal/schemautil/schemautil.go
+++ b/internal/schemautil/schemautil.go
@@ -369,7 +369,7 @@ func ResourceDataGet(d ResourceData, dto any, fns ...MapModifier) error {
 	}
 
 	for _, f := range fns {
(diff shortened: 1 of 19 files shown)


byashimov:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/13219374929)

byashimov:
I'll be looking for another solution. This one can't be really generated: the values must be set manually.

PR #2044, opened 2025-02-11
fix: dissociate static ips upon service deletion

byashimov:
Static IPs are not dissociated upon service deletion on the BE in some cases.


byashimov:
This doesn't work. It is impossible to unassign IP of a running service.

PR #2070, opened 2025-02-25
fix(organization_project): added sweeper

vmyroslav:
- added sweeper for `organization_project` resource


rriski (dismissed):

.github/workflows/examples-tests.yml
```diff
@@ -50,6 +50,11 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.11"
```
rriski, line 56: Do we need to use a specific version or should we omit this to use the latest version?

rriski (dismissed):

rriski (approved):

PR #2072, opened 2025-02-26
feat: add governance_access_resource [GOV-630]

roope-kar:
for managing governance subscriptions, marked as beta as the API and resource are prone to change.

Resolves: #xxxxx.


internal/sdkprovider/service/governance/governance_subscription.go
```diff
@@ -0,0 +1,256 @@
...
+		},
+		Timeouts: schemautil.DefaultResourceTimeouts(),
+		Schema:   aivenGovernanceSubscriptionSchema,
+	}
+}
+
+func resourceGovernanceSubscriptionCreate(ctx context.Context, d *schema.ResourceData, client avngen.Client) error {
+	var req governance.OrganizationGovernanceSubscriptionCreateIn
+
+	req.SubscriptionName = d.Get("subscription_name").(string)
```
roope-kar, line 145: The reason why I did not use the preferred schemautil.ResourceDataGet and schemautil.ResourceDataSet was because in the schema, the subscription_data is a list and API wants single object. There we also some optional fields that needed to be excluded if not defined.

internal/sdkprovider/service/governance/governance_subscription.go
```diff
@@ -0,0 +1,256 @@
...
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice(governance.PermissionTypeChoices(), false),
+				Description:  userconfig.Desc("Specifies whether the action is explicitly allowed or denied for the service user on the specified resource").ForceNew().PossibleValuesString(governance.PermissionTypeChoices()...).Build(),
+			},
+			"host": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(1, 256),
+				Description:  userconfig.Desc("The IP address from which a principal is allowed or denied access to the resource. Use `*` for all hosts").ForceNew().MaxLen(256).Build(),
```
byashimov, line 81: nit: there is `IsCIDR` validator [organization_vpc.go](https://github.com/aiven/terraform-provider-aiven/blob/main/internal/sdkprovider/service/vpc/organization_vpc.go#L36-L36)

internal/sdkprovider/service/governance/governance_subscription.go
byashimov: Hey. It is better to stick to the schema. I would flatten `subscription_data` field. Is there any reason why it is turned into a property?

internal/sdkprovider/service/governance/governance_subscription.go
```diff
@@ -0,0 +1,256 @@
...
+		acls[i]["pattern_type"] = string(acl.PatternType)
+		acls[i]["operation"] = string(acl.Operation)
+		acls[i]["permission_type"] = string(acl.PermissionType)
+		acls[i]["host"] = acl.Host
+	}
+	subscriptionData[0]["acls"] = acls
+
+	d.SetId(subscription.SubscriptionId)
+
+	return d.Set("subscription_data", subscriptionData)
```
byashimov, line 202: Create operation should return "read" handler, so it can prove that the state it reads is equal to local.

internal/sdkprovider/service/governance/governance_subscription.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func resourceGovernanceSubscriptionRead(ctx context.Context, d *schema.ResourceData, client avngen.Client) error {
+	subscription, err := client.OrganizationGovernanceSubscriptionGet(
+		ctx,
+		d.Get("organization_id").(string),
+		d.Id(),
```
byashimov, line 209: Likely, the `id` is a composite field: `org_id/sub_id`, because that's what you need to read the resource on `terraform import`.
The subscription id should be stored in a separate field, like `subscription_id`.

byashimov (changes requested):
Please update the go mod files if you use a custom client branch.

internal/sdkprovider/service/governance/governance_subscription.go
roope-kar: Agreed, I wanted to keep it 1:1 with the schema as well thus I did not flatten it. The subscription_data is a property because the content schema could be different for other subscription_type (Let's say we implement this for PostgreSQL and come up with different type)

Though one option could be to make the resource based on `subscription_type`. In this case, aiven_governance_kafka_subscription. While it would not be 1:1 with API schema, it would provide more flat structure and perhaps reduce congitive complexity.

internal/sdkprovider/service/governance/governance_subscription.go
roope-kar: No problem, your suggestion sounds more accurate 👍🏼 

I reasoned it so that since the create already returns the subscription we would save an API call by not calling the read API.

internal/sdkprovider/service/governance/governance_subscription.go
roope-kar: Thanks!

internal/sdkprovider/service/governance/governance_subscription.go
roope-kar: Regarding the API schema, I think there is room for improvement and we'll probably try to simplify it with some abstraction but I'm hoping the Beta mark will give us room for future improvements.

internal/sdkprovider/service/governance/governance_subscription.go
roope-kar: Hey, I did not use this for now, since I realized it needs to support csv of ips (which matches the kafka-native-acl schema)

roope-kar:
Everything should be up to date. We are waiting for docs to catch up to pair up a release with console.

byashimov (changes requested):
Hey! Please add your module here [acceptance-tests.yml](https://github.com/aiven/terraform-provider-aiven/blob/541475a3025ea23330a4ff58c95a997e80c612b4/.github/workflows/acceptance-tests.yml#L45-L45) and I'll run the tests. Thanks!

byashimov (approved):

PR #2090, opened 2025-03-10
feat(organization): add new fields

### LINKED ISSUE #2086: Be able to list billing groups or find out which billing group is associated with an organization
## What is currently missing?

Currently the only way to find out which billing group is associated with an organization is to use the deprecated `aiven_account` data source.

Either provide the billing group as an attribute of the organization data source or be able to list all billing groups.

byashimov:
- Adds `parent_account_id` and `primary_billing_group_id` fields

Resolves #2086


rsalevsky:
The fields added here are not part of the new endpoints, the account endpoints used here are or will be deprecated soon.

byashimov:
There will be new resources to cover the case. Closing this PR.

PR #2102, opened 2025-03-12
fix(aiven_azure_org_vpc_peering_connection): added new state check to peering connections

vmyroslav:
- added new state check to peering connection resources


internal/sdkprovider/service/vpc/vpc_peering_connection_state.go
```diff
@@ -28,7 +28,7 @@ func getDiagnosticsFromState(pc peeringConnectionState) diag.Diagnostics {
 					"completes the setup in their cloud account. The steps needed in the user cloud "+
 					"account depend on the used cloud provider. Find more in the state info: %s",
 					stateInfoToString(pc.GetStateInfo()))}}
-		case "DELETED":
+		case "DELETED", "DELETING":
```
byashimov, line 31: nit: I would replace literals with the constants from the client https://github.com/aiven/go-client-codegen/blob/main/handler/organizationvpc/organizationvpc.go
That would help to:
1. prove the values are valid, and do not have typos
2. verify breaking changes

byashimov, line 31: nit2: `if pc.GetState() != "ACTIVE" {` can be an empty case.

byashimov (changes requested):
Please replace literals with the enum values. Thanks.

internal/sdkprovider/service/vpc/vpc_peering_connection_state.go
vmyroslav: We use strings for the state all over these resources.
1. Changing it here wouldn't ,make any difference
2. This string also works with the types coming from the old client. Comparing them would anyway require converting them to string.

rriski: > We use strings for the state all over these resources.

Which resources are you referring to? IMO we should migrate to codegen client enums where possible. The states may change on the backend and using codegen client makes it clearer where the "magic" string comes from and what it relates to.

> This string also works with the types coming from the old client. Comparing them would anyway require converting them to string.

Are we using both clients here? If so, as the old one is deprecated, shouldn't we migrate to the codegen client? That would make this point moot. Additionally, we seem to be missing a few connection states here, namely "APPROVED", "APPROVED_PEER_REQUESTED" and "ERROR".

vmyroslav: This code is a copy from old peering connections which use the old client

byashimov (approved):

PR #2105, opened 2025-03-14
docs: update migration guide and get started

staceysalamon-aiven:

Fixes the code for the `principal_id`. Updates the get started example to use the new resource and also fixes outdated UI labels. Makes some other minor improvements to the example.

### DIFF
--- a/docs/guides/update-deprecated-resources.md
+++ b/docs/guides/update-deprecated-resources.md
@@ -84,9 +84,6 @@ To replace resources that are deprecated:
 The `aiven_project_user` and `aiven_organization_group_project` resources have been replaced by
 [the `aiven_organization_permission` resource](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/organization_permission).
 
-   ~> **Important**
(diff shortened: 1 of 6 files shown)


docs/guides/update-deprecated-resources.md
```diff
@@ -160,6 +156,10 @@ resource "aiven_organization_group_project" "example" {
       }
       ```
 
+      ~> **Important**
+      Do not use the `aiven_project_user` or  `aiven_organization_group_project` resource with `aiven_organization_permission`.
```
byashimov, line 160: Can we make it clear that this might lead to unexpected consequences? I'm not sure of the best wording, but something like this might work: These all control permissions at the organization level and may produce unexpected results when used together.

staceysalamon-aiven: What about: 

> Do not use the `aiven_project_user` or  `aiven_organization_group_project` resource with `aiven_organization_permission`. All of these resources manage organization-level permissions and using them together can cause conflicts and unexpected behavior.

byashimov: Yeah, that is much better, thank you :)

byashimov (approved):

PR #2109, opened 2025-03-17
docs: add guide and example for autoscaler

staceysalamon-aiven:

Updates the `additional_disk_storage` description for the autoscaler case to avoid confusion. Adds an example and guide for setting up disk autoscaler.


docs/guides/disk-autoscaler.md
```diff
@@ -0,0 +1,81 @@
...
+
+    ```bash
+    terraform apply --auto-approve
+     ```
+
+## Disable disk autoscaling
+
+To disable the disk autoscaler without destroying the endpoint or service:
+
+1. Remove the `aiven_service_integration` for the autoscaler. Keep the `aiven_service_integration_endpoint` for the autoscaler.
```
byashimov, line 69: Hey. Probably having an endpoint without an integration doesn't make much sense.

staceysalamon-aiven: My inital thought was that someone might want to use the endpoint for a different service, but this doesn't really warrant having separate instructions for anyway. 👍

docs/guides/disk-autoscaler.md
```diff
@@ -0,0 +1,81 @@
...
+your service with it using the `aiven_service_integration` resource.
+
+## Enable disk autoscaler when creating a service
+
+Follow [the disk autoscaler example](https://github.com/aiven/terraform-provider-aiven/tree/main/examples/autoscaler_integration)
+to learn how to to create an Aiven for PostgreSQL® service, autoscaler endpoint, and enable the autoscaler integration on the service.
+
+## Enable disk autoscaler on an existing service
+
+1. If the service has `additional_disk_space` configured, remove the field. The additional storage will be adjusted by the autoscaler.
```
byashimov, line 21: The autoscaler must be created before the field removal. So after step 5.

byashimov (changes requested):

templates/guides/disk-autoscaler.md
```diff
@@ -0,0 +1,63 @@
...
+    terraform plan
+    ```
+
+4. To create the endpoint and enable the autoscaler, apply the configuration by running:
+
+    ```bash
+    terraform apply --auto-approve
+     ```
+
+5. If the service has `additional_disk_space` configured, remove the field. The additional storage will be adjusted by the autoscaler.
```
byashimov, line 63: Hey. Could you add something like "when the field is removed, the plan output will be empty"? I just want to make it clear how this should work.
If the scaler isn’t enabled, Terraform will still allow the field to be removed, which could lead to a disaster, but the plan won't be empty. Thanks!

staceysalamon-aiven: Good idea! I added a note about this, let me know what you think of the wording.

byashimov (approved):

PR #2110, opened 2025-03-17
fix(ci): fix acceptance tests

vmyroslav:
### Problem
Our CI pipeline was [encountering](https://github.com/aiven/terraform-provider-aiven/actions/runs/13888765905/job/38857150882#step:5:43) intermittent "text file busy" errors during Terraform provider acceptance tests.

### Analysis
The testing framework (terraform-plugin-testing) was installing a separate Terraform binary for each test process. When running in [parallel](https://github.com/aiven/terraform-provider-aiven/blob/v4.37.0/.github/workflows/acceptance-tests.yml#L39), we encountered race conditions where processes attempted to execute binaries before filesystem operations were fully complete.

### Root Cause
During the initialization phase, the testing framework [decides](https://github.com/hashicorp/hc-install/blob/main/installer.go#L36) which Terraform binary to use for tests. By default, it uses the [latest version installer](https://github.com/hashicorp/hc-install/blob/main/releases/latest_version.go#L82), which downloads, extracts, and immediately attempts to execute the binary.
This creates a race condition between filesystem write operations and execution attempts, leading to the "text file busy" error - most likely leading to [this original issue](https://github.com/golang/go/issues/22315).

### Solution
By specifying the environment variable [TF_ACC_TERRAFORM_PATH](https://github.com/hashicorp/terraform-plugin-testing/blob/769b589e946c50cb619be92652c4f916368f1818/internal/plugintest/environment_variables.go#L110), we force the terraform-plugin-testing framework to use a pre-installed binary across all test processes, completely avoiding re-installation of terraform binary.

### Implementation:
- Pre-install Terraform once during the setup phase
- Export the binary path to all test jobs
- Set TF_ACC_TERRAFORM_PATH for all test runs

### Benefits

- Eliminated "text file busy" errors completely
- We can increase the parallelism to speed up the pipeline a little

### Links
The pipelines without the retry, but with installed TF: [1](https://github.com/aiven/terraform-provider-aiven/actions/runs/13917067399), [2](https://github.com/aiven/terraform-provider-aiven/actions/runs/13910321724), [3](https://github.com/aiven/terraform-provider-aiven/actions/runs/13919642121).


byashimov:
I would remove the retries to let it clearly fail. With retries we will have to go through every run and see if it has retried builds. It would be easier to notice failed builds instead.

vmyroslav:
Removed retries

PR #2120, opened 2025-03-21
feat: org billing resource & datasource

rriski:

Adds org billing resource & datasource to TF


rriski:
[tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/14169928431)

PR #2121, opened 2025-03-24
chore: refactor acctest env var handling

rriski:

* Unify and simplify env var handling in acceptance tests
* Fix semgrep in CI (I can separate this out to another PR if wanted)

Example semgrep error (https://github.com/aiven/terraform-provider-aiven/actions/runs/14044395840/job/39321731411?pr=2121):

Capture

### DIFF
--- a/internal/sdkprovider/service/kafka/kafka_connector_test.go
+++ b/internal/sdkprovider/service/kafka/kafka_connector_test.go
@@ -126,7 +126,6 @@ func testAccCheckAivenKafkaConnectorResourceDestroy(s *terraform.State) error {
 	return nil
 }
 
-// nosemgrep: kafka connectors need kafka with business plans
 func testAccKafkaConnectorResource(name string) string {
 	return fmt.Sprintf(`
 data "aiven_project" "foo" {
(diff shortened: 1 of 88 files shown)


internal/sdkprovider/service/kafka/kafka_connector_test.go
```diff
@@ -265,7 +265,7 @@ data "aiven_project" "foo" {
 resource "aiven_kafka" "bar" {
   project                 = data.aiven_project.foo.project
   cloud_name              = "google-europe-west1"
-  plan                    = "startup-2"
+  plan                    = "business-4"
```
byashimov, line 268: This used to work on a cheaper plan. Or does it fail with the new checks?

rriski: Hmm right, I'm not sure how this worked before, as the resource specified `kafka_connect = true` which is not supported in startup-2. Even the old comment says: "nosemgrep: kafka connectors need kafka with business plans"

vmyroslav (approved):

PR #2133, opened 2025-04-04
refactor(billing_group): generate resource files

byashimov:
- Uses generated schemas, models, and marshallers.
- Converts email fields into lists of objects, 
  aligning with the API to prevent future issues from schema changes.

Some comments.

1. Generated files use the `zz_` prefix, similar to Kubebuilder for Kubernetes,
   this pushes them down in file listings and makes them easier to identify visually.
2. `Expand` and `Flatten` naming is used by TF developers, 
   likely to avoid conflicts where `Set/Get` clashes with the `Set` type,
   so you don't have to `SetSet` or `GetSet`.
3. "Flatten" functions use unsafe casting. 
   The values come from the struct—they must be safe to use.

Resolves NEX-1352.


byashimov:
Doesn't update the changelog, because the resource is not released yet.

byashimov:
We have released emails as lists of strings. This PR is obsolete.

PR #2136, opened 2025-04-07
chore(release): v4.38.0

byashimov:
- Replaces type List with type Set in organization package (better UX)
- Disables address test that resets `company_name` field until the BE is fixed
- Adds tests discovery
- Fixes the sweepers
- Fixes the tests
- Fixes the updated client type issues (the API schema has changed)


.github/workflows/acceptance-tests.yml
```diff
@@ -32,42 +32,29 @@ jobs:
...
+
+  test:
+    needs: [setup_aiven_project_suffix, find_tests]
     runs-on: ubuntu-latest
     env:
       ACC_TEST_PARALLELISM: 5
     strategy:
       max-parallel: 5
-      fail-fast: false
+      fail-fast: true
```
vmyroslav, line 55: I would prefer false as default here and inputs for manual workflow

.github/workflows/acceptance-tests.yml
```diff
@@ -32,42 +32,29 @@ jobs:
...
-          project,
-          redis,
-          servicecomponent,
-          serviceintegration,
-          staticip,
-          thanos,
-          valkey,
-          vpc,
-        ]
+        pkg_path: ${{ fromJson(needs.find_tests.outputs.matrix) }}
```
vmyroslav, line 57: nit: would be nice to have a package name in the matrix name instead of `test ./internal/...`

Issue #2132, opened 2025-04-04
aiven_clickhouse_database import is partial


jean-mercier-hivebrite:
## What happened?

When i try to import an aiven_clickhouse_database
i'm using the latest version of the provider "4.37.0"

```
import {
  id = "myproject/clickhouse/db_name"
  to = aiven_clickhouse_database.clickhouse_db
}

resource "aiven_clickhouse_database" "clickhouse_db" {
  project             = "myproject"
  service_name = "clickhouse"
  name               = "db_name"
}

```

When i do a plan i got this output

```
# aiven_clickhouse_database.clickhouse_db must be replaced
# (imported from "myproject/clickhouse/db_name")
# Warning: this will destroy the imported resource
-/+ resource "aiven_clickhouse_database" "clickhouse_db" {
 ~ id                                        = "myproject/clickhouse/db_name" -> (known after apply)
     name                                 = "db_name"
    + project                            = "myproject" # forces replacement
    + service_name                = "clickhouse" # forces replacement
    + termination_protection = false
 }
```

So it's show that the ressource looks imported but only partially as it's want to add the project and service name forcing the replacement

## What did you expect to happen?

I expect to see no change after the import

byashimov:
Hey @jean-mercier-hivebrite!
Thank you for the report. I'll come up with a fix asap.

PR #2138, opened 2025-04-07
fix(clickhouse_grant): import

byashimov:
- Adds missing importer
- Removes Clickhouse plan validator to use a cheaper plan for the test

Resolves NEX-1356, #2134

PR #2139, opened 2025-04-07
fix(clickhouse_database): import

byashimov:
- Sets required fields on import
- Removes the default value for `termination_protection` default value,
  which caused unnecessary plan diffs
    
Resolves NEX-1357, #2132

### DIFF
--- a/internal/sdkprovider/service/clickhouse/clickhouse_database.go
+++ b/internal/sdkprovider/service/clickhouse/clickhouse_database.go
@@ -23,7 +23,6 @@ var aivenClickhouseDatabaseSchema = map[string]*schema.Schema{
 	"termination_protection": {
 		Type:        schema.TypeBool,
 		Optional:    true,
-		Default:     false,
 		Description: userconfig.Desc(`Client-side deletion protection that prevents the ClickHouse database from being deleted by Terraform. Enable this for production databases containing critical data.`).DefaultValue(false).Build(),
 	},
 }
@@ -77,6 +76,12 @@ func resourceClickhouseDatabaseRead(ctx context.Context, d *schema.ResourceData,
 		return diag.FromErr(schemautil.ResourceReadHandleNotFound(err, d))
 	}
 
+	if err := d.Set("project", projectName); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("service_name", serviceName); err != nil {
+		return diag.FromErr(err)
+	}
 	if err := d.Set("name", database.Name); err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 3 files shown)


byashimov:
Must be merged after https://github.com/aiven/terraform-provider-aiven/pull/2138

byashimov:
Tests https://github.com/aiven/terraform-provider-aiven/actions/runs/14320941482

internal/sdkprovider/service/clickhouse/clickhouse_grant.go
```diff
@@ -107,6 +107,9 @@ func ResourceClickhouseGrant() *schema.Resource {
 		DeleteContext: resourceClickhouseGrantDelete,
 		Schema:        aivenClickhouseGrantSchema,
 		Timeouts:      schemautil.DefaultResourceTimeouts(),
+		Importer: &schema.ResourceImporter{
```
vmyroslav, line 110: Looks like it has some conflicts with [another](https://github.com/aiven/terraform-provider-aiven/pull/2138/files#diff-cbe84f9ccb8893bfe143f754cae1383fc3c26af3e14eac485e6453a7437d7cb7R110) PR which is related to the grant.

internal/sdkprovider/service/clickhouse/clickhouse_database.go
```diff
@@ -77,6 +76,12 @@ func resourceClickhouseDatabaseRead(ctx context.Context, d *schema.ResourceData,
 		return diag.FromErr(schemautil.ResourceReadHandleNotFound(err, d))
 	}
 
+	if err := d.Set("project", projectName); err != nil {
```
vmyroslav, line 79: Should we add `termination_protection` field as well?

byashimov: It is a virtual field. It doesn't exist in the API. Really confusing. I have also removed the default value, which false, as it does change the state, while it makes no sense to have false default value for a bool field.

jean-mercier-hivebrite: Yes was confusing me when i import and it's import with a false value even i set true in the destination ressource. but it's not a blocker as the import block do the change, but it's a confusing drift

vmyroslav (approved):

internal/sdkprovider/service/clickhouse/clickhouse_database.go
vmyroslav: @jean-mercier-hivebrite If you set the `termination_protection` in the destination explicitly then the state should be valid. Please let us know if you have any issues with it.

jean-mercier-hivebrite: i'm use to use the import block it's make the import and the plan at once 

`import {
  id = "${var.aiven_project}/${var.aiven_clickhouse_service_name}/db_name"
  to = aiven_clickhouse_database.db_name
}`

And in this case with the previous behaviour you have import and change to put terminaison protection to "true"

you have the same problem on kafka_topic import by the way

jean-mercier-hivebrite: but it's not a blocker

vmyroslav (approved):

PR #2142, opened 2025-04-09
docs: add Thanos examples

staceysalamon-aiven:

* Adds a simple Thanos example for use in the getting started for this service.
* Replaces the M3 integration example with one using Thanos.
* Adds another integration example that replicates instructions in the Aiven Docs for the Console.


examples/kafka_pg_grafana/variables.tf
```diff
@@ -0,0 +1,27 @@
+variable "aiven_token" {
+  description = "Aiven token"
+  type        = string
```
vmyroslav, line 3:
```suggestion
  type        = string
  sensitive   = true
```

examples/thanos/variables.tf
```diff
@@ -0,0 +1,15 @@
+variable "aiven_token" {
+  description = "Aiven token"
+  type        = string
```
vmyroslav, line 3:
```suggestion
  type        = string
  sensitive   = true
```

vmyroslav (approved):

PR #2143, opened 2025-04-09
docs: add alloydb example for getting started

staceysalamon-aiven:


examples/alloydbomni/variables.tf
```diff
@@ -0,0 +1,15 @@
+variable "aiven_token" {
+  description = "Aiven token"
+  type        = string
```
vmyroslav, line 3:
```suggestion
  type        = string
  sensitive = true
```

vmyroslav (approved):

PR #2146, opened 2025-04-11
docs: update opensearch documentation

staceysalamon-aiven:

Updates docs for OpenSearch resources and data sources.

### DIFF
--- a/internal/sdkprovider/service/opensearch/opensearch.go
+++ b/internal/sdkprovider/service/opensearch/opensearch.go
@@ -12,7 +12,7 @@ func opensearchSchema() map[string]*schema.Schema {
 	s[schemautil.ServiceTypeOpenSearch] = &schema.Schema{
 		Type:        schema.TypeList,
 		Computed:    true,
-		Description: "OpenSearch server provided values",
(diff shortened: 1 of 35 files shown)


examples/alloydbomni/variables.tf
```diff
@@ -0,0 +1,15 @@
+variable "aiven_token" {
+  description = "Aiven token"
+  type        = string
```
vmyroslav, line 3:
```suggestion
  type        = string
  sensitive   = true
```

internal/sdkprovider/service/opensearch/opensearch.go
```diff
@@ -12,7 +12,7 @@ func opensearchSchema() map[string]*schema.Schema {
 	s[schemautil.ServiceTypeOpenSearch] = &schema.Schema{
 		Type:        schema.TypeList,
 		Computed:    true,
-		Description: "OpenSearch server provided values",
+		Description: "Vaues provided by the OpenSearch server.",
```
vmyroslav, line 15:
```suggestion
		Description: "Values provided by the OpenSearch server.",
```

PR #2148, opened 2025-04-14
fix(clickhouse_grant): don't escape wildcards

joshua-rule:
Fixes an issue where when specifying the wildcard (*) in clickhouse grant database/table etc. would error when attempting to apply the grant

For example setting database to * cause grants to be set as `*`.* rather than *.*

Also fixes issue where when having database = * would cause terraform plan to show changes on subsequent plans after apply

Resolves: #xxxxx.


vmyroslav:
@joshua-rule Thanks for your PR. There is one concern in this implementation.
What if the user will have an entity (TABLE, USER, etc.) with the name `*` which is definitely not the best name, but still legit from the ClickHouse perspective.
In that case will have a totally different results:
```
GRANT INSERT ON test-db.`*` TO foo-role;
VS
GRANT INSERT ON test-db.* TO foo-role;
```
So maybe it is better to implement this check in place where [create](https://github.com/aiven/terraform-provider-aiven/blob/main/internal/sdkprovider/service/clickhouse/grant.go#L170) and [revoke](https://github.com/aiven/terraform-provider-aiven/blob/main/internal/sdkprovider/service/clickhouse/grant.go#L198) the grant.

joshua-rule:
@vmyroslav ah yes, I forgot that * can be used as a name =( your commit LGTM (sorry was on leave for easter)

byashimov (approved):

PR #2169, opened 2025-04-23
chore(go): replaced tools with go toolchain

vmyroslav:
- replaced existing tools with new go toolchain


byashimov (approved):
Great job!

PR #2174, opened 2025-04-25
refactor(organization_address): use structs instead of maps

byashimov:
Resolves NEX-1431.

Uses structs to marshal and unmarshal values to/from TF/API:

1. Uses the `json` package to resolve typing issues.
2. Modifiers can now use typed instances instead of untyped maps.

Additionally moves the address resource into the organization directory.


internal/plugin/service/organization/address/address.go
```diff
@@ -99,7 +78,20 @@ func (c *view) Update(ctx context.Context, plan, state *dataModel) diag.Diagnost
 		return diags
 	}
 
-	return flattenData(ctx, plan, rsp)
+	plan.OrganizationID = types.StringValue(rsp.OrganizationId)
+	plan.AddressID = types.StringValue(rsp.AddressId)
+	return c.Read(ctx, plan)
```
byashimov, line 83: The `View` design allows us to re-use the `Read` method, like in SDKv2.

internal/plugin/util/converters.go
```diff
@@ -9,41 +9,58 @@ import (
...
+// reqRsp — a map that would be used to unmarshal a Request/Response object
+// in *I — input type, for instance, a dataModel or foo.CreateIn dto.
+// Hint: to ignore "in" type and modify only the map, use `any`, for instance:
+//
+//	func foo[T any](reqRsp map[string]any, _ *T) error {
+//		reqRsp["foo"] = "bar"
+//	}
+//
+// In this case, the function can be used with any type of Request/Response object.
+type MapModifier[I any] func(reqRsp map[string]any, in *I) error
```
byashimov, line 55: This is the coolest part: the map can be modified and the `in` would be `dtoModel` or gen-client response (depends on the use-case).

This allows us to convert emails to a list of strings or to a list of maps.

internal/plugin/service/organization/address/zz_converters.go
```diff
@@ -0,0 +1,145 @@
...
+	AddressID      *string   `json:"address_id,omitempty"`
+	AddressLines   *[]string `json:"address_lines"`
+	City           *string   `json:"city"`
+	CompanyName    *string   `json:"company_name,omitempty"`
+	CountryCode    *string   `json:"country_code"`
+	CreateTime     *string   `json:"create_time,omitempty"`
+	OrganizationID *string   `json:"organization_id"`
+	State          *string   `json:"state,omitempty"`
+	UpdateTime     *string   `json:"update_time,omitempty"`
+	ZipCode        *string   `json:"zip_code,omitempty"`
```
byashimov, line 46: Each field is a pointer to handle all the required/optional cases as simple as `if v != nil`.

internal/plugin/service/organization/address/zz_converters.go
```diff
@@ -0,0 +1,145 @@
...
+
+// expandData turns TF object into Request
+func expandData[R any](ctx context.Context, data *dataModel, rqs *R, modifiers ...util.MapModifier[dtoModel]) diag.Diagnostics {
+	dto := new(dtoModel)
+	if !data.AddressLines.IsNull() {
+		vAddressLines, diags := util.ExpandSet[string](ctx, data.AddressLines)
+		if diags.HasError() {
+			return diags
+		}
+		dto.AddressLines = &vAddressLines
```
byashimov, line 57: Values are strictly typed now.

internal/plugin/util/converters.go
```diff
@@ -9,43 +9,60 @@ import (
...
+
+// Expand sets values to Request from a nested obejct.
```
vmyroslav, line 57:
```suggestion
// Expand sets values to Request from a nested object.
```

vmyroslav (approved):

PR #2183, opened 2025-04-29
refactor(organization): generate resource

byashimov:
Resolves NEX-1446.


internal/plugin/adapter/common.go
```diff
@@ -4,22 +4,41 @@ import (
...
 
 	avngen "github.com/aiven/go-client-codegen"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
 )
 
-type DataView[T any] interface {
+// DatView datasource view interface
+type DatView[T any] interface {
```
vmyroslav, line 13: nit: mb `DSView`

internal/plugin/adapter/resource.go
```diff
@@ -39,13 +39,13 @@ func NewResource[T any](
 
 type resourceAdapter[T any] struct {
 	name         string
-	view         View[T]
+	view         ResView[T]
```
vmyroslav, line 42: I guess we need to implement `ConfigValidators` as well as in datasource

internal/plugin/provider.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/aiven/terraform-provider-aiven/internal/plugin/service/organization/address"
 	"github.com/aiven/terraform-provider-aiven/internal/plugin/service/organization/billinggroup"
 	"github.com/aiven/terraform-provider-aiven/internal/plugin/service/organization/billinggrouplist"
-	"github.com/aiven/terraform-provider-aiven/internal/plugin/service/organization/org"
+	organization2 "github.com/aiven/terraform-provider-aiven/internal/plugin/service/organization/organization"
```
vmyroslav, line 23: nit: I guess this package will remain as organization, so maybe in makes sense to add an alias for `github.com/aiven/terraform-provider-aiven/internal/plugin/service/organization` instead

vmyroslav (approved):

internal/plugin/provider.go
byashimov: Will be replaced with the migration of the rest of the resources.

internal/plugin/adapter/common.go
byashimov: Don't know )
I tried to make them of equal length. Somehow it easier to read that way to me. If you don't have a strong opinion on that, I would keep this naming.

internal/plugin/adapter/resource.go
byashimov: That's correct. I'll fix this with the next PR.

PR #2193, opened 2025-05-02
chore(go): golangci-lint upgrade

vmyroslav:

Resolves: #xxxxx.


changelog/differ_test.go
```diff
@@ -3,6 +3,8 @@ package main
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
byashimov, line 7: Please run `fmt-imports`, this change (and others) will appear with next code generation.

changelog/differ_test.go
```diff
@@ -122,7 +124,7 @@ func TestCompare(t *testing.T) {
 	for _, opt := range tests {
 		t.Run(opt.name, func(t *testing.T) {
 			got, err := diffItems(opt.old, opt.new)
-			assert.NoError(t, err)
+			require.NoError(t, err)
```
byashimov, line 127: Do we need to exit on the first error?

examples_tests/postgres_test.go
```diff
@@ -19,8 +19,6 @@ func TestPostgresTestSuite(t *testing.T) {
 }
 
 func (s *PostgresTestSuite) TestPostgres() {
-	s.T().Parallel()
-
```
byashimov, line 23: Why did you remove parallel runs?

internal/acctest/template/generator_framework_test.go
```diff
@@ -251,7 +253,7 @@ func TestFrameworkGenerateTemplate(t *testing.T) {
 			t.Parallel()
 
 			got, err := generator.GenerateTemplate(tt.schema, tt.resourceType, tt.kind)
-			assert.NoError(t, err)
+			require.NoError(t, err)
 			assert.Equal(t, normalizeHCL(tt.want), normalizeHCL(got), "Generated template mismatch")
```
byashimov, line 257: I would let it go through all the cases. Wdyt?

internal/schemautil/helpers_test.go
```diff
@@ -72,12 +74,12 @@ func TestDetermineMixedOrganizationConstraintIDToStore(t *testing.T) {
...
 			got, err := ConvertOrganizationToAccountID(ctx, tt.input, client)
 			if tt.wantErr {
-				assert.Error(t, err)
+				require.Error(t, err)
 
 				return
 			}
 
-			assert.NoError(t, err)
+			require.NoError(t, err)
```
byashimov, line 82: Why require? I mean, those cases are stateless. Let it fail as many times as it can.

internal/schemautil/schemautil_test.go
```diff
@@ -112,7 +111,7 @@ func TestResourceDataSetAddForceNew(t *testing.T) {
 	// Missing force_new field
 	d := mocks.NewMockResourceData(t)
 	err := ResourceDataSet(d, dto, s)
-	require.Error(t, err, fmt.Errorf("%w: %q", errMissingForceNew, "force_new"))
+	require.ErrorIs(t, err, errMissingForceNew)
```
byashimov, line 114: The error message points out that the missing field is `force_new`, and not any other. Please keep the message check.

ucgenerator/models.go
```diff
@@ -197,7 +197,7 @@ func (o *object) isArray() bool {
 }
 
 func (o *object) isScalar() bool {
-	return !(o.isObject() || o.isArray())
+	return !o.isObject() && !o.isArray()
```
byashimov, line 200: So this is applies to De Morgan's law (and all other cases) https://en.wikipedia.org/wiki/De_Morgan%27s_laws
Is this changed happed by any linter? I mean, do we need to write conditions always this way?

byashimov (changes requested):

ucgenerator/models.go
vmyroslav: Yes, it is. This is a part of [staticcheck](https://golangci-lint.run/usage/linters/#staticcheck) linter which we had before. 
For my taste it makes sense, if you don't like it in general, we can disable this particular linter, or you can add `//nolint:staticcheck` to the place where it doesn't work on your opinion.

changelog/differ_test.go
vmyroslav: This is the result of formatting after the `fmt-imports` command.

changelog/differ_test.go
vmyroslav: [The same](https://github.com/aiven/terraform-provider-aiven/pull/2193#discussion_r2075177418)

examples_tests/postgres_test.go
vmyroslav: This was spotted by the linter. I believe that the warning corresponds to [this](https://github.com/stretchr/testify/issues/187) issue, which is 10 years old and doesn't look like it wouldd be resolved in `V1`, so the safer option is to avoid parallel run for test suits.

internal/schemautil/schemautil_test.go
vmyroslav: Added additional check `require.ErrorContains(t, err, "force_new")`

byashimov (approved):

PR #2200, opened 2025-05-07
docs: clarify project_id description for organization_project

staceysalamon-aiven:

Updates the description for the `project_id` field for clarity. Adds a separate example for creating a project at unit level.


examples/resources/aiven_organization_project/resource.tf
```diff
@@ -1,11 +1,19 @@
...
   tag {
-    key = "key_1"
-    value = "value_1"
+    key   = "env"
+    value = "prod"
   }
 }
+
+# Create a project within an organizational unit
+resource "aiven_organization_project" "example_project" {
```
vmyroslav, line 14: Maybe we can provide an example with some random prefix that can be generated?
```hcl
# Define a random_string resource to generate prefix
resource "random_string" "prefix" {
  length  = 8
  special = false
  upper   = false
  numeric = true
}

resource "aiven_organization_project" "foo" {
  organization_id = aiven_organization.foo.id
  project_id      = "aiven-example-${random_string.prefix.result}"
  parent_id = aiven_organization.foo.id
  billing_group_id = aiven_billing_group.foo.id
}
```
WDYT?

staceysalamon-aiven: Good idea, thanks! I'll work on adding this. This also made me realize that we don't have info about what characters you can use in the name, so I'll add that.

vmyroslav (approved):

Issue #2198, opened 2025-05-06
`dynamic` blocks in resources make `terraform plan` fail if the `for_each` value is determined at `apply`


Conflit-de-Bus:
## What happened?

Hello Aiven team!

Today I came accross a bug in the provider, relating to `dynamic` blocks in configuration, for example, `pg_user_config.ip_filter_object` in `aiven_pg` resources.

Using a `dynamic` instruction for these blocks works fine in `terraform plan`, as long as the value used for the `for_each` argument is known during the plan phase. If not, the provider crashes without finishing the plan.

I tried this test with Terraform `1.9.8` as well as `1.11.1`, and Aiven `4.39.0`.

Note that I created a `list_file` at the root of my stack, containing the following values to test things:
```json
[10, 11, 12, 13, 14]
```

Here are some examples of things that DO work, even if the use cases are a bit silly:
```hcl
# Static and unique ip_filter_object: OK

resource "aiven_pg" "pg_static" {
  project      = local.aiven_project
  service_name = "test-pg-static"
  plan         = "hobbyist"
  pg_user_config {
    ip_filter_object {
      description = "Test object"
      network     = "192.168.0.0/24"
    }
  }
}


# "dynamic" ip_filter_object, with static range: OK

resource "aiven_pg" "pg_range" {
  project      = local.aiven_project
  service_name = "test-pg-range"
  plan         = "hobbyist"
  pg_user_config {
    dynamic "ip_filter_object" {
      for_each = range(5)
      content {
        description = "Test object ${ip_filter_object.value}"
        network     = "192.168.${ip_filter_object.value}.0/24"
      }
    }
  }
}


# "dynamic" ip_filter_object, with var range: OK

variable "pg_range" {
  type    = set(number)
  default = [5, 6, 7, 8, 9]
}

resource "aiven_pg" "pg_var" {
  project      = local.aiven_project
  service_name = "test-pg-var"
  plan         = "hobbyist"
  pg_user_config {
    dynamic "ip_filter_object" {
      for_each = var.pg_range
      content {
        description = "Test object ${ip_filter_object.value}"
        network     = "192.168.${ip_filter_object.value}.0/24"
      }
    }
  }
}


# "dynamic" ip_filter_object, with data sourced range: OK

data "local_file" "file" {
  filename = "${path.module}/list_file"
}

resource "aiven_pg" "pg_file" {
  project      = local.aiven_project
  service_name = "test-pg-file"
  plan         = "hobbyist"
[...]

PR #2205, opened 2025-05-09
fix(userconfig): "known after apply" values panic

byashimov:
Resolves NEX-1484, #2198.


byashimov:
tests https://github.com/aiven/terraform-provider-aiven/actions/runs/14925548052

Issue #2210, opened 2025-05-12
Setting of segment_bytes not respected


jessewilliams:
This may be a bug, or maybe I'm just doing something wrong.  Basically, I use Terraform to manage the topics in my Kafka cluster.  For a small amount of topics it will see that segment_bytes needs to be set and will appear to set it, but it doesn't actually get changed on the topic and tfplan when run again will again think it needs to be updated.  I also see the following logs in the logfile...

2025-05-12T15:36:37.368-0700 [INFO]  provider.terraform-provider-aiven_v4.26.0: configuring server automatic mTLS: timestamp=2025-05-12T15:36:37.368-0700
2025-05-12T15:36:43.846-0700 [WARN]  Provider "registry.terraform.io/aiven/aiven" produced an unexpected new value for module.fact.module.kafka_topic["helper"].aiven_kafka_topic.kafka_topic during refresh.
      - .config[0].segment_bytes: was cty.StringVal("10000000"), but now cty.StringVal("1073741824")
2025-05-12T15:36:43.853-0700 [WARN]  Provider "registry.terraform.io/aiven/aiven" produced an invalid plan for module.fact.module.kafka_topic["helper"].aiven_kafka_topic.kafka_topic, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .topic_description: planned value cty.StringVal("") for a non-computed attribute
      - .owner_user_group_id: planned value cty.StringVal("") for a non-computed attribute
      - .config[0].index_interval_bytes: planned value cty.StringVal("4096") for a non-computed attribute
      - .config[0].segment_jitter_ms: planned value cty.StringVal("0") for a non-computed attribute
      - .config[0].min_cleanable_dirty_ratio: planned value cty.NumberFloatVal(0.5) for a non-computed attribute
      - .config[0].min_compaction_lag_ms: planned value cty.StringVal("0") for a non-computed attribute
      - .config[0].max_message_bytes: planned value cty.StringVal("1048588") for a non-computed attribute
      - .config[0].flush_messages: planned value cty.StringVal("9223372036854776000") for a non-computed attribute
      - .config[0].flush_ms: planned value cty.StringVal("9223372036854776000") for a non-computed attribute
      - .config[0].compression_type: planned value cty.StringVal("producer") for a non-computed attribute
      - .config[0].message_downconversion_enable: planned value cty.True for a non-computed attribute
[...]

PR #2212, opened 2025-05-13
fix(kafka_topic): segment_bytes is not sent to the API

byashimov:
Resolves NEX-1495, #2210.


byashimov:
tests https://github.com/aiven/terraform-provider-aiven/actions/runs/14998282041

vmyroslav (approved):

PR #2215, opened 2025-05-13
feat(clickhouse_grant): added ignore for case-sensitive privilege names

vmyroslav:
- added Set function to the `aiven_clickhouse_grant` to ignore case sensitive privilege names during the plan phase

## Why this way
- `DiffSuppressFunc` for `privilege` field does not works in this case, because the privilege is ForceNew and the hash is calculated before this function is called
- `StateFunc` for `privilege` field doesn't work because apparently there is a bug inside the plugin and when we'll transform the `Set` into the list like here: `d.Get("privilege_grant").(*schema.Set).List()` we will get additional elements in the list with empty values
- `CustomizeDiff` for the whole resource looks like a more complex option as it requires to disassemble quite complicated schema


vmyroslav:
[Tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/15000849112)

byashimov (approved):

PR #2234, opened 2025-05-28
docs: add guide for moving projects

staceysalamon-aiven:

### DIFF
--- a/docs/guides/move-projects.md
+++ b/docs/guides/move-projects.md
@@ -1,0 +1,86 @@
+---
+page_title: "Move projects"
+---
+
+# Move projects
+
+You can move a project to another [organization or organizational unit](https://aiven.io/docs/platform/concepts/orgs-units-projects) by updating the `aiven_project` or `aiven_organization_project` resource.
+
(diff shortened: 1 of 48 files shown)


docs/guides/move-projects.md
```diff
@@ -0,0 +1,85 @@
...
+  organization_id  = data.aiven_organization.main.id
+  parent_id        = data.aiven_organization.main.id
+  billing_group_id = aiven_billing_group.main.id
+}
+```
+
+## Move a project to another organization
+
+To move a project to a different organization, you must be an [organization admin](https://aiven.io/docs/platform/concepts/permissions#organization-roles-and-permissions) of both organizations.
+All users with permission to access the project lose the permissions when you move it to a different organization unless they are members of the target organization.
```
vmyroslav, line 45: This sentence looks quite important, maybe we can highlight it with bold/italic or any other options that works best in your opinion?

staceysalamon-aiven: Good call!

vmyroslav (approved):

PR #2238, opened 2025-05-30
feat(aiven_clickhouse_grant): skip named collection grants

vmyroslav:

-  skips the `NAMED COLLECTION` grants for the grants that were added outside of Terraform

## Why this way
- as we can not fully manage the NAMED COLLECTION grants in ClickHouse, [due](https://github.com/ClickHouse/ClickHouse/issues/80853) to the empty `database` field in `system.grants` table we will skip these types of GRANTS that it can be possible to assign them outside of Terraform

### DIFF
--- a/internal/sdkprovider/service/clickhouse/grant.go
+++ b/internal/sdkprovider/service/clickhouse/grant.go
@@ -7,6 +7,8 @@ import (
 	"log"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
 	"github.com/aiven/aiven-go-client/v2"
 )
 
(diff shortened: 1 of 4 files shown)


internal/sdkprovider/service/clickhouse/grant.go
```diff
@@ -136,11 +138,31 @@ func ReadPrivilegeGrants(
...
+
+		if isNamedCollectionPrivilege(grant) {
+			tflog.Debug(ctx, "Skipping named collection privilege grant", map[string]any{
+				"grantee":   grantee,
+				"database":  grant.Database,
+				"privilege": grant.Privilege,
+			})
+
+			continue
+		}
```
byashimov, line 150: So it will just skip it without telling it to user?
Why not raising an error?

vmyroslav: Discussed the purpose of it separately

PR #2306, opened 2025-06-16
ci(GHA): improving pre-release job

vmyroslav:

- added some comments to the jobs
- switched the gatekeeper to reusable workflow


.github/workflows/reusable-gatekeeper.yml
```diff
@@ -0,0 +1,38 @@
+# This workflow acts as a conditional check to determine if a CI pipeline
```
byashimov: Hey. Could you elaborate on what it determines and why please?

vmyroslav: updated

PR #2365, opened 2025-07-15
refactor(pg_database): use gen client

byashimov:
Resolves NEX-1674.


byashimov:
tests https://github.com/aiven/terraform-provider-aiven/actions/runs/16323887586

byashimov:
I tried to add a test, but that takes forever: in order to test deletion flow the service must be powered on again, hence rebuild.

byashimov:
> > I tried to add a test, but that takes forever: in order to test deletion flow the service must be powered on again, hence rebuild.
> 
> I don't see an exit from the recursion in `CheckServiceIsPowered` [function](https://github.com/aiven/terraform-provider-aiven/pull/2365/files#diff-6531b5312c16e9c85a338320bf80d70d135f7057d0d748540534d5e76ed6bc36R1151). Maybe this is the issue?

Here is the first one https://github.com/aiven/terraform-provider-aiven/pull/2365/files#diff-6531b5312c16e9c85a338320bf80d70d135f7057d0d748540534d5e76ed6bc36R1148
And here are two more https://github.com/aiven/terraform-provider-aiven/pull/2365/files#diff-6531b5312c16e9c85a338320bf80d70d135f7057d0d748540534d5e76ed6bc36R1131-R1134

I'll add the test for the clarity, I still have it stashed.

internal/schemautil/service.go
```diff
@@ -1113,3 +1115,38 @@ func ContainsRedactedCreds(config map[string]any) error {
...
+		return err
+	}
+
+	powered := s.State != service.ServiceStateTypePoweroff
+	servicePoweredMap.Store(serviceKey, powered)
+	if powered {
+		return nil
+	}
+
+	return CheckServiceIsPowered(ctx, client, project, serviceName)
```
vmyroslav: I see my mistake. Thats true, there are exit from recursion, but still don't get the necessity of it.

byashimov: The recursion would return `ErrServicePoweredOff` if the stored value is not true.
Either I put this check before and after the `ServiceGet` or I call the function recursively to keep things DRY.

I would even remove `if powered` part at the end of function to emphasize this logic.

vmyroslav: Yes, I get the idea, but to me looks like an ideal example for singleflight

vmyroslav (approved):

PR #2381, opened 2025-07-21
refactor: remove singleflight

byashimov:
Contributes to NEX-1687.

Doesn't really cache the result, but can help with burst calls only.

### DIFF
--- a/internal/schemautil/wait.go
+++ b/internal/schemautil/wait.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net"
 	"strconv"
+	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


internal/schemautil/wait.go
```diff
@@ -376,3 +377,32 @@ func WaitUntilNotFound(ctx context.Context, retryableFunc retryGo.RetryableFunc,
 
 // RetryNotFoundAttempts just a random number
 const RetryNotFoundAttempts = 10
+
+// DoOnce like once.Do but for multiple keys.
```
vmyroslav: Would be nice to specify that only error results are stored

vmyroslav (approved):

PR #2404, opened 2025-08-11
fix(plugin): infinitive recursion

byashimov:
Resolves NEX-1717.

- Fixes stack overflow in the plugin generator
- Adds a validation that prevents cycles
- Improves the patching: 
  - it is possible to set now `optional` or `required` to `false` in yaml file
  - improved merging: for instance, array element can be replaced without defining the whole node
- Reverts generator removal from the update command

### DIFF
--- a/generators/plugin/item.go
+++ b/generators/plugin/item.go
@@ -57,7 +57,7 @@ type IDAttribute struct {
 type Definition struct {
 	Beta        bool                 `yaml:"beta"`
 	Location    string               `yaml:"location"`
-	Schema      map[string]Item      `yaml:"schema,omitempty"`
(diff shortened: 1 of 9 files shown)


generators/plugin/item.go
```diff
@@ -174,6 +183,13 @@ func (item *Item) ancestors() []*Item {
 			items = append(items, v)
 		}
 		v = v.Parent
+
+		// Duplicate check
```
vmyroslav: I'm not sure about the logic on the client side and if it's the case, but from this perspective looks like it makes sense to move this check higher, before moving to the parent here: `v = v.Parent`
It would help to catch circular references if they happens for any reason.

vmyroslav (approved):

PR #2409, opened 2025-08-13
test: remove m3aggregator and influxdb tests

byashimov:
Removes EOL resources tests.


byashimov:
Fixes failing tests https://github.com/aiven/terraform-provider-aiven/actions/runs/16923932376

PR #2415, opened 2025-08-15
fix: fixed example test for PG

vmyroslav:

- fixes path in the PG example test


vmyroslav:
[Tests](https://github.com/aiven/terraform-provider-aiven/actions/runs/16984901649)

byashimov (approved):

PR #2422, opened 2025-08-21
docs: update service fork example

staceysalamon-aiven:

Changes the example for service forking from Cassandra to PG as Cassandra will be EOL soon.


examples/postgres/postgres_fork/service.tf
```diff
@@ -0,0 +1,18 @@
...
+  cloud_name   = "google-europe-west1"
+  plan         = "startup-4"
+  service_name = var.source_pg_name
+}
+
+# Forked PostgreSQL service
+resource "aiven_pg" "postgres_fork" {
+  project      = var.aiven_project
+  cloud_name   = "eu-central-1"
+  plan         = "startup-8"
```
byashimov, line 13: Hey. Does the fork need a bigger plan?

byashimov (approved):

examples/postgres/postgres_fork/service.tf
staceysalamon-aiven: This one worked for me and I thought for an example it might be good to use the lower cost plan.

byashimov: I mean, the fork has startup-8, and the source has startup-4.

staceysalamon-aiven: Oh, I see, I thought you were suggesting it should be a business plan. Sorry! In the example I'm showing that the fork can be in a different cloud and have a different plan as this is mentioned in the docs. I'll also be including the example in said docs here: https://github.com/aiven/aiven-docs/pull/1012/files#diff-417f61167e52bc1d6956f57af91ee8365dd339e6de545f93fc9dbb6148ac2b95

But maybe that's not the best example of plan change. 🤔

staceysalamon-aiven: Changed them to the same plan. 👍🏻

PR #2441, opened 2025-09-09
feat(plugin): generate validation rules

byashimov:
Resolves NEX-1844.

- Generate validation rules from definition schema
- Child notes inherit `Computed` property from the parent.
  This allows to mark a whole block as "computed" with just one attribute.
  Validation rules do not apply to computed fields.

### DIFF
--- a/generators/plugin/utils.go
+++ b/generators/plugin/utils.go
@@ -33,7 +33,12 @@ func mergeSlices[T any](args ...[]T) []T {
 	for _, a := range args {
 		merged = append(merged, a...)
 	}
-	return distinct(merged...)
+
+	result := distinct(merged...)
+	if len(result) == 0 {
+		return nil
+	}
+	return result
 }
 
 func or[T comparable](a, b T) T {
@@ -51,7 +56,7 @@ func orLonger[T ~string](a, b T) T {
 	return b
 }
 
-func orDefault[T any](v *T, def T) T {
+func ptrOrDefault[T any](v *T, def T) T {
 	if v == nil {
 		return def
 	}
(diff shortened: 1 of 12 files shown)


byashimov:
Requires this PR to fix the linting issues https://github.com/aiven/terraform-provider-aiven/pull/2440

generators/plugin/utils.go
```diff
@@ -33,7 +33,12 @@ func mergeSlices[T any](args ...[]T) []T {
 	for _, a := range args {
 		merged = append(merged, a...)
 	}
-	return distinct(merged...)
+
+	result := distinct(merged...)
+	if len(result) == 0 {
+		return nil
```
vmyroslav, line 39: Any specific reason to return the `nil` instead of empty slice here?

generators/plugin/validators.go
```diff
@@ -63,5 +67,52 @@ func genValidators(item *Item) ([]jen.Code, error) {
...
+		"ConflictsWith": item.ConflictsWith,
+		"ExactlyOneOf":  item.ExactlyOneOf,
+		"AtLeastOneOf":  item.AtLeastOneOf,
+		"AlsoRequires":  item.AlsoRequires,
+	}
+
+	validatorKeys := sortedKeys(validators)
+	validatorSeen := false
+	for _, name := range validatorKeys {
+		if validatorSeen {
```
vmyroslav: I assume there are few issues with this validation:
1.  The logic is flawed because the if `validatorSeen` check happens at the beginning of the loop.
This test will fail at the next validator check when seeing the first validator and marking it as seen:
```go
	t.Run("one validator type", func(t *testing.T) {
		item := &Item{
			ConflictsWith: []string{"other_field"},
		}

		_, err := genValidators(item)

		if err != nil {
			t.Fatalf("did not expect an error when using a single validator type, but got: %v", err)
		}
	})
```
2. I dunno if we have the one letter properties in TF configuration, but this regexp: `regexp.MustCompile(`[a-z_][a-z0-9_]+`)` would require at least 2 chars.

generators/plugin/utils.go
byashimov: The use case of this function is merging two values which both could be nil:
```go
a.ConflictsWith = mergeSlices(a.ConflictsWith, b.ConflictsWith)
```
In this case `a.ConflictsWith` will always get a "sized" slice, though it didn't have it.
But you have a point.

PR #2443, opened 2025-09-16
feat: added validation to the service maintenance window time

vmyroslav:

Resolves: NEX-1864
- added validation for `maintenance_window_time` field requiring **HH:mm:ss** format
- validation matches API, so this should not be a breaking change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ nav_order: 1
   disk usage
 - Deprecate `aiven_kafka_topic` field `config.message_format_version` in Kafka 4.0+: this configuration is removed
   and any supplied value will be ignored; for services upgraded to 4.0+, the returned value may be `"None"`.
+- Added validation for services `maintenance_window_time`: requires HH:mm:ss format
 
 ## [4.44.0] - 2025-08-14
 
--- a/internal/schemautil/schemautil.go
+++ b/internal/schemautil/schemautil.go
@@ -196,6 +196,25 @@ func ValidateEmailAddress(v any, k string) (ws []string, errors []error) {
 	return
 }
 
+// ValidateMaintenanceWindowTime ensures a string is a valid HH:mm:ss time format
+func ValidateMaintenanceWindowTime(v any, k string) (ws []string, errors []error) {
+	timeStr, ok := v.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("%q: expected string, got %T", k, v))
+		return
+	}
+
+	if ok, _ := regexp.MatchString("^(2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9]$", timeStr); !ok {
+		errors = append(
+			errors,
+			fmt.Errorf(`%q: must be in HH:mm:ss format where HH is 00-23 (two digits), mm and ss are 00-59 (e.g., "09:30:00", "23:59:59")`, k),
+		)
+		return
+	}
+
+	return
+}
+
 func BuildResourceID(parts ...string) string {
 	finalParts := make([]string, len(parts))
 	for idx, part := range parts {
--- a/internal/schemautil/schemautil_test.go
+++ b/internal/schemautil/schemautil_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"reflect"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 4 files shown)


internal/schemautil/schemautil_test.go
```diff
@@ -127,3 +128,158 @@ func TestResourceDataSetAddForceNew(t *testing.T) {
...
+			name:         "invalid format - wrong separators",
+			value:        "10-30-00",
+			key:          "maintenance_window_time",
+			expectErrors: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
byashimov: Probably an AI slop, but running this subtests in parallel looks a bit overwhelming to me.

internal/schemautil/schemautil_test.go
```diff
@@ -127,3 +128,158 @@ func TestResourceDataSetAddForceNew(t *testing.T) {
...
+	f.Add("10:00")
+	f.Add("invalid")
+
+	f.Fuzz(func(t *testing.T, input string) {
+		_, errs := ValidateMaintenanceWindowTime(input, "test_field")
+
+		// if no errors are returned, the input MUST be a valid parsable time
+		if len(errs) == 0 {
+			_, err := time.Parse("15:04:05", input)
+			require.NoError(t, err, "input %q was considered valid by function but failed standard library parsing", input)
```
byashimov, line 258: I'm wondering if we could just use the `time.Parse` instead? I mean, validate the regexp with it, then we trust it more than the regexp.
I can see it `time.Parse("15:04:05", "9:00")` works (no leading zero), but that could be a len check.
```go
const maintenanceWindowTime = "15:04:05"
var ErrMaintenanceWindowTime = fmt.Error(...)

s := fmt.Sprint(v)
_, err := time.Parse(maintenanceWindowTime, s)
if err != nil || len(maintenanceWindowTime) != len(s) {
     return fmt.Errof("%w: expected HH:mm:ss format time, got %q", ErrMaintenanceWindowTime, s)
}
```

internal/schemautil/schemautil_test.go
```diff
@@ -127,3 +128,158 @@ func TestResourceDataSetAddForceNew(t *testing.T) {
...
+		}
+
+		for i, err := range errs {
+			require.Error(t, err, "error %d should not be nil", i)
+			assert.NotEmpty(t, err.Error(), "error %d should have a non-empty message", i)
+		}
```
byashimov, line 260: nit: I think, making a custom error and asserting it is more strict approach. I can't see a case for a nil error.

internal/schemautil/schemautil_test.go
```diff
@@ -127,3 +128,158 @@ func TestResourceDataSetAddForceNew(t *testing.T) {
...
+			name:         "invalid format - only numbers",
+			value:        "123456",
+			key:          "maintenance_window_time",
+			expectErrors: true,
+		},
+		{
+			name:         "invalid format - wrong separators",
+			value:        "10-30-00",
+			key:          "maintenance_window_time",
+			expectErrors: true,
```
byashimov: Probably an AI slop, which is fine, but the key value is never changed and the `k` value doesn't matter.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ nav_order: 1
   disk usage
 - Deprecate `aiven_kafka_topic` field `config.message_format_version` in Kafka 4.0+: this configuration is removed
   and any supplied value will be ignored; for services upgraded to 4.0+, the returned value may be `"None"`.
+- Changed `maintenance_window_time` validation: added HH:mm:ss format validation
```
byashimov: nit:
```suggestion
- Add validation for services `maintenance_window_time`: require HH:mm:ss format
```
Just to tell it is for "services" for clearer change entry.

internal/schemautil/schemautil.go
```diff
@@ -196,6 +196,23 @@ func ValidateEmailAddress(v any, k string) (ws []string, errors []error) {
...
+	if !ok {
+		errors = append(errors, fmt.Errorf("%q: expected string, got %T", k, v))
+		return
+	}
+
+	// check the regex (HH:mm:ss where HH is 00-23, mm and ss are 00-59, requires two-digit hours)
+	if ok, _ := regexp.MatchString("^(2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9]$", timeStr); !ok {
+		errors = append(errors, fmt.Errorf("%q: must be in HH:mm:ss format where HH is 00-23 (two digits), mm and ss are 00-59 (e.g., \"09:30:00\", \"23:59:59\")", k))
+		return
+	}
```
byashimov, line 213: nit: 
1. The comment is probably redundant. If the error already shares the same information, it’s better to drop the comment so we don’t have to prove that it matches the error message.
2. A backtick could be used to not escape the `"` — just for a bit cleaner code

byashimov (changes requested):
Please see the comments. The AI-generated slop is something we shouldn’t be reviewing. Thanks.

internal/schemautil/schemautil_test.go
vmyroslav: It's not an AI slop, but default purpose of independent `t.Run`

> // Run runs f as a subtest of t called name. It runs f in a separate goroutine
> // and blocks until f returns or calls t.Parallel to become a parallel test.

byashimov: I mean, there is `t.Parallel()` up on the root level.
> // Run runs f as a subtest of t called name. It runs f in a separate goroutine

Yes, that's the case. These are CPU bound without IO operation calls. Go would either run them on different cpu cores or in sequence. Those are just tiny unit-tests, no need to run them in "parallel".

internal/schemautil/schemautil_test.go
vmyroslav: I tried the implementation with double parse to compare input and parsed input, but I didn't like it. To me the regexp here looks stable and self-explanatory.

byashimov (approved):

PR #2444, opened 2025-09-16
feat: add RawMap for precision-safe JSON manipulation

byashimov:
Resolves NEX-1867.

Add RawMap interface to safely modify JSON data while preserving numeric precision. Replaces `map[string]any` approach that caused integer overflow for values > 2^53 during marshal/unmarshal cycles.


internal/plugin/util/rawmap.go
```diff
@@ -0,0 +1,124 @@
...
+func (r *rawMap) Exists(keys ...string) bool {
+	_, _, _, err := jsonparser.Get(r.data, keys...)
+	return err == nil
+}
+
+func (r *rawMap) GetString(keys ...string) (string, error) {
+	return jsonparser.GetString(r.data, keys...)
+}
+
+func (r *rawMap) GetInt(keys ...string) (int, error) {
```
vmyroslav, line 62: Low chance issue, but `jsonparser.GetInt` returns `int64`, which also can lead to the overflow

vmyroslav:
LGTM, but I would consider to use [UseNumber](https://pkg.go.dev/encoding/json#Decoder.UseNumber) instead of the whole implementation and stick to default golang types.

internal/plugin/util/rawmap.go
byashimov: Yes, there’s a small chance. The `int` is used because all the code we have, as well as the generated modules, rely on `int`. Therefore, they wouldn’t accept an `int64` value in the first place on 32 bit architecture.
I double-checked: we still produce builds for i386, but TF can’t realistically run there. The entire provider would need to be reviewed to properly solve this issue.

For convenience, I’ve chosen int.

byashimov:
> LGTM, but I would consider to use [UseNumber](https://pkg.go.dev/encoding/json#Decoder.UseNumber) instead of the whole implementation and stick to default golang types.

Not sure I'm following. What do you mean "default golang types"? :)
The `jsonparser` doesn't provide custom types. Instead, it let's us "get typed value" from json without turning it into map/struct/etc.

vmyroslav:
`dec.UseNumber()` should solve the main issue with overflow, but would not require to replace `map[string]any` with custom `util.RawMap`.
```go
	b, err := json.Marshal(in)
	if err != nil {
		return fmt.Errorf("failed to marshal input: %w", err)
	}

	dec := json.NewDecoder(bytes.NewReader(b))
	dec.UseNumber() //should solve the issue

	var m map[string]any
	if err := dec.Decode(&m); err != nil {
		return fmt.Errorf("failed to decode: %w", err)
	}
```

PR #2462, opened 2025-09-25
feat(service): added service plan validation

vmyroslav:

Resolves: NEX-1895

-  added validation for service plan names with suggestions during plan phase


internal/schemautil/custom_diff.go
```diff
@@ -320,3 +326,89 @@ func CustomizeDiffAdditionalDiskSpace(ctx context.Context, diff *schema.Resource
...
+		project     = d.Get("project").(string)
+		serviceType = d.Get("service_type").(string)
+		plan        = d.Get("plan").(string)
+
+		maxSuggestions = 3
+	)
+
+	if project == "" || serviceType == "" || plan == "" {
+		return nil
+	}
```
byashimov: 1. Could you please add a comment in which case it is possible to not have one of this values? I'm not sure I can think of a real one, is that the "import" flow? I mean, if it is not a case, the code might be misleading, though it is completely legit from the function's perspective.
2. Could you please add a HasChanged check for the "plan" value? If the value remains the same, we don't need to validate it.

internal/schemautil/custom_diff.go
```diff
@@ -320,3 +326,89 @@ func CustomizeDiffAdditionalDiskSpace(ctx context.Context, diff *schema.Resource
...
+
+	plans, err := client.ProjectServicePlanList(ctx, project, serviceType)
+	if err != nil {
+		return fmt.Errorf("unable to fetch service plans: %w", err)
+	}
```
byashimov, line 366: I think, we can use there `schemautil.DoOnce` here (key project+serviceType) to lower the pressure on the API and speed up the check a little (considering the authorization, authentication and the list fetch).

byashimov (changes requested):
Nice idea with the "Did you mean one of"! Please have a loot at the suggestions.

internal/schemautil/custom_diff.go
vmyroslav: Done

internal/schemautil/custom_diff.go
vmyroslav: Added `plan` value diff check

byashimov (approved):

byashimov (approved):

PR #2477, opened 2025-10-01
feat(plugin): generate legacy timeouts block

byashimov:
Resolves NEX-1907.

The Plugin Framework doesn't have the `default` timeout key. That's a breaking change for the resources we are migrating from SDKv2.

- Adds a custom `timeouts` block to the generator
- Adds timeouts string validator
- Uses timeouts to cancel the context in CRUD operations
- Disables `unused-parameter` linter rule (a waste of time)

### DIFF
--- a/internal/plugin/adapter/resource.go
+++ b/internal/plugin/adapter/resource.go
@@ -108,6 +108,13 @@ func (a *resourceAdapter[T]) Create(
 		return
 	}
 
+	ctx, cancel, d := withTimeout(ctx, plan.TimeoutsObject(), timeoutCreate)
+	diags.Append(d...)
+	if diags.HasError() {
+		return
+	}
+	defer cancel()
+
 	diags.Append(a.view.Create(ctx, plan.SharedModel())...)
 	if diags.HasError() {
 		return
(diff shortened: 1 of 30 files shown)


byashimov:
tests https://github.com/aiven/terraform-provider-aiven/actions/runs/18172680831

internal/plugin/adapter/resource.go
```diff
@@ -144,23 +159,30 @@ func (a *resourceAdapter[T]) Update(
...
-		plan  = a.newModel()
-		state = a.newModel()
-		diags = &rsp.Diagnostics
+		config = a.newModel()
+		state  = a.newModel()
+		diags  = &rsp.Diagnostics
 	)
 
-	diags.Append(req.Plan.Get(ctx, plan)...)
+	diags.Append(req.Config.Get(ctx, config)...)
```
vmyroslav: I assume that this change is necessary to retrieve the timeouts from the `config`, but I'm wondering if it can lead to a situation where we can lose some `computed` values due to the fact that we read values from config.
In other methods we read from the plan.

internal/plugin/adapter/resource.go
```diff
@@ -108,6 +108,13 @@ func (a *resourceAdapter[T]) Create(
 		return
 	}
 
+	ctx, cancel, d := withTimeout(ctx, plan.TimeoutsObject(), timeoutCreate)
+	diags.Append(d...)
+	if diags.HasError() {
+		return
+	}
+	defer cancel()
```
vmyroslav, line 116: Does it makes sense to handle `context.DeadlineExceeded` explicitly for better user experience?

internal/plugin/adapter/resource.go
byashimov: I guess, you are right. I have made a separate PR to explain why we need "config" and tried to explain the differences between all three (plan, state, config).

https://github.com/aiven/terraform-provider-aiven/pull/2481

internal/plugin/adapter/resource.go
vmyroslav: Discussed

vmyroslav (approved):

PR #2493, opened 2025-10-09
fix(kafka_quota): added retry to the kafka quotas read

vmyroslav:

Contributes to: NEX-1933
- added retry to the `kafka quotas` read
- fixed organization resources sweeping order

### DIFF
--- a/internal/sdkprovider/service/account/sweep.go
+++ b/internal/sdkprovider/service/account/sweep.go
@@ -3,6 +3,7 @@ package account
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/aiven/go-client-codegen/handler/account"
@@ -117,10 +118,17 @@ func sweepAccounts(ctx context.Context) func(region string) error {
 			return fmt.Errorf("error retrieving a list of accounts : %w", err)
 		}
 
+		// sort accounts so child accounts come first
+		sort.Slice(accounts, func(i, j int) bool {
+			iHasParent := accounts[i].ParentAccountId != nil && *accounts[i].ParentAccountId != ""
+			jHasParent := accounts[j].ParentAccountId != nil && *accounts[j].ParentAccountId != ""
+			return iHasParent && !jHasParent
+		})
+
 		for _, a := range accounts {
 			if err = client.AccountDelete(ctx, a.AccountId); err != nil {
 				if common.IsCritical(err) {
-					return fmt.Errorf("error destroying account %s during sweep: %w", a.AccountName, err)
+					return fmt.Errorf("error destroying account %q during sweep: %w", a.AccountName, err)
 				}
 			}
 		}
--- a/internal/sdkprovider/service/kafka/kafka_quota.go
+++ b/internal/sdkprovider/service/kafka/kafka_quota.go
@@ -56,7 +56,7 @@ Defines the bandwidth limit in bytes/sec for each group of clients sharing a quo
 Every distinct client group is allocated a specific quota, as defined by the cluster, on a per-broker basis.
 Exceeding this limit results in client throttling.`,
 		ValidateFunc: validation.IntBetween(0, 1073741824),
-		AtLeastOneOf: []string{"consumer_byte_rate", "producer_byte_rate", "request_percentage"},
+		AtLeastOneOf: quotaOptionalFields(),
 	},
 	"producer_byte_rate": {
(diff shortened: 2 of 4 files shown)


internal/sdkprovider/service/account/sweep.go
byashimov, line 136: Looks like, it can be solved by sorting.

```go
sort.Slice(accounts, func(i, j int) bool {
  return accounts[i].ParentAccountId != nil && accounts[j].ParentAccountId == nil
})

for _, a := range childAccounts {
	err = client.AccountDelete(ctx, a.AccountId); err != nil {
	if common.IsCritical(err) {
		return fmt.Errorf("error destroying account %q during sweep: %w", a.AccountName, err)
	}
}
```
1. uses same removal loop
2. though, the "child" mark is lost in this solution, the API error will let us know if that's a case
3. collapses both levels of `if err` handling.

internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -208,18 +208,39 @@ func resourceKafkaQuotaDelete(ctx context.Context, d *schema.ResourceData, clien
...
-// We need to retry reading the resource because the Kafka quota may not be immediately available after creation.
+// readKafkaQuotaWithRetry retries reading a Kafka quota resource to handle eventual consistency on API side.
 func readKafkaQuotaWithRetry(ctx context.Context, d *schema.ResourceData, client avngen.Client) error {
-	return retry.RetryContext(ctx, 3*time.Second, func() *retry.RetryError {
+	expectedConsumerRate := d.Get("consumer_byte_rate")
+	expectedProducerRate := d.Get("producer_byte_rate")
+	expectedRequestPct := d.Get("request_percentage")
+
+	return retry.RetryContext(ctx, 5*time.Second, func() *retry.RetryError {
 		err := resourceKafkaQuotaRead(ctx, d, client)
```
byashimov, line 225: Jfyi, `RetryContext` comes from sdkv2 package https://discuss.hashicorp.com/t/terraform-plugin-framework-what-is-the-replacement-for-waitforstate-or-retrycontext/45538

I usually try to use `github.com/avast/retry-go` but this is not mandatory.

internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -208,18 +208,39 @@ func resourceKafkaQuotaDelete(ctx context.Context, d *schema.ResourceData, clien
...
-		if err == nil {
-			return nil
+		if err != nil {
+			// resource may not be yet available after create
+			if avngen.IsNotFound(err) {
+				return &retry.RetryError{Err: err, Retryable: true}
+			}
+
+			return &retry.RetryError{Err: err, Retryable: false}
```
byashimov: nit:

```go
return &retry.RetryError{Err: err, Retryable: avngen.IsNotFound(err)}
```

internal/sdkprovider/service/kafka/kafka_quota.go
```diff
@@ -208,18 +208,39 @@ func resourceKafkaQuotaDelete(ctx context.Context, d *schema.ResourceData, clien
...
+		if cMiss || pMiss || rMiss {
+			return &retry.RetryError{
+				Err:       fmt.Errorf("quota values not yet updated on backend"),
+				Retryable: true,
+			}
 		}
+
+		return nil
 	})
 }
```
byashimov, line 241: I think this code is hard to extend, instead we can loop over the keys:

```go
// can be used with AtLeastOneOf validations too
func optionalFields()[]string {
    return []string{"consumer_byte_rate", "producer_byte_rate", "request_percentage"}
}

func readKafkaQuotaWithRetry(...){
	planValues := make(map[string]any, len(optionalFields())
	for k, v := range optionalFields() {
		planValues := d.Get(k)
	}

	...

	if err != nil {
		return &retry.RetryError{Err: err, Retryable: avngen.IsNotFound(err)}
	}

	for k, v := range planValues {
		if v != d.Get(k) {
			return &retry.RetryError{..., Retryable=true)
		}
	}

	return nil
}

internal/sdkprovider/service/account/sweep.go
vmyroslav: upd

internal/sdkprovider/service/kafka/kafka_quota.go
vmyroslav: upd

PR #2497, opened 2025-10-13
feat(plugin): generate `UseStateForUnknown`

byashimov:
Resolves NEX-1950.

Generates `UseStateForUnknown` [plan modifier](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown) for computed fields.

### DIFF
--- a/generators/plugin/item.go
+++ b/generators/plugin/item.go
@@ -72,21 +72,22 @@ type Definition struct {
 
 type Item struct {
 	Parent              *Item     `yaml:"-"`
-	AppearsIn           AppearsIn `yaml:"appearsIn"`
-	Name                string    `yaml:"name"`
-	InIDAttribute       bool      `yaml:"inIDAttribute"`       // If field is part of ID attribute
-	IDAttributePosition int       `yaml:"idAttributePosition"` // Position in the ID field
+	AppearsIn           AppearsIn `yaml:"-"` // In Create, Read, Update, Delete request or response
+	Name                string    `yaml:"-"`
+	InIDAttribute       bool      `yaml:"-"` // If field is part of ID attribute
+	IDAttributePosition int       `yaml:"-"` // Position in the ID field
 
 	// Tagged fields are exposed to the Definition.yaml
 	Properties map[string]*Item `yaml:"properties"`
 	Items      *Item            `yaml:"items"` // Array item or Map item
 
 	// User-defined fields for YAML generation
-	UserRequired  *bool `yaml:"required"`
-	UserComputed  *bool `yaml:"computed"`
-	UserOptional  *bool `yaml:"optional"`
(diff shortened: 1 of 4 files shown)


generators/plugin/item.go
```diff
@@ -71,21 +71,22 @@ type Definition struct {
...
-	UserComputed  *bool `yaml:"computed"`
-	UserOptional  *bool `yaml:"optional"`
-	UserSensitive *bool `yaml:"sensitive"`
-	UserForceNew  *bool `yaml:"forceNew"`
+	OverrideRequired   *bool `yaml:"required"`
+	OverrideComputed   *bool `yaml:"computed"`
+	OverrideOptional   *bool `yaml:"optional"`
+	OverrideSensitive  *bool `yaml:"sensitive"`
+	OverrideForceNew   *bool `yaml:"forceNew"`
+	UseStateForUnknown bool  `yaml:"useStateForUnknown"`
```
vmyroslav, line 90: I believe it is intentional to make it a property, not a point here unlike others. I just wanna understand why?

generators/plugin/schemas.go
```diff
@@ -223,10 +223,19 @@ func genAttributeValues(isResource bool, item *Item) (jen.Dict, error) {
...
 	if !item.IsNested() {
 		if isResource {
+			typedPlanmodifier := getTypedImport(item.Type, planmodifierTypedImport)
+			planModifiers := make([]jen.Code, 0)
 			if item.ForceNew {
-				typedPlanmodifier := getTypedImport(item.Type, planmodifierTypedImport)
+				planModifiers = append(planModifiers, jen.Qual(typedPlanmodifier, "RequiresReplace").Call())
+			}
+
+			if item.UseStateForUnknown {
```
vmyroslav, line 237: Based on the TF docs it is mostly used for computed fields. Should we check it here to prevent unintentional usage?

generators/plugin/item.go
byashimov: `Override*` fields are "computed" initially by the generator. Mostly, it gets them right, granted all the operation endpoints are provided. But in some rare cases one may want to set `ForceNew` to `false`.
`UseStateForUnknown` is a pure user setting.

generators/plugin/schemas.go
byashimov: Right, that's on the field description too. 
Unfortunately, that's not the only field that requires another field or conflicts with it:

- `ForceNew` also can't be computed only
- A field can't be both `Required` and `Optional`
 
And so on. That said it is a bigger topic, but I don't want to overcomplicate the code that could "do the job" by going with simply "if foo" approach.

Probably, some sort of a schema validation or a `Validate()` method could do a better job in an isolated scope that then can be easily tested.

PR #2503, opened 2025-10-17
feat(kafka_topic): set nil config

byashimov:
Resolves NEX-1969.

Sets `nil` to topic.config if the configuration file doesn't have it. This prevents showing diffs in the Plugin Framework for optional+computed blocks, since the framework doesn't support them.

See: https://github.com/hashicorp/terraform-plugin-framework/issues/883


byashimov:
tests https://github.com/aiven/terraform-provider-aiven/actions/runs/18651271379

PR #2509, opened 2025-10-22
refactor(plugin): replace view interface with individual functions

byashimov:
Resolves NEX-1986.

Replaces interface-based view pattern with individual functions, adopting an SDKv2-like approach.

The interface-stacking pattern had limitations:
- Method name conflicts between resources and datasources required prefixes (Dat/Res)
- Interface contracts forced implementing all methods, preventing selective overrides (it's not Python)

The function-based approach enables generating all CRUD functions while allowing users to use only the ones they need, making code generation more flexible and extensible.


internal/plugin/adapter/resource.go
```diff
@@ -173,7 +197,7 @@ func (a *resourceAdapter[T]) Update(
 		return
 	}
 
-	diags.Append(a.view.Update(ctx, plan.SharedModel(), state.SharedModel(), config.SharedModel())...)
+	diags.Append(a.resource.Update(ctx, a.client, plan.SharedModel(), state.SharedModel(), config.SharedModel())...)
```
vmyroslav: This was before, but anyway, looks like we add timeout to the context after the update operation

internal/plugin/adapter/datasource.go
```diff
@@ -103,18 +99,17 @@ func (a *datasourceAdapter[T]) Read(
...
 		return
 	}
 
 	diags.Append(rsp.State.Set(ctx, state)...)
 }
 
-func (a *datasourceAdapter[T]) ConfigValidators(ctx context.Context) []datasource.ConfigValidator {
-	v, ok := a.view.(DatConfigValidators[T])
-	if !ok {
+func (a *datasourceAdapter[T, M]) ConfigValidators(ctx context.Context) []datasource.ConfigValidator {
```
vmyroslav: I guess i should be `[M, T]` for consistency

Issue #2194, opened 2025-05-06
Data source for plans


sorenhansendk:
## What is currently missing?

The provider does not currently expose a data source that returns the sizing information (number of nodes, CPU, memory and disk) for an existing plan in a given cloud.

## How could this be improved?

<pre>
data "aiven_kafka_plan" "example" {
  cloud_name = "google-europe-west4"
  plan_name  = "business-4"
}

output "node_count" { value = data.aiven_kafka_plan.example.node_count }
output "cpu"        { value = data.aiven_kafka_plan.example.cpu }
output "memory"     { value = data.aiven_kafka_plan.example.memory }
output "disk"       { value = data.aiven_kafka_plan.example.disk }
</pre>

Inputs

- cloud_name – (Required) The name/ID of the cloud
- plan_name  – (Required) The name/ID of the plan

Returned attributes

- **node_count** (total number of nodes in the plan)
- **cpu** (vCPUs per node or aggregated total - whichever matches the API)
- **memory** (Memory in MiB or GiB)
- **disk** (Disk size in GiB)

## Is this a feature you would work on yourself?

* [ ] I plan to open a pull request for this feature.

PR #2527, opened 2025-10-29
feat(aiven_service_plan_list): added new data source for service plans

vmyroslav:

Contributes to: NEX-1990
- added new data source for service plans `aiven_service_plan_list`


docs/guides/find_service_plan.md
```diff
@@ -0,0 +1,36 @@
+---
+page_title: "Find the right service plan"
+---
+
+# Find the right service plan
+
+Use the `aiven_plans` data source to query available service plans and their regional availability.
```
byashimov:
```suggestion
Use the `aiven_service_plan_list` data source to query available service plans and their regional availability.
```

templates/guides/find_service_plan.md
```diff
@@ -0,0 +1,36 @@
+---
+page_title: "Find the right service plan"
+---
+
+# Find the right service plan
+
+Use the `aiven_plans` data source to query available service plans and their regional availability.
```
byashimov:
```suggestion
Use the `aiven_service_plan_list` data source to query available service plans and their regional availability.
```

examples/data-sources/aiven_service_plan_list/data-source.tf
```diff
@@ -0,0 +1,4 @@
+data "aiven_service_plan_list" "kafka_plans" {
+  project      = "example-project-name"
+  service_type = "kafka"
+}
```
byashimov, line 6: I think the guide you have made could fit here. It’s shown on the same page as the datasource itself.

internal/plugin/service/planlist/zz_datasource.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"service_plans": schema.ListNestedBlock{
+				MarkdownDescription: "List of plans available for this type of service.",
+				NestedObject: schema.NestedBlockObject{Attributes: map[string]schema.Attribute{
+					"regions": schema.ListAttribute{
+						Computed:            true,
+						ElementType:         types.StringType,
+						MarkdownDescription: "Service plan hourly price per cloud region.",
```
byashimov: This description now a bit obsolete, since the price was removed.

definitions/service_plan_list.yml
```diff
@@ -0,0 +1,25 @@
...
+  fields:
+    - project
+    - service_type
+remove:
+  - service_plans/node_count
+  - service_plans/shard_count
+  - service_plans/max_memory_percent
+  - service_plans/backup_config
+  - service_plans/service_type
+operations:
```
byashimov, line 17: I have double-checked the issue https://github.com/aiven/terraform-provider-aiven/issues/2194 which is not directly referred in this PR, but we probably could cover it here as well.

There are two separate endpoints that you need to combine to get node count, cpu. etc, and the price:

1. https://api.aiven.io/doc/#tag/Project/operation/ProjectServicePlanPriceGet
2. https://api.aiven.io/doc/#tag/Project/operation/ProjectServicePlanSpecsGet

This will be a real challenge for the generator.
While this datasource can have everything natively. That said, the issue might be solved like this:

```hcl
output "node_count" { value = local.business_plan.node_count }
output "cpu"        { value = local.business_plan.regions[region].node_cpu_count.maximum }
output "memory"     { value = local.business_plan.regions[region].node_memory_mb.maximum }
output "disk"       { value = local.business_plan.regions[region].disk_space_mb }
```

What do you think?

internal/plugin/service/planlist/plan_list.go
```diff
@@ -0,0 +1,68 @@
...
+
+		index := fmt.Sprintf("[%d]", i)
+		err := rsp.Set(regionNames, "service_plans", index, "regions")
+		if err != nil {
+			return fmt.Errorf("failed to set regions for plan %d: %w", i, err)
+		}
+	}
+
+	return nil
+}
```
byashimov, line 68: Just a suggestion: the API field "regions" originally comes with objects. If we change its type to a list of strings, there will be no way back.

I suggest having a virtual field here, something like "region_names", so "just in case" we always have an option to add "regions" back with whatever values we need from it.

generators/plugin/schemas.go
```diff
@@ -205,7 +205,7 @@ func genSetNestedBlock(isResource bool, item *Item) (jen.Code, error) {
 	}
 
 	t := "SetNestedBlock"
-	if item.IsObject() {
+	if item.IsObject() || item.IsList() {
```
byashimov, line 208: Nice catch 👍

definitions/service_plan_list.yml
vmyroslav: As described in the ticket, I want to have a separate resource just for one plan with all specified information. 
There are too much data inside one list, the main purpose for it from from my perspective was a list with proper names and the other one could contain details only if needed.

internal/plugin/service/planlist/zz_datasource.go
vmyroslav: Done

internal/plugin/service/planlist/plan_list.go
vmyroslav: Done

byashimov (approved):

PR #2533, opened 2025-10-30
feat(aiven_plans): added new ds for aiven_service_plan

vmyroslav:

Resolves: NEX-1990 & #2194 
- added new DS for `aiven_service_plan`


internal/plugin/service/plan/plan.go
```diff
@@ -0,0 +1,70 @@
...
+			}
+		}
+
+		if err = rsp.Set(cloudName, "cloud_name"); err != nil {
+			return fmt.Errorf("failed to set cloud_name: %w", err)
+		}
+
+		if err = rsp.Set(servicePlan, "service_plan"); err != nil {
+			return fmt.Errorf("failed to set service_plan: %w", err)
+		}
```
byashimov, line 64: Out of curiosity, why did you set service_plan and the cloud_name? They are already in the plan. Was there an issue?

Issue #2501, opened 2025-10-14
Data source for Organization groups


mikkoc:
## What is currently missing?

A way to get the list of available `organization_user_groups`

## How could this be improved?

Add a data source to list the available groups in the Organization.
Like https://registry.terraform.io/providers/aiven/aiven/latest/docs/data-sources/organization_user_list but for groups.

## Is this a feature you would work on yourself?

No

PR #2538, opened 2025-11-04
feat(organization_user_group_list): added new ds for organization_user_groups

ndk:
Resolves: NEX-1968, #2501.

Added Organization groups datasource.


internal/plugin/service/organization/usergrouplist/user_group_list.go
```diff
@@ -0,0 +1,37 @@
+package usergrouplist
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+
+	avngen "github.com/aiven/go-client-codegen"
```
vmyroslav: We have shared formatters. Please use it via `task fmt`

definitions/organization_user_group_list.yml
```diff
@@ -0,0 +1,36 @@
...
+idAttribute:
+  fields:
+    - organization_id
+operations:
+  UserGroupsList: read
+schema:
+  user_groups:
+    description: User Groups
+    items:
+      properties:
```
vmyroslav: You can avoid specifying the whole list of properties if they are suitable in the API response. However, I dunno if we should expose all information, and does it have any benefits, e.g. `create_time`, `update_time`

internal/plugin/service/organization/usergrouplist/user_group_list_test.go
```diff
@@ -0,0 +1,48 @@
...
+  organization_id = data.aiven_organization.org.id
+  depends_on      = [aiven_organization_user_group.group]
+}
+`, acc.OrganizationName(), groupName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(dataSourceName, "id"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "user_groups.#"),
+					resource.TestCheckTypeSetElemNestedAttrs(dataSourceName, "user_groups.*", map[string]string{
+						"user_group_name": groupName,
+						"description":     "test group description",
```
vmyroslav, line 42: Lets also check other properties that we have for this DS

internal/plugin/provider.go
```diff
@@ -201,6 +202,7 @@ func (p *AivenProvider) DataSources(context.Context) []func() datasource.DataSou
 			billinggroup.NewDataSource,
 			billinggrouplist.NewDataSource,
 			project.NewDataSource,
+			usergrouplist.NewDataSource,
```
vmyroslav: This resource doesn't use any experimental endpoints, so looks like we can make it public

.gitignore
```diff
@@ -62,6 +62,7 @@ tmp/
 temp/
 dist/
 opa/dist/
+.env.op
```
vmyroslav: Do we really need this or it's just an artifact?

ndk: removed

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ nav_order: 1
 - Change `aiven_influxdb` resource field `influxdb_user_config`: deprecate: This property is deprecated
 - Upgraded `go` version to `1.25`
 - Change `aiven_service_component` datasource field `usage` (enum): add `ha_replica`
+- Add `aiven_organization_user_group_list` data source
```
byashimov, line 41: Let's add some description to the new resource:

```suggestion
- Add `aiven_organization_user_group_list` data source: List all user groups within an organization
```

PR #2541, opened 2025-11-05
fix(organization): retry `AccountList` 404

byashimov:
Resolves NEX-2034.

The `AccountList` takes no parameters, but randomly fails with 404. Adds retries for eventual consistency.

### DIFF
--- a/internal/plugin/service/organization/organization/organization.go
+++ b/internal/plugin/service/organization/organization/organization.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	avngen "github.com/aiven/go-client-codegen"
 	"github.com/aiven/go-client-codegen/handler/account"
+	"github.com/avast/retry-go"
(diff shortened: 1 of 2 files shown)


internal/plugin/service/organization/organization/organization.go
```diff
@@ -178,7 +180,22 @@ func getAccountID(ctx context.Context, client avngen.Client, state *tfModel) (st
...
+			list, err = client.AccountList(ctx)
+			return err
+		},
+		// This error can randomly occur.
+		// Granted, the token is valid, we can retry `AccountList` safely, because the token is the only "parameter".
+		// This might take up to a minute or more.
+		retry.RetryIf(avngen.IsNotFound),
+		retry.LastErrorOnly(true),
+		retry.Attempts(10),
+		retry.Delay(time.Second*10),
```
vmyroslav, line 196: Even if the API is down, 100 sec waiting time looks a little extreme, can we make it shorter?

byashimov: No, because it actually takes over a minute according to my observations.

vmyroslav (approved):

PR #2545, opened 2025-11-06
feat(plugin): generate list DTOs filtering

byashimov:
Resolves NEX-2029.

- Generates "for loop" to find a managed resource in list DTOs.
  See `aiven_application_user_token` view
- Generates `aiven_application_user` and `aiven_application_user_token`,
  renames the modules according to Go naming convention


generators/plugin/item.go
```diff
@@ -63,7 +63,34 @@ func operationToHandler() map[Operation]AppearsIn {
...
+
+type Operations []Operation
+
+func (o Operations) ByID(operationID OperationID) Operation {
+	for _, operation := range o {
+		if operation.ID == operationID {
+			return operation
+		}
+	}
+	return Operation{}
```
ndk, line 84: How about letting people to distinguish not-found cases? For instance, we can introduce `IsZero/IsEmpty()` I'm not sure if we need it. :)

byashimov: The way we use it, it is fine if the value is empty. Check out the use case: only the match counts.

PR #2546, opened 2025-11-06
feat(clickhouse_user): added write only password

vmyroslav:

Resolves: NEX-2032
- added write-only passwords for ClickHouse user

### DIFF
--- a/internal/schemautil/schema_converter.go
+++ b/internal/schemautil/schema_converter.go
@@ -6,7 +6,21 @@ import (
 
 func ResourceSchemaAsDatasourceSchema(d map[string]*schema.Schema, required ...string) map[string]*schema.Schema {
 	s := make(map[string]*schema.Schema)
+
+outer:
 	for k, v := range d {
+		// skip write-only fields in data sources
+		if v.WriteOnly {
+			continue
+		}
+
+		// skip fields that require write-only fields (they only make sense together)
+		for _, rw := range v.RequiredWith {
+			if d[rw].WriteOnly {
+				continue outer
+			}
+		}
+
 		s[k] = &schema.Schema{
 			Type:        v.Type,
 			Computed:    true,
@@ -15,6 +29,7 @@ func ResourceSchemaAsDatasourceSchema(d map[string]*schema.Schema, required ...s
 			Elem:        v.Elem,
 		}
 	}
+
 	for _, k := range required {
 		s[k].Required = true
 		s[k].Optional = false
--- a/internal/sdkprovider/service/clickhouse/clickhouse_user.go
+++ b/internal/sdkprovider/service/clickhouse/clickhouse_user.go
@@ -7,6 +7,7 @@ import (
 	avngen "github.com/aiven/go-client-codegen"
 	"github.com/aiven/go-client-codegen/handler/clickhouse"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/aiven/terraform-provider-aiven/internal/common"
 	"github.com/aiven/terraform-provider-aiven/internal/schemautil"
@@ -16,7 +17,6 @@ import (
 var aivenClickhouseUserSchema = map[string]*schema.Schema{
 	"project":      schemautil.CommonSchemaProjectReference,
 	"service_name": schemautil.CommonSchemaServiceNameReference,
-
 	"username": {
 		Type:         schema.TypeString,
 		Required:     true,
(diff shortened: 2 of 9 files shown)


internal/sdkprovider/service/clickhouse/clickhouse_user.go
```diff
@@ -28,7 +27,22 @@ var aivenClickhouseUserSchema = map[string]*schema.Schema{
...
+		WriteOnly:     true,
+		RequiredWith:  []string{"password_wo_version"},
+		ConflictsWith: []string{"password"},
+		Description:   "The password of the ClickHouse user (write-only, not stored in state). Must be used with password_wo_version.",
+	},
+	"password_wo_version": {
+		Type:         schema.TypeInt,
+		Optional:     true,
+		RequiredWith: []string{"password_wo"},
+		Description:  "Version number for password_wo. Increment this to rotate the password.",
```
byashimov: 1. Could you wrap field names with backticks please? I think, it makes just a bit better UX.
2. I would set validation rule requiring to have `>=1` as a valid value for `password_wo_version` to distinguish it from the "zero" value
3. ~and `length > 1` for `password_wo` (if TF allows to have `""`).~ Clickhouse can have an empty password.

So you don't need to validate the values in the code.

internal/sdkprovider/service/clickhouse/clickhouse_user.go
```diff
@@ -62,17 +78,33 @@ func resourceClickhouseUserCreate(ctx context.Context, d *schema.ResourceData, c
...
+			return err
+		}
+		return resourceClickhouseUserRead(ctx, d, client)
+	}
+
+	// use generated password
+	if u.Password != nil {
+		if err = d.Set("password", *u.Password); err != nil {
+			return err
+		}
```
byashimov, line 107: 1. We could rely on `password_wo_version` which if greater than `1` is a solid sign to detect `password_wo`. 
2. I personally would go for `if else` approach, so eventually you have just one `resourceClickhouseUserRead`. Because if we add a new field/logic there are now two branches before "read".

internal/sdkprovider/service/clickhouse/clickhouse_user.go
```diff
@@ -133,3 +182,57 @@ func resourceClickhouseUserDelete(ctx context.Context, d *schema.ResourceData, c
 
 	return nil
 }
+
+// setWriteOnlyPassword sets a custom password using the write-only password pattern
+func setWriteOnlyPassword(ctx context.Context, d *schema.ResourceData, client avngen.Client, projectName, serviceName, uuid string) error {
+	passWo := d.GetRawConfig().GetAttr("password_wo")
+	if passWo.IsNull() {
+		return fmt.Errorf("password_wo must be provided when using write-only password mode")
+	}
```
byashimov: Please use `d.Get("password_wo")` here if possible. Or leave a comment explaining the reasons. The raw config is an unusual way to get values and might be mistakenly interpreted as if it was a special case. I hope it is not. Better have it explicit.

internal/sdkprovider/service/clickhouse/clickhouse_user.go
```diff
@@ -133,3 +182,57 @@ func resourceClickhouseUserDelete(ctx context.Context, d *schema.ResourceData, c
...
+			Password: &customPassword,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("cannot set ClickHouse user password: %w", err)
+	}
+
+	if err = d.Set("password_wo_version", d.Get("password_wo_version").(int)); err != nil {
+		return err
+	}
```
byashimov: I think, the `ResourceData` already has the value. Do you need to set it back?

internal/sdkprovider/service/clickhouse/clickhouse_user.go
```diff
@@ -133,3 +182,57 @@ func resourceClickhouseUserDelete(ctx context.Context, d *schema.ResourceData, c
...
+
+		if oldRaw == nil || oldRaw.(int) == 0 {
+			return nil
+		}
+
+		oldVersion := oldRaw.(int)
+		newVersion := newRaw.(int)
+
+		if newVersion <= oldVersion {
+			return fmt.Errorf("password_wo_version must be incremented (old: %d, new: %d)", oldVersion, newVersion)
```
byashimov: Isn't it like a "hash" so we detect the change? I mean, it shouldn't be an issue going back and forth until the value has been changed.

internal/schemautil/schema_converter.go
```diff
@@ -7,6 +7,11 @@ import (
 func ResourceSchemaAsDatasourceSchema(d map[string]*schema.Schema, required ...string) map[string]*schema.Schema {
 	s := make(map[string]*schema.Schema)
 	for k, v := range d {
+		// skip write-only fields in data sources
+		if v.WriteOnly {
+			continue
+		}
+
```
byashimov, line 16: I would also check for dependencies, no need to walk the whole tree, though:

```go
for _, rw := range v.RequiredWith {
    if s[rw].WriteOnly {
        continue outer
    }
}
```

If this helps you can revert changes in [clickhouse_user_data_source.go‎](https://github.com/aiven/terraform-provider-aiven/pull/2546/files#diff-d178b81cab0090fe982c9366b618ad35ac2b82545a1fe5bb766bb50956447ded) so we don't need to track this manually.

internal/sdkprovider/service/clickhouse/clickhouse_user.go
```diff
@@ -28,7 +27,22 @@ var aivenClickhouseUserSchema = map[string]*schema.Schema{
 		Type:        schema.TypeString,
 		Sensitive:   true,
 		Computed:    true,
-		Description: "The password of the ClickHouse user.",
+		Description: "The password of the ClickHouse user (generated).",
```
byashimov: nit: I would add something like "empty when `password_wo` is used. You did a nice job there by resetting it when `wo` field is used, I think it makes sense to put that detail in the description.

internal/schemautil/schema_converter.go
vmyroslav: done

internal/sdkprovider/service/clickhouse/clickhouse_user.go
vmyroslav: I still use passWo as branching logic to get this variable. change to `if else`

internal/sdkprovider/service/clickhouse/clickhouse_user.go
vmyroslav: Done

PR #2555, opened 2025-11-12
chore(tests): added backward compatibility test template

vmyroslav:

Resolves: NEX-2043
- added an example of compatibility test, showing how to test backward compatibility to ensure that new versions can properly work with an existing resource state
- added few wrappers that may simplify this type o testing

### DIFF
--- a/internal/acctest/backward_compat.go
+++ b/internal/acctest/backward_compat.go
@@ -1,0 +1,125 @@
+package acctest
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+var (
+	cachedVersion string
+	versionOnce   sync.Once
+)
+
+// GetLastStableVersion returns the most recent published version of the Aiven provider.
+// Returns an error if the version cannot be detected from environment or git tags.
+func GetLastStableVersion() (string, error) {
+	var err error
+	versionOnce.Do(func() {
+		if envVersion := os.Getenv("AIVEN_BACKWARD_COMPAT_VERSION"); envVersion != "" {
+			cachedVersion = envVersion
+			return
+		}
+
+		cachedVersion, err = gitTags()
+	})
+
+	if cachedVersion == "" && err == nil {
+		err = fmt.Errorf("version not found: set AIVEN_BACKWARD_COMPAT_VERSION or ensure git tags are available")
+	}
+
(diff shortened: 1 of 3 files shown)


internal/acctest/backward_compat.go
```diff
@@ -0,0 +1,115 @@
...
+		Source:            "aiven/aiven",
+		VersionConstraint: version,
+	}
+}
+
+// tryGitTags attempts to get the latest version from git tags
+// Returns empty string on failure
+func tryGitTags() string {
+	// get all tags sorted by version
+	cmd := exec.Command("git", "tag", "--sort=-version:refname", "--list", "v*")
```
byashimov: The `list` doesn't take arguments, what does `v*` mean?

internal/acctest/backward_compat.go
```diff
@@ -0,0 +1,115 @@
...
+
+	lines := strings.Split(strings.TrimSpace(string(output)), "\n")
+	if len(lines) == 0 {
+		return ""
+	}
+
+	// get the first (latest) tag
+	version := strings.TrimSpace(lines[0])
+
+	return strings.TrimPrefix(version, "v")
```
byashimov: nit: `git tag -l --sort=-version:refname | head -n 1`

internal/acctest/backward_compat.go
```diff
@@ -0,0 +1,115 @@
...
+	"os"
+	"os/exec"
+	"strings"
+	"sync"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+// LastStableVersion is the fallback version if detection fails.
+const LastStableVersion = "4.46.1"
```
byashimov: nit: probably shouldn't be public

byashimov: I was confused first, but then I realized that this is probably "a pinned version" we explicitly want to have here and test against to. Is this correct? If so I would update the description, because it is an explicit reason. One could just use the "last" tag, but we might know the version with compatibility issues and put it here.

internal/acctest/backward_compat.go
vmyroslav: `v` is a search pattern here

internal/acctest/backward_compat.go
vmyroslav: Implemented

internal/acctest/backward_compat.go
vmyroslav: Removed hardcoded version and rely only on env variable or git tags as discussed

Issue #2571, opened 2025-11-21
Support ephemeral values for service users


michael-follmann:
## What is currently missing?

It can be considered as best practice to avoid sensitive data in the Terraform state. [Ephemeral resources and write-only arguments](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/ephemeral) are good way to achieve this.

However, the service user resources, like `aiven_kafka_user`, don't not support this yet.

## How could this be improved?

I suggest something similar to the [resource aws_secretsmanager_secret_version](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret_version#secret_string_wo-1) (`secret_string` argument) and [ephemeral aws_secretsmanager_secret_version](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/secretsmanager_secret_version):

**Providing password as ephemeral**:

```terraform
ephemeral "aiven_kafka_user" "example" {
  username     = "..."
  project      = "..."
  service_name = "..."
}
```

or

```terraform
ephemeral "aiven_kafka_user_password" "example" {
  username     = "..."
  project      = "..."
  service_name = "..."
}
```

or 

```terraform
ephemeral "aiven_service_user_password" "example" {
  username     = "..."
  project      = "..."
  service_name = "..."
  service_type = "kafka"
}
```

**Setting password from external ephemeral resource**:

```terraform
resource "aiven_kafka_user" "example" {
  username            = "..."
  password_wo         = ephemeral.aws_secretsmanager_secret_version.kafka_user_password.secret_string
  password_wo_version = provider::time::rfc3339_parse(data.aws_secretsmanager_secret.kafka_user_password.last_changed_date).unix
  project             = "..."
  service_name        = "..."
}
```

## Is this a feature you would work on yourself?

* [ ] I plan to open a pull request for this feature.

maciekm:
Thanks for the `password_wo` addition.

However, the `aiven_kafka_user` resource also exposes `access_key` (the client's TLS private key) as computed attribute, and is still stored in plaintext in state.

From a security standpoint, `access_key` is equivalent to a password, it's the secret that authenticates the client in mTLS. Leaving it in state while removing password doesn't fully address the concern raised in this issue.

Are there plans to address it?

PR #2572, opened 2025-11-24
feat(service_user): added WO password to service_user

vmyroslav:

Resolves: NEX-2051 && #2571 

Added write-only password support to service users

Extends write-only password functionality (introduced for ClickHouse) to additional service user resources, allowing passwords to be managed without storing them in Terraform state.

Service users with write-only password support:
  - `aiven_cassandra_user`
  - `aiven_kafka_user`
  - `aiven_mysql_user`
  - `aiven_opensearch_user`
  - `aiven_pg_user`
  - `aiven_valkey_user`


ndk (approved):

internal/acctest/backward_compat.go
```diff
@@ -84,6 +84,12 @@ type BackwardCompatConfig struct {
 
 	// OldProviderVersion is the version to test against (defaults to latest stable)
 	OldProviderVersion string
+
+	// PlanOnly controls whether verification step runs with PlanOnly mode (defaults to false).
+	// When false, verification step performs a refresh during plan (`terraform plan`) to allow Read functions to migrate state.
+	// When true, verification only compares state to schema without calling Read functions
+	// (equivalent to `terraform plan -refresh=false`).
+	PlanOnly bool
```
byashimov, line 95: nit: the description is clear and correct. Though I suggest adding an example where it comes handy or important. This PR is the one that ran into the case where that matters: refreshing the plan looses certain information which is important for computed and wo fields (I guess).

internal/schemautil/schemautil.go
```diff
@@ -538,3 +542,15 @@ func Remarshal(in, out any) error {
...
+// Later schemas take precedence over earlier ones in case of key conflicts.
+func MergeSchemas(base map[string]*schema.Schema, schemas ...map[string]*schema.Schema) map[string]*schema.Schema {
+	for _, s := range schemas {
+		for k, v := range s {
+			base[k] = v
+		}
+	}
+
+	return base
+}
```
byashimov, line 550: nit: It is more or less a generic operation, already covered in `lo`.
https://github.com/samber/lo?tab=readme-ov-file#assign

internal/schemautil/service_user.go
```diff
@@ -165,3 +157,148 @@ func TestAccCheckAivenServiceUserAttributes(n string) resource.TestCheckFunc {
...
+		"password_wo": {
+			Type:          schema.TypeString,
+			Optional:      true,
+			Sensitive:     true,
+			WriteOnly:     true,
+			RequiredWith:  []string{"password_wo_version"},
+			ConflictsWith: []string{"password"},
+			ValidateFunc:  validation.StringIsNotEmpty,
+			Description:   "The password of the service user (write-only, not stored in state). Must be used with `password_wo_version`. Cannot be empty.",
+		},
```
byashimov, line 179: nit: It turns out, "write only" requires Terraform >=1.11 https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments. I would add that to the description just in case.

internal/sdkprovider/service/kafka/kafka_user_test.go
```diff
@@ -14,33 +14,117 @@ import (
...
+			{
+				// optional password
+				Config: testAccKafkaUserResource(rName, "FinalPass$456"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "service_name", fmt.Sprintf("test-acc-sr-%s", rName)),
+					resource.TestCheckResourceAttr(resourceName, "project", acc.ProjectName()),
+					resource.TestCheckResourceAttr(resourceName, "username", fmt.Sprintf("user-%s", rName)),
+					resource.TestCheckResourceAttr(resourceName, "password", "FinalPass$456"),
+					resource.TestCheckNoResourceAttr(resourceName, "password_wo"),
+					resource.TestCheckResourceAttr(resourceName, "password_wo_version", "0"),
```
byashimov, line 185: nit: The readability can be improved by moving the common checks into a function or putting the ones that matter first. It's a bit hard to follow what I should be focusing on, which check indicates the reason.

internal/sdkprovider/service/mysql/mysql_user.go
```diff
@@ -15,51 +15,47 @@ import (
...
+		},
+		"access_key": {
+			Type:        schema.TypeString,
+			Sensitive:   true,
+			Computed:    true,
+			Description: "Access certificate key for the user.",
+		},
+	}
+
+	return schemautil.MergeSchemas(s, schemautil.ServiceUserPasswordSchema())
```
byashimov, line 58: nit: Service typed descriptions can be preserved/achieved by passing "Mysql" to `ServiceUserPasswordSchema` as an argument. Not necessarily, but if there are cases where you find specific description helpful, this is an option.

internal/schemautil/service_user.go
```diff
@@ -165,3 +157,148 @@ func TestAccCheckAivenServiceUserAttributes(n string) resource.TestCheckFunc {
...
+					Operation:   service.ServiceUserCredentialsModifyOperationTypeResetCredentials,
+				})
+			if err != nil {
+				return fmt.Errorf("cannot update optional password: %w", err)
+			}
+		}
+	}
+
+	return nil
+}
```
byashimov, line 264: nit: I think, the whole thing can be simplified (both for update and create). Please have a look, this can be heavily tested using `ResourceData` only.
```go
func shouldReset(d ResourceData) (string, bool) {
  // The validations we have prove only one field can be set, 
  // non of them can be empty
  password := d.GetRawConfig().GetAttr("password_wo").AsString()
  if password == "" {
    password = d.Get("password").(string)
  }

  // Either it is a new resources, or any of the "triggers" fired
  ok := d.ID() == "" || d.HasChange("password") || d.HasChange("password_wo_version")
  return password, ok
}

func UpsertPassword(ctx context.Context, d ResourceData, client avngen.Client) error {
  password, ok := shouldReset(d)
  if !ok {
    return // nothing to do
  }
  
  // projectName, serviceName, username — can be reducted from arguments,
  // and fetched here, they are in the schema, e.g.: d.Get("project")
  _, err := client.ServiceUserCredentialsModify(...NilIfZero(password)...)
  return err
}
```

internal/schemautil/service_user.go
```diff
@@ -165,3 +157,148 @@ func TestAccCheckAivenServiceUserAttributes(n string) resource.TestCheckFunc {
...
+func ServiceUserPasswordSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"password": {
+			Type:             schema.TypeString,
+			Optional:         true,
+			Sensitive:        true,
+			Computed:         true,
+			DiffSuppressFunc: EmptyObjectDiffSuppressFunc,
+			ConflictsWith:    []string{"password_wo"},
+			Description:      "The password of the service user (auto-generated if not provided). Empty when using `password_wo`.",
```
byashimov: Let's add a validation. This field can't have empty value. E.g. `validation.StringIsNotEmpty`

internal/acctest/backward_compat.go
vmyroslav: Added additional note to the description. Hard to justify, just to have a chance to check if needed. As far as I now some devs prefer have -`refresh=false` on CI. So we can test it if needed

internal/schemautil/schemautil.go
vmyroslav: Used `lo` inside function, but still left `MergeSchemas` cause to me `lo.Assign` lacks of transparency

internal/schemautil/service_user.go
vmyroslav: `terraform-plugin-docs` [does](https://github.com/aiven/terraform-provider-aiven/pull/2572/files#diff-1da603931d905d3662182e8a1555cc585e97e1eeabebafe294ac48ccb677a627R48) it under the hood

internal/sdkprovider/service/kafka/kafka_user_test.go
vmyroslav: Tried to add separators with short comments to make them more readable

internal/schemautil/service_user.go
vmyroslav: Added

internal/schemautil/service_user.go
vmyroslav: Refactored with this approach. Added unit tests

PR #2587, opened 2025-12-04
feat(plugin): mark beta if any of operations is experimental

byashimov:
Resolves NEX-2165.

When endpoints become stable, this will remove the "beta" flag automatically.

### DIFF
--- a/generators/plugin/main.go
+++ b/generators/plugin/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dave/jennifer/jen"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"github.com/samber/lo"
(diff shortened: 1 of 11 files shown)


generators/plugin/main.go
```diff
@@ -352,6 +353,11 @@ func fromOperationID(scope *Scope, operationID OperationID, root *Item) error {
 		return err
 	}
 
+	// If the operation is experimental and user has not overridden the beta flag.
+	if path.Experimental && scope.Definition.Beta == nil {
+		scope.Definition.Beta = lo.ToPtr(true)
```
ndk, line 358: BTW, in upcoming Golang 1.26 we don't need it anymore because `new` will accept expressions as well :)
https://go.dev/doc/go1.26

ndk (approved):

PR #2599, opened 2025-12-10
feat(plugin): add oneOf fields to the field description

byashimov:
Resolves NEX-2184.

Improves documentation by listing "oneOf" required fields.

### DIFF
--- a/generators/plugin/utils.go
+++ b/generators/plugin/utils.go
@@ -80,9 +80,9 @@ func sortedKeys[K ~string, V any](m map[K]V) []K {
 
 var reNewline = regexp.MustCompile(`\s*\n+\s*`)
 
-func fmtDescription(isResource bool, item *Item) string {
+func fmtDescription(def *Definition, entity entityType, item *Item) string {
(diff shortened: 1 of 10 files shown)


generators/plugin/utils.go
```diff
@@ -94,8 +94,12 @@ func fmtDescription(isResource bool, item *Item) string {
 		b.PossibleValuesString(schemautil.FlattenToString(item.Enum)...)
 	}
 
+	isResource := entity.isResource()
 	if !isResource {
 		b.MarkAsDataSource()
+		if item.IsRootProperty() && slices.Contains(def.Datasource.ExactlyOneOf, item.Name) {
```
ndk, line 100: Does it work with nested items? Shouldn't we check the first item instead?

byashimov: ExactlyOneOf doesn't have semantic to support that. It can be either way `foo.0.bar` or `foo/bar`. Currently we don't have a such case, so I don't implement it.

PR #2604, opened 2025-12-11
refactor(organizational_unit): generate

byashimov:
Resolves NEX-2186.

Generates `aiven_organizational_unit` resource and datasource.


ndk (approved):

docs/data-sources/organizational_unit.md
```diff
@@ -15,21 +15,28 @@ Gets information about an organizational unit.
...
   name      = "Example organizational unit"
-  parent_id = aiven_organization.main.id
 }
 ```
 
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
-### Required
+### Optional
```
vmyroslav, line 24: This looks a little suspicious. As we lost `name` as required attribute and we do not have `id` in this data source as an optional param

definitions/organizational_unit.yml
```diff
@@ -0,0 +1,46 @@
...
+    type: update
+    resultKey: account
+schema:
+  id:
+    description: The ID of this resource.
+  name:
+    description: The name of the organizational unit.
+  parent_id:
+    required: true
+    forceNew: false # the value can be account_id or organization_id. It should be possible to change it without recreating the resource.
```
vmyroslav, line 45: I do not what was the original idea, but looks like it was an issue before and an issue now. We mark it as not `ForceNew`, but ignore it within the update request, potentially creating a drift.

docs/data-sources/organizational_unit.md
byashimov: Hey. The `id` is optional here too. I have made it "either `id` or `name` required" because that's what the generator already can do — we have a datasource with the same logic. Usually, the `id` is required, not any other. So to make `name` only, we need to change the generator. Additionally:

1. I have improved the documentation previously, it now says "either field":
> - `id` (String) The ID of this resource. Exactly one of the fields must be specified: `id` or `name`.
> - `name` (String) The name of the organizational unit. Exactly one of the fields must be specified: `id` or `name`.
2. There is a validation rule that fail the plan stage https://github.com/aiven/terraform-provider-aiven/pull/2604/changes/BASE..d26b8e55984d3283c3a4e42e2ba295ebbf351d4d#diff-b6f488e37704c0db764f367073603c7b104830ee904b903386ab9803ab17650bR107
3. The example shows `name` defined https://github.com/aiven/terraform-provider-aiven/pull/2604/changes/BASE..d26b8e55984d3283c3a4e42e2ba295ebbf351d4d

It could have been `atLeastOne` instead of `exactlyOne` but that's requires additional validation making sure `name` belongs to the `id` and vice versa.

definitions/organizational_unit.yml
byashimov: The thing is, the `parent_id` should be able to go from `org_id` to `account_id` without recreating the resource. The `ForceNew` is detected by the generator, as it can't be updated. But we leave a way to changed it without consequences.
_Probably_ it must validate, that the org id belongs to the account id. Because for now, it does nothing about it on update. I mean, once the resource is created, it doesn't matter what do you put there.
So I can improve that by adding additional API call and making sure it belongs where it should: same account, same org.

PR #2605, opened 2025-12-11
refactor(mysql_database): generate

byashimov:
Resolves NEX-2187.

- Generates `aiven_mysql_database`
- Adds `termination_protection` support on adapter level


PLUGIN_MIGRATION.md
```diff
@@ -4,108 +4,108 @@
...
+|  95 | aiven_service_plan                          | yes    |     1 |
+|  96 | aiven_service_plan_list                     | yes    |     1 |
 |  97 | aiven_static_ip                             |        |     1 |
 |  98 | aiven_thanos                                |        |     2 |
 |  99 | aiven_transit_gateway_vpc_attachment        |        |     2 |
 | 100 | aiven_valkey                                |        |     2 |
 | 101 | aiven_valkey_user                           |        |     2 |
 +-----+---------------------------------------------+--------+-------+
-|     | TOTAL MIGRATED 13%                          | 24     |   178 |
+|     | TOTAL MIGRATED 16%                          | 28     |   178 |
```
byashimov, line 109: The `aiven_organizational_unit` PR didn't regenerate the report, because the report was merged right before the unit's PR. Hence, "migrated" counts 2 resources and 2 datasources (13 -> 16%).
Also sorted it alphabetically. It used to be: plugin true/false, a-z — a freshly migrated resource would shift it up in the list making the diff completely unreadable.

docs/data-sources/mysql_database.md
```diff
@@ -25,11 +25,22 @@ data "aiven_mysql_database" "example_database" {
...
+### Optional
+
+- `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 
 ### Read-Only
 
-- `id` (String) The ID of this resource.
-- `termination_protection` (Boolean) Client-side deletion protection that prevents the database from being deleted by Terraform. Enable this for production databases containing critical data. The default value is `false`.
+- `id` (String) Resource ID composed as: `project/service_name/database_name`.
+- `termination_protection` (Boolean, Deprecated) Client-side deletion protection that prevents the resource from being deleted by Terraform. **Resource can still be deleted in the Aiven Console.**. The default value is `false`. **Deprecated**: Instead use [`prevent_destroy`](https://developer.hashicorp.com/terraform/tutorials/state/resource-lifecycle#prevent-resource-deletion)
```
byashimov, line 39: This field is completely useless in datasource, but it remains for the backward compatibility.

internal/plugin/service/mysql/database/database_test.go
```diff
@@ -0,0 +1,183 @@
...
+					}
+					return rs.Primary.ID, nil
+				},
+				ImportStateCheck: func(s []*terraform.InstanceState) error {
+					if len(s) != 1 {
+						return fmt.Errorf("expected only one instance to be imported, state: %#v", s)
+					}
+					attributes := s[0].Attributes
+					if !strings.EqualFold(attributes["project"], projectName) {
+						return fmt.Errorf("expected project to match '%s', got: '%s'", projectName, attributes["project_name"])
```
vmyroslav: Guess it's a typo, but look like should be `attributes["project_name"]`

internal/plugin/service/mysql/database/zz_converter.go
```diff
@@ -0,0 +1,116 @@
...
+	if api.DatabaseName != nil && !state.DatabaseName.IsUnknown() {
+		state.DatabaseName = util.StringPointerValue(api.DatabaseName)
+	}
+	if api.Project != nil && !state.Project.IsUnknown() {
+		state.Project = util.StringPointerValue(api.Project)
+	}
+	if api.ServiceName != nil && !state.ServiceName.IsUnknown() {
+		state.ServiceName = util.StringPointerValue(api.ServiceName)
+	}
+	if api.TerminationProtection != nil || state.TerminationProtection.IsUnknown() {
```
vmyroslav, line 94: Based on the generator code I thought that this field should not be here, am I wrong?

internal/plugin/service/mysql/database/database_test.go
byashimov: The `project` is the right key. This part is migrated from the original test, and I guess it always worked as expected, and that's why we never got this error with `project_name`, which is invalid. Thanks.

internal/plugin/service/mysql/database/zz_converter.go
byashimov: The `flattenData` gets values from response _and_ modifiers like if they come from the response. This is correct.

byashimov:
> nit: looks like a good candidate for backward compatibility test

I have added compatibility test as a parallel test that shares the service.

PR #2608, opened 2025-12-15
feat(plugin): drop 404 warnings on DELETE

byashimov:
Resolves NEX-2171.

Removes warnings when DELETE gets 404 from the API.

### DIFF
--- a/internal/plugin/errmsg/retry.go
+++ b/internal/plugin/errmsg/retry.go
@@ -52,3 +52,11 @@ func WarnDiagError(diags diag.Diagnostics, f func(error) bool) diag.Diagnostics 
 	}
 	return diags
 }
+
+// DropDiagError drops DiagError from diagnostics if they match the filter function.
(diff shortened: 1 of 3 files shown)


internal/plugin/errmsg/retry.go
```diff
@@ -52,3 +52,16 @@ func WarnDiagError(diags diag.Diagnostics, f func(error) bool) diag.Diagnostics
 	}
 	return diags
 }
+
+// DropDiagError drops DiagError from diagnostics if they match the filter function.
+func DropDiagError(diags diag.Diagnostics, f func(error) bool) diag.Diagnostics {
```
ndk, line 57: How about using `slices.DeleteFunc`?
https://pkg.go.dev/slices#DeleteFunc

byashimov: Updated.

ndk (approved):

PR #2617, opened 2025-12-17
test: lock free shared resources

byashimov:
Resolves NEX-2212.

Speeds up backward compatibility tests with a shared resource.

- `BackwardCompatibilitySteps` can download providers while a shared service is created
- `acc.CreateTestService` can take an existing service name, but won't delete it
- `acc.CreateTestService` registers service removal as `t.Cleanup`

### DIFF
--- a/internal/acctest/acctest.go
+++ b/internal/acctest/acctest.go
@@ -21,6 +21,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/samber/lo"
+	"github.com/stretchr/testify/require"
 
 	"github.com/aiven/terraform-provider-aiven/internal/common"
 	"github.com/aiven/terraform-provider-aiven/internal/plugin/errmsg"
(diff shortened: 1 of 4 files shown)


internal/acctest/acctest.go
```diff
@@ -262,40 +263,72 @@ func WithPlan(plan string) CreateTestServiceOpt {
...
 
 // CreateTestService creates a test service that can be shared between tests.
-func CreateTestService(ctx context.Context, projectName, serviceName string, opts ...CreateTestServiceOpt) (func(), error) {
+// If the service already exists, it will be used as is and won't be deleted after the tests.
+func CreateTestService(t *testing.T, projectName, serviceName string, opts ...CreateTestServiceOpt) <-chan error {
 	client, err := GetTestGenAivenClient()
 	if err != nil {
-		return nil, err
+		require.NoError(t, err, "getting Aiven generated client failed")
 	}
```
ndk:
```suggestion
	require.NoError(t, err, "getting Aiven generated client failed")
```

ndk:
```suggestion
	require.NoError(t, err)
```

internal/acctest/acctest.go
```diff
@@ -262,40 +263,72 @@ func WithPlan(plan string) CreateTestServiceOpt {
...
 
-	req := &service.ServiceCreateIn{
-		ServiceName: serviceName,
-	}
+	// Creates cancel context to manage the goroutine lifecycle
+	ctx, cancel := context.WithCancel(t.Context())
+	t.Cleanup(cancel)
```
ndk: You don't need to worry about it.
https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/testing/testing.go;l=2459
https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/testing/testing.go;l=1556

internal/acctest/acctest.go
```diff
@@ -262,40 +263,65 @@ func WithPlan(plan string) CreateTestServiceOpt {
...
+		req := &service.ServiceCreateIn{ServiceName: serviceName}
+		for _, opt := range opts {
+			opt(req)
+		}
 
-	for _, opt := range opts {
-		opt(req)
-	}
+		_, err = client.ServiceCreate(ctx, projectName, req)
+		require.Error(t, err, "creating service %q failed", serviceName)
```
vmyroslav: I believe it's a typo and should be `NoError`

internal/acctest/acctest.go
```diff
@@ -262,40 +263,65 @@ func WithPlan(plan string) CreateTestServiceOpt {
 }
 
 // CreateTestService creates a test service that can be shared between tests.
-func CreateTestService(ctx context.Context, projectName, serviceName string, opts ...CreateTestServiceOpt) (func(), error) {
+// If the service already exists, it will be used as is and won't be deleted after the tests.
+func CreateTestService(t *testing.T, projectName, serviceName string, opts ...CreateTestServiceOpt) <-chan error {
```
vmyroslav, line 267: nit: should we lock here to prevent possible race conditions?

byashimov: It is meant to be used before a subtest is registered. So no locking needed. Or probably I don't follow. Could you elaborate please?

vmyroslav: Ok. I thought about the case if we want to use the same service across different parallel tests.

byashimov: I think it is easier then to use `t.Run`, i.e. turn a test into a subtest. Rather than solving global locks puzzle. 
Though there is a case when I would use the same instance to test the database and the user resources together simultaneously. It is frankly even makes sense, since Terraform would also run those in parallel.
But we run resource level tests in different builds.

PR #2619, opened 2025-12-18
feat(plugin): add min and max values to description

byashimov:
Resolves NEX-2214.

- Improves field description by adding min and max values
- Fixes min/max values merge
- Fixes `aiven_organization_project.base_port` min/max validation

### DIFF
--- a/internal/schemautil/userconfig/desc.go
+++ b/internal/schemautil/userconfig/desc.go
@@ -43,6 +43,10 @@ type DescriptionBuilder struct {
 	withPossibleValues []string
 	// withMaxLen is a flag that indicates if the maximum length should be included.
 	withMaxLen int
+	// withMinimum is a flag that indicates if the minimum value should be included.
+	withMinimum *int
+	// withMaximum is a flag that indicates if the maximum value should be included.
+	withMaximum *int
 	// withDefaultValue is a flag that indicates if the default value should be included.
 	withDefaultValue any
 	// withUseReference is a flag that indicates if the reference should be used.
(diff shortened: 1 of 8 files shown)


byashimov:
The org project requires 10000-30000 port, see:
- [Create](https://api.aiven.io/doc/#tag/Organizations/operation/OrganizationProjectsCreate)
- [Update](https://api.aiven.io/doc/#tag/Organizations/operation/OrganizationProjectsUpdate)
The values `1-65535` comes from response and is not validates by terraform.

internal/schemautil/userconfig/desc.go
```diff
@@ -111,6 +115,16 @@ func (db *DescriptionBuilder) MaxLen(length int) *DescriptionBuilder {
 	return db
 }
 
+func (db *DescriptionBuilder) Minimum(value int) *DescriptionBuilder {
```
vmyroslav, line 118: nit: I guess for the builder the `With` prefix looks like a standard for the fluid interface.
Also, we have a pointer there, so maybe it make sense to accept a pointer as well, to remove the value.

vmyroslav (approved):

internal/schemautil/userconfig/desc.go
byashimov: Sorry. I don't follow.
> I guess for the builder the With prefix looks like a standard for the fluid interface.

What do you mean?

> Also, we have a pointer there, so maybe it make sense to accept a pointer as well, to remove the value.

I'm having hard times to imagine a use case for that. Could you give me an example please?

vmyroslav: `Minimum` the intent is not totally clear, do we set the value or do we get. `WithMinimum` - looks more standard imho

byashimov: Got it. Other methods don't have "with" prefix. So I kept the naming for the consistency.

PR #2638, opened 2025-12-29
fix(adapter/resource): handle ModifyPlan destroy and null state safely

ndk:
Resolves NEX-2223.

- Generates aiven_pg_database

### DIFF
--- a/internal/plugin/adapter/resource.go
+++ b/internal/plugin/adapter/resource.go
@@ -410,21 +410,32 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
 		return
 	}
 
+	if req.Plan.Raw.IsNull() {
+		// There is no plan to modify if the resource is marked for deletion
+		return
+	}
+
 	var (
 		plan   = instantiate[M]()
-		state  = instantiate[M]()
 		config = instantiate[M]()
 		diags  = &rsp.Diagnostics
 	)
 
-	if !req.Plan.Raw.IsNull() {
-		// If resource is not marked for deletion.
-		diags.Append(req.Plan.Get(ctx, plan)...)
-		diags.Append(req.Config.Get(ctx, config)...)
-	}
-	diags.Append(req.State.Get(ctx, state)...)
+	diags.Append(req.Plan.Get(ctx, plan)...)
+	diags.Append(req.Config.Get(ctx, config)...)
 	if diags.HasError() {
 		return
+	}
+
+	var sharedState *T
+	if !req.State.Raw.IsNull() {
+		// During create planning, state is null. Let the ModifyPlan hook decide how to handle a nil state
+		state := instantiate[M]()
+		diags.Append(req.State.Get(ctx, state)...)
+		if diags.HasError() {
+			return
+		}
+		sharedState = state.SharedModel()
 	}
 
 	ctx, cancel, d := withTimeout(ctx, plan.TimeoutsObject(), timeoutRead)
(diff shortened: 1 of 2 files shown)


definitions/pg_database.yml
```diff
@@ -0,0 +1,38 @@
...
+  - id: ServiceDatabaseCreate
+    type: create
+  - id: ServiceDatabaseDelete
+    type: delete
+  - id: ServiceDatabaseList
+    type: read
+    resultKey: databases
+    resultListLookupKeys:
+      DatabaseName: database_name
+schema:
```
ndk: Added the schema overrides for `lc_collate` and `lc_ctype` to set the same default (`en_US.UTF-8`) as Aiven and the old SDKv2 resource. Without a schema default, omitted values plan as `null`, but the API returns `en_US.UTF-8` on read.

internal/plugin/adapter/resource.go
```diff
@@ -410,31 +410,43 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
 		return
 	}
 
+	if req.Plan.Raw.IsNull() {
```
ndk, line 413: `req.Plan.Raw.IsNull()` means a destroy plan, so there is no plan to modify. Hence we return early and don't call the resource `ModifyPlan` hook.

internal/plugin/adapter/resource.go
```diff
@@ -410,31 +410,43 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
...
-	}
-	diags.Append(req.State.Get(ctx, state)...)
+	diags.Append(req.Plan.Get(ctx, plan)...)
+	diags.Append(req.Config.Get(ctx, config)...)
 	if diags.HasError() {
 		return
 	}
 
+	var sharedState *T
+	if !req.State.Raw.IsNull() {
```
ndk, line 431: Also, `req.State.Raw.IsNull()` happens during create planning. Decoding a null state into a struct produces a conversion error, so we only read state when it exists.

internal/plugin/service/pg/database/database_test.go
```diff
@@ -0,0 +1,318 @@
...
+		serviceName,
+		acc.WithServiceType("pg"),
+		acc.WithPlan("startup-4"),
+		acc.WithCloud("google-europe-west1"),
+	)
+
+	client, err := acc.GetTestGenAivenClient()
+	require.NoError(t, err)
+
+	t.Run("backward compatibility test", func(t *testing.T) {
```
vmyroslav: I would , probably, add few more backward compatibility test cases to it, cause we have tricky moifiers for `lc_` fields. So maybe check when we have config with specified `lc_` and empty `lc_`.

definitions/pg_database.yml
```diff
@@ -0,0 +1,38 @@
+# yaml-language-server: $schema=.schema.yml
+location: internal/plugin/service/pg/database
+resource:
+  refreshState: true
+  removeMissing: true
+  terminationProtection: true
+  description: Creates and manages an [Aiven for PostgreSQL®](https://aiven.io/docs/products/postgresql) database.
+datasource:
+  description: Gets information about an Aiven for PostgreSQL® database.
```
vmyroslav: To me the previous description was a little more explicit. Cause there is a confusion between the database and service, imho.

> Gets information about a database in an Aiven for PostgreSQL® service

internal/plugin/adapter/resource.go
```diff
@@ -410,31 +410,43 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
 		return
 	}
 
+	if req.Plan.Raw.IsNull() {
+		// There is no plan to modify if the resource is marked for deletion
+		// TODO: should we allow ModifyPlan to handle deletions?
```
vmyroslav: Is this `todo` is still valid?

definitions/pg_database.yml
ndk: I've just repeated after `definitions/mysql_database.yml` :)
Do you prefer `Creates and manages a database in an Aiven for PostgreSQL® service.` as it was in SDSv2?

vmyroslav: Ok. Let's be consistent then.

PR #2639, opened 2025-12-29
tests(service_user): added tests to validate that password changes

vmyroslav:

- added more test steps to verify that `service_user password` rotation works as expected

Resolves: NEX-2224


internal/sdkprovider/service/kafka/kafka_user_test.go
```diff
@@ -74,6 +74,7 @@ resource "aiven_kafka_user" "foo" {
 					resource.TestCheckResourceAttr(resourceName, "password", "Test$1234"),
 					resource.TestCheckNoResourceAttr(resourceName, "password_wo"),
 					resource.TestCheckNoResourceAttr(resourceName, "password_wo_version"),
+					acc.TestAccCheckServiceUserPasswordMatches(resourceName, "Test$1234"),
```
byashimov: While all these checks are valid, I think you can significantly improve the UX here by naming validation with the use cases:

```go
acc.TestAccServiceUserHasCustomPassword(resourceName, "custom-password")
acc.TestAccServiceUserHasWOPassword(resourceName, "wo-password")
acc.TestAccServiceUserHasGeneratedPassword(resourceName) // e.g. prefix "AVN"
```

You could run checks that validate both local and remote states. For instance, WO check might prove `password` field is empty, and `password_wo_version` is not. These checks are same across all the types and _must_ have, hence having a specific function that does all the checks improves readability and maintainability.

Secondly, as I think `TestAccCheckServiceUserPasswordNotMatches` that checks the transition is a bit confusing. Because, while it does what it does, we don't state what it is equal for, we don't know for sure. Instead, asserting a statement "it now has a generated password" is much clearer.

Additionally, these checks might require specific prefix for a password, for instance `acc-wo-` for `password_wo`. This might provide extra layer of a certainty to improve the `TestAccCheckServiceUserPasswordNotMatches` case.

internal/sdkprovider/service/opensearch/opensearch_user_test.go
```diff
@@ -27,13 +27,108 @@ func TestAccAivenOpenSearchUser_basic(t *testing.T) {
...
+					resource.TestCheckNoResourceAttr(resourceName, "password_wo"),
+					resource.TestCheckResourceAttr(resourceName, "password_wo_version", "0"),
+					acc.TestAccCheckServiceUserPasswordExists(resourceName),
+				),
+			},
+			{
+				// final optional password
+				Config: testAccOpenSearchUserResource(rName, "FinalPass$456"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "service_name", fmt.Sprintf("test-acc-sr-%s", rName)),
```
byashimov: nit: `acctest.RandName` can improve maintainability for the service name.

internal/sdkprovider/service/kafka/kafka_user_test.go
vmyroslav: Upd

internal/sdkprovider/service/opensearch/opensearch_user_test.go
vmyroslav: upd

PR #2640, opened 2025-12-29
feat(changelog): add format command

byashimov:
Resolves NEX-2225.

The `CHANGELOG.md` is currently reformatted during schema updates, which overwrites manually added lines and makes `git blame` harder to use.

- Sets `PROVIDER_AIVEN_ENABLE_BETA=true` globally in `Taskfile.dist`
- Sets `PROVIDER_AIVEN_ENABLE_BETA=false` explictly for `task build`

### DIFF
--- a/TaskInternal.yml
+++ b/TaskInternal.yml
@@ -39,6 +39,7 @@ vars:
   SWEEP: global                          # Default scope for resource sweeper
   OLD_SCHEMA: .oldSchema.json            # Temp file to store previous provider schema for diffing
   CHANGELOG_CMD: "{{.GO_CMD}} run ./generators/changelog/..." # Command to run the internal changelog generation tool
+  CHANGELOG_FILE: CHANGELOG.md
 
   # --- Docs ---
   DEPRECATED_RESOURCES: "influxdb m3 redis"   # List of deprecated resources and data sources to exclude from docs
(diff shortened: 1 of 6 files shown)


TaskInternal.yml
```diff
@@ -125,6 +126,11 @@ tasks:
       - find docs -name "*.md" -type f -exec sed -i'' -e 's/[[:space:]]*$//' {} +
       - find docs -name "*.md" -type f -exec sed -i'' -e '${/^$/d}' {} +
 
+  fmt-changelog:
```
vmyroslav, line 127: Do we need here BETA flag here? As we discussed, it is not transitional and will fail if its not in shell

generators/changelog/main.go
```diff
@@ -82,10 +88,16 @@ func parseFlags() (*flags, error) {
 	flag.StringVar(&f.schemaFile, "schema", "", "schema file path (for save/diff)")
 	flag.StringVar(&f.changelogFile, "changelog", "", "changelog output file path")
 	flag.BoolVar(&f.reformat, "reformat", false, "reformat the whole changelog file")
+	flag.BoolVar(&f.format, "format", false, "formats the last section of the changelog")
```
vmyroslav:
```suggestion
	flag.BoolVar(&f.format, "format", false, "formats the unreleased section of the changelog")
```

TaskInternal.yml
byashimov: I set it on the top level. Ran schema updates from the branch, things look fine https://github.com/aiven/terraform-provider-aiven/pull/2643

vmyroslav (approved):

PR #2652, opened 2026-01-06
fix(adapter/resource): handle ModifyPlan destroy and null state safely

ndk:

Resolves NEX-2223.

Fixes `ModifyPlan` to safely handle destroy plans and create-time null state. The adapter now skips the custom `ModifyPlan` hook when the plan is null (destroy) and passes a nil state to the hook when there is no prior state, avoiding conversion errors.

## Why this way

Terraform Plugin Framework represents destroy plans as `req.Plan.Raw.IsNull()` and create planning can have `req.State.Raw.IsNull()`. Handling these cases prevents plan-time failures.

### DIFF
--- a/internal/plugin/adapter/resource.go
+++ b/internal/plugin/adapter/resource.go
@@ -387,23 +387,27 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
 	req resource.ModifyPlanRequest,
 	rsp *resource.ModifyPlanResponse,
 ) {
-	// Checks if termination protection is enabled for this resource
-	if a.resource.TerminationProtection && req.Plan.Raw.IsNull() {
-		// req.Plan.Raw.IsNull() means "marked for deletion":
-		// https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#resource-destroy-plan-diagnostics
-		enabled := false
-		rsp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("termination_protection"), &enabled)...)
-		if rsp.Diagnostics.HasError() {
-			return
+	if req.Plan.Raw.IsNull() {
+		// Checks if termination protection is enabled for this resource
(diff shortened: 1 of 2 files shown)


internal/plugin/adapter/resource.go
```diff
@@ -410,31 +410,42 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
 		return
 	}
 
+	if req.Plan.Raw.IsNull() {
+		// There is no plan to modify if the resource is marked for deletion
+		return
+	}
+
```
byashimov, line 416: Let's combine this with the block at 391:
```go
if req.Plan.Raw.IsNull() {
	if !a.resource.TerminationProtection {
        // Termination protection is not enabled
        return
    }

    // Validate TerminationProtection
}
```

internal/plugin/adapter/resource.go
```diff
@@ -412,29 +416,35 @@ func (a *resourceAdapter[M, T]) ModifyPlan(
...
+		}
+		sharedState = state.SharedModel()
+	}
+
 	ctx, cancel, d := withTimeout(ctx, plan.TimeoutsObject(), timeoutRead)
 	diags.Append(d...)
 	if diags.HasError() {
 		return
 	}
 	defer cancel()
```
byashimov, line 445: Since `req.State.Get` takes the context, we should be setting timeout before marshalling the state.

And if I may suggest the naming `stateOrNil` kinda emphasize it can be nil.

internal/plugin/adapter/resource_test.go
```diff
@@ -0,0 +1,1261 @@
+package adapter
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/aiven/aiven-go-client/v2"
```
byashimov, line 8: We should avoid using deprecated client.
UPD: I see it is required by the interface.

PR #2657, opened 2026-01-07
refactor(pg_database): generate

ndk:
Resolves NEX-2223.

Generates aiven_pg_database


internal/plugin/service/pg/database/database.go
```diff
@@ -0,0 +1,194 @@
...
+
+	if v, changed := normalize(plan.LcCollate, state.LcCollate); changed {
+		plan.LcCollate = v
+	}
+	if v, changed := normalize(plan.LcCtype, state.LcCtype); changed {
+		plan.LcCtype = v
+	}
+
+	return nil
+}
```
byashimov: The flattenModifier, expandModifier, modifyPlan can be safely removed. Default value works as expected.
Additionally, it makes sense to set `minLength: 1` to both fields.

internal/plugin/service/pg/database/database.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	})
+	if err != nil {
+		diags.Append(errmsg.FromError("Waiting for database deletion failed", err))
+		return diags
+	}
+
+	schemautil.ForgetDatabase(state.Project.ValueString(), state.ServiceName.ValueString(), state.DatabaseName.ValueString())
+	return diags
+}
```
byashimov, line 79: I can't see any issues with waiting the database in my local tests. Let's try to remove the whole thing and see if the error occurres in our nightly tests.

internal/plugin/service/pg/database/database.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	diags.Append(createView(ctx, client, plan)...)
+	if !diags.HasError() {
+		return diags
+	}
+
+	// We have already checked for the existence of the database.
+	// Getting a conflict here means the client retried the request.
+	return errmsg.DropDiagError(diags, avngen.IsAlreadyExists)
```
byashimov, line 34: `if diags.HasError()` is quite a pattern. Once you get used to it, you can't notice the exclamation mark right there on line 37. It took me a while to see it.

I would remove the whole block 37-39 and let diags go through `errmsg.DropDiagError`.

internal/plugin/service/pg/database/database_test.go
```diff
@@ -0,0 +1,396 @@
...
+					resource.TestCheckResourceAttr(datasourceName, "database_name", dbName),
+				),
+			}),
+		})
+	})
+
+	t.Run("backward compatibility test (empty lc_*)", func(t *testing.T) {
+		dbName := acc.RandName("compatibility-empty-lc")
+		config := testAccPGDatabaseWithDatasource(projectName, serviceName, dbName, `  lc_collate = ""
+  lc_ctype   = ""`)
```
byashimov: This test is invalid.

SDKv2 has no `nil` value, so it can't distinguish an empty string and `nil`. When TF got `""` it sent `nil`, see `NilIfZero` [here](https://github.com/aiven/terraform-provider-aiven/blob/18fde831e6282d2e8b58e3ddfb92f1483eae3dbd/internal/sdkprovider/service/pg/pg_database.go#L93). Apparently, `""` is a more of a "bug", because there is no such a value: `lc_collate = ""`.

I think, it is a bad UX, because if TF accepts that, then it must set it, right? Having `minLength` validation is an improvement.

internal/plugin/service/pg/database/database_test.go
```diff
@@ -0,0 +1,396 @@
...
+		acc.WithCloud("google-europe-west1"),
+	)
+
+	client, err := acc.GetTestGenAivenClient()
+	require.NoError(t, err)
+
+	t.Run("backward compatibility test (default lc_*)", func(t *testing.T) {
+		dbName := acc.RandName("compatibility")
+		config := testAccPGDatabaseWithDatasource(projectName, serviceName, dbName, "")
+		resource.Test(t, resource.TestCase{
```
byashimov:
```suggestion
		resource.ParallelTest(t, resource.TestCase{
```

All tests in this suite can run in parallel. I have tested that locally.

internal/plugin/service/pg/database/database.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		diags.Append(errmsg.FromError("Service is powered off", err))
+		return diags
+	}
+
+	diags.Append(readView(ctx, client, state)...)
+
+	if errmsg.HasDiagError(diags, avngen.IsNotFound) {
+		schemautil.ForgetDatabase(state.Project.ValueString(), state.ServiceName.ValueString(), state.DatabaseName.ValueString())
+	}
```
byashimov: You don't need to clean cache here, there is no case for that. Once TF exits, the whole cache is gone.

definitions/pg_database.yml
```diff
@@ -0,0 +1,38 @@
...
+  lc_collate:
+    optional: true
+    default: en_US.UTF-8
+    forceNew: true
+    description: Default string sort order (`LC_COLLATE`) of the database.
+  lc_ctype:
+    optional: true
+    default: en_US.UTF-8
+    forceNew: true
+    description: Default character classification (`LC_CTYPE`) of the database.
```
byashimov, line 38: Please set `minLength: 1` for both fields.

internal/plugin/service/pg/database/database.go
ndk: I'm not sure it's a good idea to change this behavior during the SDKv2 to Framework migration. I assume there was a reason this logic was added in the original implementation: https://github.com/aiven/terraform-provider-aiven/pull/2379/files#diff-73aa5d86ad0e85646bfe8b508214f08acb886b78d0ad2a11647a685120821bf7R168

My guess is that it can fail in a delete then create flow when recreating the same database name. I'd rather add a test that targets this case, because local/night runs may not catch it.

byashimov (approved):

PR #2661, opened 2026-01-08
fix(kafka_schema_configuration): handle error when Kafka Schema Registry is disabled

vmyroslav:

-  updated the `aiven_kafka_schema_configuration` resource to catch 403 errors during `Read` and `Delete` operations.
-  when a 403 error occurs, the provider now fetches the parent Kafka service configuration to verify if the Schema Registry is disabled
- updated acceptance tests to include lifecycle verification: enabling the feature, disabling it and re-enabling it.

Resolves: NEX-2245

### DIFF
--- a/internal/sdkprovider/service/kafkaschema/kafka_schema_configuration.go
+++ b/internal/sdkprovider/service/kafkaschema/kafka_schema_configuration.go
@@ -2,15 +2,24 @@ package kafkaschema
 
 import (
 	"context"
+	"errors"
+	"path/filepath"
+	"sync"
 
-	"github.com/aiven/aiven-go-client/v2"
+	avngen "github.com/aiven/go-client-codegen"
 	"github.com/aiven/go-client-codegen/handler/kafkaschemaregistry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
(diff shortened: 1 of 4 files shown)


byashimov (changes requested):
The solution LGTM, though:

1. Please update the changelog — this issue has a customer report, at least
2. I would replace the client with the new one
3. It is a common practice to have schema+topics 1:1, and we have tests with hundreds of topics. I propose to cache the `ServiceGet` response.

vmyroslav:
> The solution LGTM, though:
> 
> 1. Please update the changelog — this issue has a customer report, at least
> 2. I would replace the client with the new one
> 3. It is a common practice to have schema+topics 1:1, and we have tests with hundreds of topics. I propose to cache the `ServiceGet` response.

Updated

internal/sdkprovider/service/kafkaschema/kafka_schema_configuration.go
```diff
@@ -113,22 +131,77 @@ func resourceKafkaSchemaConfigurationRead(ctx context.Context, d *schema.Resourc
...
+	return false, nil
+}
+
+// isSchemaRegistryAPIError checks if the error is a 403 Forbidden error.
+// The Aiven API returns 403 when an optional feature (like Schema Registry) is disabled.
+func isSchemaRegistryAPIError(err error) bool {
+	var aivenErr aiven.Error
+	if errors.As(err, &aivenErr) {
+		return aivenErr.Status == 403
+	}
```
byashimov: Looks like, the error check is the only old client package usage. Let's remove it.

byashimov (approved):

PR #2665, opened 2026-01-09
docs(kafka): enable SASL auth example

staceysalamon-aiven:

Adds an example for using SASL auth for a Kafka service to be included in the Aiven docs.


examples/kafka/kafka_sasl_authentication/variables.tf
```diff
@@ -0,0 +1,26 @@
+variable "aiven_token" {
+  description = "Aiven token"
+  type        = string
```
vmyroslav, line 3: Lets mark token as sensitive, please

vmyroslav (approved):

PR #2667, opened 2026-01-12
refactor: store `DoOnce` result

byashimov:
Resolves NEX-2252.

Adds internal storage to `DoOnce` which is a common use case.

### DIFF
--- a/internal/schemautil/database.go
+++ b/internal/schemautil/database.go
@@ -48,16 +48,14 @@ func (w *DatabaseDeleteWaiter) Conf(timeout time.Duration) *retry.StateChangeCon
 }
 
 var (
-	initialDatabases     sync.Map
-	initialDatabasesCall DoOnce
-	ErrDbAlreadyExists   = fmt.Errorf("database already exists")
+	seenDatabases       sync.Map
+	serviceDatabaseList DoOnce[bool]
+	ErrDbAlreadyExists  = fmt.Errorf("database already exists")
 )
 
 func ForgetDatabase(projectName, serviceName, dbName string) {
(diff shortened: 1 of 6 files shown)


internal/schemautil/database.go
```diff
@@ -48,16 +48,14 @@ func (w *DatabaseDeleteWaiter) Conf(timeout time.Duration) *retry.StateChangeCon
...
+	ErrDbAlreadyExists  = fmt.Errorf("database already exists")
 )
 
 func ForgetDatabase(projectName, serviceName, dbName string) {
-	serviceKey := filepath.Join(projectName, serviceName)
-	dbKey := filepath.Join(serviceKey, dbName)
-	initialDatabasesCall.Forget(serviceKey)
-	initialDatabases.Delete(dbKey)
+	dbKey := filepath.Join(projectName, serviceName, dbName)
+	seenDatabases.Delete(dbKey)
```
vmyroslav, line 58: Should we also call `serviceDatabaseList.Forget(projectName, serviceName)`?

internal/schemautil/service.go
```diff
@@ -1130,21 +1125,16 @@ func ServicePoweredForget(project, serviceName string) {
...
+		return s.State == service.ServiceStateTypePoweroff, nil
+	}, project, serviceName)
 
-	isOn, _ := servicePoweredMap.Load(key)
-	if !isOn.(bool) {
+	if isOff {
 		return ErrServicePoweredOff
 	}
-	return nil
+	return err
```
ndk: Shouldn't it be something like this?
```suggestion
	if err == nil && isOff {
		return ErrServicePoweredOff
	}
	return err
```

PR #2830, opened 2026-04-09
feat(plugin): load examples from the OpenAPI spec

byashimov:
Resolves NEX-2450.

The request/response objects in the spec contain a full example that we can use for our snippets.

The next step would be to use those examples for the docs website.

### DIFF
--- a/generators/plugin/main.go
+++ b/generators/plugin/main.go
@@ -418,6 +418,11 @@ func fromOperationID(scope *Scope, operation *Operation, root *Item) error {
 			return err
 		}
 
+		err = loadExamples(root, path.Request.Content.ApplicationJSON.Example)
+		if err != nil {
+			return err
+		}
+
 		operation.Request = request
 	}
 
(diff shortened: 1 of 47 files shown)


html/pipeline-mockup.html
```diff
@@ -0,0 +1,544 @@
+<!DOCTYPE html>
```
vmyroslav: I guess it was included by mistake

byashimov: Thanks. Had rebased too often.

generators/plugin/main.go
```diff
@@ -845,3 +859,54 @@ func fillOptionalFields(parent *Item, item *Item, name string) error {
...
+	return nil
+}
+
+func castSafely(kind SchemaType, value any) (any, bool) {
+	var v any
+	var ok bool
+	switch kind {
+	case SchemaTypeString:
+		v, ok = value.(string)
+	case SchemaTypeInteger:
```
vmyroslav, line 900: Looks like all integer examples would be ignored here, because all numbers were casted to `float` by json marshaling

byashimov: True. But there is yaml definitions as well.

vmyroslav: Yes, but the question was: is it intentional to skip the integer examples from spec, like this:

`"pool_size": {"type": "integer", "minimum": 1, "maximum": 10000, "description": "Size of PGBouncer's PostgreSQL side connection pool"}`

vmyroslav (approved):

PR #2836, opened 2026-04-14
feat(plugin): generate example tf files

byashimov:
Resolves NEX-2454.

Generates examples from the spec.


byashimov:
merged with https://github.com/aiven/terraform-provider-aiven/pull/2838

PR #2846, opened 2026-04-21
fix(kafka_topic): change message timestamp fields type

byashimov:
- Change `aiven_kafka_topic` field `config.message_timestamp_after_max_ms` type to integer
- Change `aiven_kafka_topic` field `config.message_timestamp_before_max_ms` type to integer


byashimov:
These are new fields, and there is a type conversion https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion
Should do the job.

ndk (approved):

PR #2851, opened 2026-04-21
feat(byoc): add BYOC permission resource

d812k:
Resolves IP-881.

adds a tf replacement of the `avn byoc cloud permissions`, a necessary step for tf BYOC deployment


internal/plugin/service/byoc/permissions/permissions.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return readView(ctx, client, d)
+}
+
+func updatePermissions(ctx context.Context, client avngen.Client, d adapter.ResourceData) error {
+	err := upsertPermissions(ctx, client, d)
+	if err != nil {
+		return err
+	}
+	return readView(ctx, client, d)
```
vmyroslav: We have a `refreshState: true` that you can add to the definitions file, that would re-read the state for you.
With this flag looks like you can get rid of `createView` and `updateView` and reuse default generated in `zz_view`.

vmyroslav:
@d812k Thanks for this PR.

internal/plugin/service/byoc/permissions/permissions.go
d812k: thanks for the comment! I've updated the code

docs/resources/byoc_permissions.md
```diff
@@ -0,0 +1,64 @@
...
+  projects = ["my-project"]
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `accounts` (Set of String) Accounts.
```
vmyroslav, line 35: I dunno the internals of the resource, so just wanna double check. Are we sure that both accounts and projects should be required and there is no conflict there?

d812k: the lists are authorithative here as well, if account or project is left out it is not kept in place, but wiped. so in my opinion having it as required makes sense

vmyroslav (dismissed):

d812k:
@vmyroslav had to push again to fix the tests, could you please re-approve?

vmyroslav (approved):

PR #2854, opened 2026-04-22
fix(byoc): harden forceNew in aws entity resource

d812k:
Resolves: IP-882.

Mark `cloud_provider`, `cloud_region`, `deployment_model`, and `reserved_cidr` as `forceNew` on `aiven_byoc_aws_entity`. The OpenAPI spec advertises these fields in the update body, but the backend cannot change them in place


byashimov:
Hey @d812k 
I really think this change should happen on the backend when possible.
It's not Terraform only, but the go client is affected as well.

vmyroslav (dismissed):

d812k:
@byashimov could you please elaborate your comment. do you mean that these changes have to be implemented elsewhere?

vmyroslav (approved):

byashimov:
> @byashimov could you please elaborate your comment. do you mean that these changes have to be implemented elsewhere?

I mean, if that's possible, I would remove them from the request body in the spec. So we don't need to patch it in Terraform, and go client.

PR #2861, opened 2026-04-27
refactor(project_vpc): migrate to plugin

ndk:

Resolves: #NEX-2451.

Migrates `aiven_project_vpc` to the Plugin Framework.

This also adds two small generator features needed for this migration:
- `populateFieldsFromID` fills missing ID fields from the old composed ID, for example `project/project_vpc_id`.
- `onSuccess` runs custom code after a generated operation succeeds. `aiven_project_vpc` uses it to wait until VPC deletion is really finished.


definitions/aiven_project_vpc.yml
```diff
@@ -0,0 +1,25 @@
+# yaml-language-server: $schema=.schema.yml
+location: internal/plugin/service/vpc/projectvpc
+resource:
+  description: Creates and manages a VPC for an Aiven project.
+  removeMissing: true
+  refreshState: true
+  refreshStateWaiter: true
+clientHandler: vpc
```
vmyroslav, line 31: nit: I would transfer the descriptions from the old code, imho they were quite better then the defaults

vmyroslav (dismissed):

vmyroslav:
nit: do you see any benefits not to delete the old files in this PR?

generators/plugin/views_test.go
ndk, line 1: Tests in this file are written by LLM. However it looks ok to me.

internal/plugin/service/vpc/projectvpc/datasource.go
```diff
@@ -0,0 +1,200 @@
...
+		return fmt.Errorf("project name should be alphanumeric")
+	}
+	if v, ok := d.GetOk("vpc_id"); ok && v.(string) != "" {
+		_, err := schemautil.SplitResourceID(v.(string), 2)
+		if err != nil {
+			return fmt.Errorf("invalid vpc_id, should have the following format {project_name}/{project_vpc_id}: %w", err)
+		}
+	}
+
+	return nil
```
byashimov: These two could have had the `pattern` set to a regexp instead of this validation.

internal/plugin/service/vpc/projectvpc/datasource.go
```diff
@@ -0,0 +1,200 @@
...
+
+func dataSourceRead(ctx context.Context, client avngen.Client, d adapter.ResourceData) error {
+	var projectName, projectVPCID, cloudName string
+	if v, ok := d.GetOk("vpc_id"); ok && v.(string) != "" {
+		chunks, err := schemautil.SplitResourceID(v.(string), 2)
+		if err != nil {
+			return fmt.Errorf("error splitting vpc_id: %w", err)
+		}
+		projectName = chunks[0]
+		projectVPCID = chunks[1]
```
byashimov: This can be achieved using the `planModifier`.

internal/plugin/service/vpc/projectvpc/datasource.go
```diff
@@ -0,0 +1,200 @@
...
+			},
+			"state": datasourceschema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: userconfig.Desc("State of the VPC.").PossibleValuesString(vpc.VpcStateTypeChoices()...).Build(),
+			},
+		},
+		Blocks:              map[string]datasourceschema.Block{"timeouts": timeouts.Block(ctx)},
+		MarkdownDescription: "Gets information about the VPC for an Aiven project.",
+	}
+}
```
byashimov: I think, the moment we are forced to have the whole schema handwritten, it cancels the idea of delivering changes to Terraform. Moreover, the yaml file has no idea about these fields. That's sort of a secret knowledge, that one may debug for a while.

definitions/aiven_project_vpc.yml
```diff
@@ -0,0 +1,33 @@
...
+idAttributeComposed:
+  fields: [project, project_vpc_id]
+  populateFieldsFromID: true
```
byashimov: This functionality must be either in the adapter by default, or in the ResourceData. No need it to generate.

definitions/aiven_project_vpc.yml
```diff
@@ -0,0 +1,33 @@
...
+legacyTimeouts: true
+expandModifier: true
+schema:
+  cloud_name:
+    description: The cloud provider and region where the service is hosted in the format `CLOUD_PROVIDER-REGION_NAME`. For example, `google-europe-west1` or `aws-us-east-2`.
```
byashimov: According to the examples, the provided format is invalid.
I mean, it looks like it can be kebab-case only.

definitions/aiven_project_vpc.yml
byashimov: I guess, somewhere here https://github.com/aiven/terraform-provider-aiven/blob/14eac0ad12c9671bb18ac81379aa937dd9941dbc/internal/plugin/adapter/resourcedata.go#L373
Like split id and use "Set()" to set the values.

internal/plugin/service/vpc/projectvpc/projectvpc.go
```diff
@@ -48,32 +50,51 @@ func refreshStateWaiter(ctx context.Context, client avngen.Client, d adapter.Res
...
-			getRsp, err := client.VpcGet(ctx, project, vpcID)
+			rspGet, err := client.VpcGet(ctx, project, vpcID)
 			if avngen.IsNotFound(err) {
+				// The VPC is deleated completely.
 				return nil
 			}
+
 			if err != nil {
-				return retry.Unrecoverable(err)
+				return err
```
byashimov: Not sure why GET might be unrecoverable, so I have removed this logic.

vmyroslav (approved):

PR #2868, opened 2026-04-29
(adapter): split SchemaTypeMap and SchemaTypeObject treatment

d812k:
Resolves: IP-992.

fromTFValueAny collapsed SchemaTypeMap and SchemaTypeObject into one case and looked up each incoming key in sch.Properties. Map schemas don't populate Properties, so every key on a MapAttribute failed with unknown property `<key>`.
This PR splits two schema types into separate decode cases; Map now recurses into sch.Items, Object keeps the strict sch.Properties[key] lookup.

### DIFF
--- a/internal/plugin/adapter/marshalling_test.go
+++ b/internal/plugin/adapter/marshalling_test.go
@@ -1,0 +1,166 @@
+// Copyright (c) 2026 Aiven, Helsinki, Finland. https://aiven.io/
+
+package adapter
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/stretchr/testify/require"
+)
+
+// TestFromTFValueAny_Map covers the decode path for SchemaTypeMap.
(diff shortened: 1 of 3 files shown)


vmyroslav (approved):

byashimov:
Hey @d812k 
Thank you for your contribution. Could you please add a TF resource test? I mean, the map case. Thank you.

internal/plugin/adapter/marshalling_test.go
```diff
@@ -0,0 +1,166 @@
...
+			},
+		}
+		val := tftypes.NewValue(looseType, map[string]tftypes.Value{
+			"name":  tftypes.NewValue(tftypes.String, "alpha"),
+			"extra": tftypes.NewValue(tftypes.String, "surprise"),
+		})
+
+		_, err := fromTFValueAny(sch, val)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), `unknown property "extra"`)
```
byashimov, line 164: Oh, gosh. I need to fix this!

PR #2870, opened 2026-04-30
fix(plugin/adapter): return empty-but-known value for empty list/set

d812k:
Resolves: IP-925.

Encodes empty collections as `NewValue(<type>, []tftypes.Value{})`. This preserves the semantic distinction between "unset" (nil Go input -> null tftypes value) and "empty" (empty Go input -> empty-but-known tftypes value) and fixes the inconsistent result error.

toTFValue encoded empty List and Set collections as
  `tftypes.NewValue(<type>, nil)`, which produces a null value rather
  than an empty-but-known one. For resources with a Required Set or List
  attribute that the user supplies as empty (e.g.
  `aiven_byoc_permissions.accounts = []`), Terraform rejects the apply
  result with:

    Provider produced inconsistent result after apply
    ... .accounts: was cty.SetValEmpty(cty.String), but now null.

  The Aiven API call itself succeeds, but the provider fails the
  post-apply consistency check, so the resource ends up tainted in state
  while the backend holds the intended configuration -- blocking every
  subsequent plan/apply.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ nav_order: 1
 - Fix plugin-framework adapter decoding map attributes as objects
 - Change `aiven_byoc_aws_entity` resource field `deployment_model` (enum): add `hipaa`, `pci_dss`
 - Change `aiven_service_integration` field `integration_type` (enum): add `datahub_metadata_ingestion`
+- Fix plugin-framework adapter encoding empty lists and sets as null
 
 ## [4.55.2] - 2026-04-22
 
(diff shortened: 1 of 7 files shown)


byashimov:
Unfortunately, this fix breaks another case https://github.com/aiven/terraform-provider-aiven/actions/runs/25170154817/job/73819003781#step:7:126
We have two of them:
1. user did set `[]`, we set `nil`, you have fixed this part
2. user did not set the field (i.e. equal `nil`), the API returns `[]` (the broken case).
It is tricky.

PR #2875, opened 2026-05-05
fix(plugin): normalize empty containers in tfValue

byashimov:
Resolves NEX-2491.

- Lift `[]` / `{}` from config when missing in plan.
- Drop empty `[]` / `{}` from plan when missing in config.


byashimov:
This approach doesn't work, because the Read request doesn't have the config to read from https://github.com/hashicorp/terraform-plugin-framework/issues/878

byashimov:
Potentially, fixed by https://github.com/aiven/terraform-provider-aiven/pull/3005

PR #2881, opened 2026-05-08
feat(upgradepipeline): add upgrade step resource

ndk:
Resolves: NEX-2433

Add `aiven_upgrade_step` resource for managing Upgrade Pipeline step rules. The resource maps `[source project, service]` to `[destination project, service]` under an organization.

`last_validation` is intentionally removed for now. The generator currently emits it as a nested block, which makes Terraform treat API-populated values as a block count change after create.


definitions/aiven_upgrade_step.yml
```diff
@@ -0,0 +1,27 @@
+# yaml-language-server: $schema=.schema.yml
+beta: false # Endpoints are Experimental in Aiven API, but this resource should not require the beta provider gate.
```
vmyroslav: The upstream API marks `UpgradePipeline` endpoints as `Experimental`, we don't have any "gates" for it, but it would be just marked as a beta, which in my opinion is a good thing for a start.

vmyroslav:
nit: LGTM, but I would also add a sweeper to the new resource as a part of the PR.

definitions/aiven_upgrade_step.yml
byashimov: Agree. The generator detects experimental endpoints and marks resource as beta. Please remove the flag and let the generator do the job when endpoints are GA.

definitions/aiven_upgrade_step.yml
```diff
@@ -0,0 +1,27 @@
...
+    type: read
+  - id: UpgradePipelineStepUpdate
+    type: update
+rename:
+  destination_project_name: destination_project
+  destination_service_name: destination_service
+  source_project_name: source_project
+  source_service_name: source_service
+remove:
+  - last_validation
```
byashimov, line 18: This is a new resource, no need to rename the fields.
As I can see from the docs `destination_project_name` is the API native name.
Let's stick to the API naming.

definitions/aiven_upgrade_step.yml
```diff
@@ -0,0 +1,27 @@
...
+schema:
+  auto_validation_delay_days:
+    computed: true
+    useStateForUnknown: true
```
byashimov: Why `computed=true`? The [UpgradePipelineStepGet](https://api.aiven.io/doc/#tag/Upgrade_pipeline/operation/UpgradePipelineStepGet) has it in the response.
If the resource can live without the `schema` block, then I would remove it completely. Probably we will figure out how to set `useStateForUnknown` at some point.

byashimov (changes requested):

definitions/aiven_upgrade_step.yml
ndk: If the schema override is removed entirely, the generator makes it just `Optional`. The acceptance test then fails with `.auto_validation_delay_days: was null, but now cty.NumberIntVal(7)`, because the API returns the default value. That's why this attribute also needs to be `Computed`.

definitions/aiven_upgrade_step.yml
```diff
@@ -0,0 +1,20 @@
+# yaml-language-server: $schema=.schema.yml
+location: internal/plugin/service/upgradepipeline/step
+resource:
+  description: Creates and manages an Aiven Upgrade Pipeline step between source and destination services. This limited access feature requires contacting Aiven sales.
```
ndk: @staceysalamon-aiven I think you would like to have a look at this :)

staceysalamon-aiven: Thanks! This is the standard wording we normally use: `This feature is in the limited availability stage and subject to change. Contact the [Aiven sales team](http://aiven.io/contact) to request access.`

byashimov (dismissed):

definitions/aiven_upgrade_step.yml
byashimov: The field is optional, that's why it is not strictly detected as computed. Thanks for explaining.

vmyroslav (dismissed):

definitions/aiven_upgrade_step.yml
ndk: I've extended the definition with a new field `limited: true`
So, it should be added automatically. Please, have a look at `docs/resources/upgrade_step.md` in this PR

ndk:
Folks, I added `limited: true` support to the generator. Pls, check it out. :)

docs/resources/upgrade_step.md
```diff
@@ -0,0 +1,72 @@
...
+description: |-
+  Creates and manages an Aiven Upgrade Pipeline step between source and destination services.
+  This resource is in the limited availability stage and may change without notice. To enable this feature, contact the sales team http://aiven.io/contact
+---
+
+# aiven_upgrade_step (Resource)
+
+Creates and manages an Aiven Upgrade Pipeline step between source and destination services.
+
+**This resource is in the limited availability stage and may change without notice.** To enable this feature, contact the [sales team](http://aiven.io/contact)
```
staceysalamon-aiven: Looks good, just missing a period (.) at the end. 

```suggestion
**This resource is in the limited availability stage and may change without notice.** To enable this feature, contact the [sales team](http://aiven.io/contact).
```

staceysalamon-aiven (changes requested):

staceysalamon-aiven (dismissed):

definitions/.schema.yml
```diff
@@ -15,6 +15,11 @@ properties:
     $comment: |
       Indicates whether this resource or datasource is in beta status.
       If not specified, the beta status may be inferred from OperationID schemas.
+  limited:
```
byashimov: I think, we should name it `limitedAvailability` instead.

vmyroslav (dismissed):

byashimov (dismissed):

staceysalamon-aiven (approved):
The new LA notice looks good, thanks!

byashimov (approved):

vmyroslav (approved):

PR #2893, opened 2026-05-13
feat(plugin): generate datasource lookup

byashimov:
Resolves NEX-2518.

Until now, a data source lookup by an alternative key (e.g. `name` -> `id`) was implemented
through a hand-written `planModifier`. That approach has several drawbacks:

- `planModifier` makes an extra API call to resolve the id, then `readView` makes its own, so
  every lookup pays for two round-trips.
- It must be hand-written per resource, which is exactly what the generator is meant to avoid.
- The lookup body is bulky and copy-pasted across resources — a maintenance red flag.
- The list-endpoint `operationID` stays outside the YAML definition.
- The list endpoint carries metadata (descriptions, validations, etc.) we never surface to the
  schema.

This change moves the lookup into the generator. A `read` operation marked `datasourceLookup:
true` with `resultListLookupKeys` drives:

- an id-empty branch inlined into `readView` — when the id is missing, the generator calls the
  list op, finds the match via `adapter.FindOne`, and flattens it into state; when the id is set,
  control skips the branch and runs the canonical read directly (no extra round-trip, no separate
  view function);
- `ExactlyOneOf(id, ...)` / `RequiredTogether(...)` datasource config validators;
- the `Optional` flag on the alternative attributes;
- the example hint above the lookup block;
- the field-level description sentence.

When the lookup and the canonical read endpoints return different shapes (a directory of ids vs.
the actual payload), set `resultIDField: <GoField>` on the `datasourceLookup` op. The generator
then emits `d.Set(<id>, match.<GoField>)` and lets control fall through to the canonical read
body in the same `readView`, and keeps the lookup response out of the data source schema. Used
by `aiven_organization_user_list`.

Migrated `aiven_clickhouse_user`, `aiven_flink_application`, `aiven_organization`,
`aiven_organization_user_list`, and `aiven_organizational_unit` to the new pattern and removed
their hand-written `planModifier` files. Skill docs updated accordingly.


byashimov:
Found an issue, moving back to draft.

definitions/aiven_clickhouse_user.yml
```diff
@@ -24,9 +23,13 @@ operations:
...
     resultKey: users
-    resultListLookupKeysOr: true
     resultListLookupKeys:
       Uuid: uuid
+  - id: ServiceClickHouseUserList
+    type: read
+    datasourceLookup: true
+    resultKey: users
+    resultListLookupKeys:
       Name: username
```
byashimov, line 33: This change doesn't really play nice with "OR" case, creating duplicated call block. Probably we could revise this later if we gain more cases like this.

internal/plugin/service/clickhouse/database/zz_view.go
```diff
@@ -57,21 +56,13 @@ func readView(ctx context.Context, client avngen.Client, d adapter.ResourceData)
 	if err != nil {
 		return err
 	}
-	found := adapter.FilterIndex(rsp, func(i int) bool {
+	match, err := adapter.FindOne(rsp, func(i int) bool {
```
ndk, line 59: This looks a bit dangerous to me.

Before, when we didn't find anything in the list, we returned `http.StatusNotFound`. Now this responsibility moved to `adapter.FindOne`, and `FindOne` returns `adapter.ErrNotFound` instead.

I'm thinking about possible side effects here. What about `refreshState`? Its retries rely on `http.StatusNotFound`, not `adapter.ErrNotFound`. Could this make refresh fail immediately instead of retrying?

byashimov: Thanks. Added the new error to the check.

ndk (approved):

PR #2901, opened 2026-05-19
feat(kafka_topic): generate in plugin

byashimov:
Resolves NEX-1818.

Generates `aiven_kafka_topic` and migrates the Terraform Plugin Framework.


docs/resources/kafka_topic.md
```diff
@@ -101,10 +130,7 @@ Optional:
 
 Required:
 
-- `key` (String) Tag key. Maximum length: `64`.
-
-Optional:
-
+- `key` (String) Tag key. Length must be between `1` and `64`.
```
vmyroslav, line 135: I guess this change is not intentional and breaking

definitions/aiven_kafka_topic.yml
```diff
@@ -0,0 +1,134 @@
+# yaml-language-server: $schema=.schema.yml
+location: internal/plugin/service/kafka/topic
+version: 1
```
vmyroslav, line 3: > I didn't test this one manually. I guess it's better to doublecheck, I assume that it may not work for old versions without state upgraders.

UPD: checked the history, V0 was dropped at version 3 already, so should not be the case.

docs/resources/kafka_topic.md
byashimov: Hey. Thanks for double-checking. This is indeed a breaking change (the `key` is marked required now). Though I believe it always was "required":
1. https://api.aiven.io/doc/#tag/Service:_Kafka/operation/ServiceKafkaTopicCreate
2. https://api.aiven.io/doc/#tag/Service:_Kafka/operation/ServiceKafkaTopicUpdate

Both `key` and `value` are required for create and update.

vmyroslav:
JFYI: We had a tiny update in the docs, we need to transfer [these changes](https://github.com/aiven/terraform-provider-aiven/pull/2930) to the version

vmyroslav (dismissed):

ndk (dismissed):

ndk (approved):

PR #2907, opened 2026-05-25
feat(byok): add data sources for CMK accessors

mlatief:

Adds per-provider data sources for retrieving CMK accessors:
- aiven_cmk_accessor_aws
- aiven_cmk_accessor_azure
- aiven_cmk_accessor_gcp
- aiven_cmk_accessor_oci


byashimov (approved):
It took me a bit to realize that `CMKAccessorsList` does not return a list but an object. So technically it should work. Could you confirm please that these datasources work? Otherwise let's mark them `beta` just in case. Thank you.

PR #2917, opened 2026-06-01
fix(byoc): fix BYOC resources dependencies

mlatief:
Changes to resource `aiven_byoc_aws_provision`:

- When destroying a provisioned custom cloud environment, the CustomCloudEnvironmentDelete endpoint should be called to free the managed cloud resources before the IAM role and other network resources are destroyed. Now this operation is added to the resource.

- A custom cloud should be `active` before it can be used to create services on it, this adds a `refreshStateDesired` to the resource so it waits for the custom cloud environment to be 'active'.

- Add `custom_cloud_names` to this resource and remove it from `aiven_byoc_aws_entity`. The custom cloud names are only populated after provisioning, and services should depend on this resource rather than `aiven_byoc_aws_entity`, in order to attempt cleaning up the managed cloud resources only after all the services are destroyed.

Changes to resource `aiven_byoc_aws_entity`:

- Remove `aws_iam_role_arn` since it is not valid before the entity is defined first, and it is rather an input to provision the custom cloud.

- Also remove computed `errors`, `state`, and `update_time` as they would unncessarily invalidate this resource's state on every update, causing a cascade of unrelated updates to dependent BYOC resources.

Resolves: IP-998

### DIFF
--- a/internal/plugin/service/byoc/aws_entity/zz_resource.go
+++ b/internal/plugin/service/byoc/aws_entity/zz_resource.go
@@ -46,11 +46,6 @@ func resourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "Google account identifier.",
 			},
-			"aws_iam_role_arn": schema.StringAttribute{
-				MarkdownDescription: "Amazon Resource Name. Maximum length: `2048`.",
-				Optional:            true,
-				Validators:          []validator.String{stringvalidator.LengthAtMost(2048)},
-			},
 			"aws_subnets_bastion": schema.MapAttribute{
 				Computed:            true,
 				ElementType:         types.StringType,
@@ -91,11 +86,6 @@ func resourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "ID of a custom cloud environment.",
 			},
-			"custom_cloud_names": schema.SetAttribute{
-				Computed:            true,
-				ElementType:         types.StringType,
-				MarkdownDescription: "Cloud names that can be used to provision a service on this BYOC.",
-			},
 			"deployment_model": schema.StringAttribute{
 				MarkdownDescription: "Deployment model for the BYOC cloud. The possible values are `direct_ipsec_ingress`, `hipaa`, `ipsec_ingress`, `pci_dss`, `standard` and `standard_public`. Changing this property forces recreation of the resource.",
 				PlanModifiers:       []planmodifier.String{stringplanmodifier.RequiresReplace()},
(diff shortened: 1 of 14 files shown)


internal/plugin/service/byoc/aws_provision/provision.go
```diff
@@ -0,0 +1,35 @@
...
+	organizationID := d.Get("organization_id").(string)
+	customCloudEnvironmentID := d.Get("custom_cloud_environment_id").(string)
+	return retry.Do(
+		func() error {
+			cce, err := client.CustomCloudEnvironmentGet(ctx, organizationID, customCloudEnvironmentID)
+			if err != nil {
+				return retry.Unrecoverable(err)
+			}
+
+			if cce.State == byoc.CustomCloudEnvironmentStateTypeActive {
```
vmyroslav: I see few other states like: `CustomCloudEnvironmentStateTypeCreationFailed`, `CustomCloudEnvironmentStateTypeDeletionFailed`, etc. should we stop waiting in case if we got a signal that creation failed for any reason?

internal/plugin/service/byoc/aws_entity/zz_resource.go
```diff
@@ -46,11 +46,6 @@ func resourceSchema(ctx context.Context) schema.Schema {
 				Computed:            true,
 				MarkdownDescription: "Google account identifier.",
 			},
-			"aws_iam_role_arn": schema.StringAttribute{
```
vmyroslav, line 49: I see this field is still used in acceptance tests

internal/plugin/service/byoc/aws_provision/provision.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return retry.Do(
+		func() error {
+			cce, err := client.CustomCloudEnvironmentGet(ctx, organizationID, customCloudEnvironmentID)
+			if avngen.IsNotFound(err) {
+				// The CCE is deleted.
+				return nil
+			}
+
+			if cce.State == byoc.CustomCloudEnvironmentStateTypeDeleted {
```
vmyroslav, line 37: This one is risky, cause if the error is not `404`, it would lead to panic. We need to process an error first.

internal/plugin/service/byoc/aws_provision/provision.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/aiven/terraform-provider-aiven/internal/common"
+	"github.com/aiven/terraform-provider-aiven/internal/plugin/adapter"
+)
+
+func refreshStateWaiter(ctx context.Context, client avngen.Client, d adapter.ResourceData) error {
+	organizationID := d.Get("organization_id").(string)
+	customCloudEnvironmentID := d.Get("custom_cloud_environment_id").(string)
+	return retry.Do(
+		func() error {
+			cce, err := client.CustomCloudEnvironmentGet(ctx, organizationID, customCloudEnvironmentID)
```
vmyroslav: The 5XX errors are retried under the hood by client, however we notice quite often the eventual consistency from the API after the creation. I would suggest to exclude 404 errors from `Unrecoverable`

mlatief: at this point the CCE should have been at least existing, the backing resource should have been created already in `aws_byoc_aws_entity`, and this resource starts by binding the CCE with the IAM role that should have been created as well, so 404 error when calling `CustomCloudEnvironmentGet`  would be unexpected at this point. Do you think we should still retry even if we get 404 here?

mlatief: will retry if we get 404 👍

vmyroslav:
One small note:
The`deleteViewInternal` calls `CustomCloudEnvironmentDelete` and guarantees the CCE is torn down before the role is destroyed. There is also a nuance with **service → provision** ordering.

A service's `cloud_name` is a plain string. For Terraform to destroy services before this resource's CCE teardown, the user must explicitly reference `aiven_byoc_aws_provision.custom_cloud_names`. If
they hardcode the cloud name instead and on destroy Terraform may tear down `provision` concurrently with the services.

vmyroslav (dismissed):

vmyroslav (approved):

PR #2919, opened 2026-06-08
feat(plugin): remove computed blocks in resources

byashimov:
Resolves NEX-2592.

Resources can't have computed blocks in the Plugin Framework, yet we must keep them in datasources.


byashimov:
I had to update the schemas because of the breaking changes in the client.

ndk (approved):

PR #2947, opened 2026-06-24
feat(plugin): generate docs for plugin framework resources

byashimov:
Generated Plugin Framework resources previously relied on tfplugindocs to render their
documentation. tfplugindocs renders required nested blocks as optional and forced
workarounds in the generator (the "Required property." prefix), so the generated pages
did not always reflect the real schema.

This adds a dedicated documentation generator for Plugin Framework resources and data
sources. genDoc writes a full Markdown page (front matter, schema groups, nested
schemas, timeouts and import) per entity into docs-plugin/, while tfplugindocs keeps
rendering SDK resources and guides into docs-tfplugindocs/. A new docs task runs both
and merges the results into docs/ via tools/scripts/docs-merge.sh.

The registry front matter description must be plain text. Rather than depending on
tfplugindocs' MPL-2.0 mdplain package, mdplain.go provides an independent, single-pass
Markdown-to-text converter. It reads inline code spans verbatim, which keeps Kafka
topic and ACL patterns such as logs_*_foo_* intact where a regexp-based stripper would
corrupt them.

Root deprecation callouts (~>) now live in the doc generator instead of the shared
DescriptionBuilder, so the builder no longer needs the IsRoot toggle and always emits
the inline "Deprecated" suffix used by attributes.

Resolves NEX-2658.


examples_tests/plugin_generated_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build all || examples
+
```
byashimov, line 2: Failed on the CI.

vmyroslav (approved):

PR #2948, opened 2026-06-24
fix(service_user): preserve planned password on refresh

vmyroslav:

- preserve the user-supplied `password` from the plan during the post-apply
Resolves: NEX-2648


byashimov (dismissed):

internal/plugin/service/mysql/user/user.go
```diff
@@ -67,6 +67,11 @@ func flattenModifier(ctx context.Context, client avngen.Client) adapter.MapModif
 			dto["authentication"] = v
 		}
 
+		// password: use plan value if known, otherwise let the API value through.
+		if v := d.Get("password").(string); v != "" {
+			dto["password"] = v
+		}
+
```
byashimov, line 77: nit: extend the comment, that this is the eventual consistency (desperate) solution. Not a regular thing, I mean.

vmyroslav: upd

byashimov (approved):

PR #2953, opened 2026-06-26
feat(plugin): render block list max

byashimov:
Resolves NEX-2662.

### DIFF
--- a/generators/plugin/doc.go
+++ b/generators/plugin/doc.go
@@ -365,6 +365,14 @@ func docTypeLabel(def *Definition, entity entityType, item *Item, name string) (
 		} else {
 			b.WriteString("Block List")
 		}
+		if entity.isResource() && (item.IsObject() || item.MaxItems > 0) {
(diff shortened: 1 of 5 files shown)


ndk (approved):

ndk (dismissed):

generators/plugin/doc.go
```diff
@@ -365,6 +365,10 @@ func docTypeLabel(def *Definition, entity entityType, item *Item, name string) (
 		} else {
 			b.WriteString("Block List")
 		}
+		if entity.isResource() && (item.IsObject() || item.MaxItems > 0) {
+			// Obejcts have maxItems=1
```
ndk:
```suggestion
			// Obejcts have maxItems=1 by default
```

byashimov: Objects are strictly set to `1`. It's not default or changeable.

byashimov: @ndk updated

ndk (approved):

PR #2955, opened 2026-06-29
fix(kafka_topic): stop persistent changes when config matches API defaults

### LINKED ISSUE #2954: Persistent changes in Kafka Topic config when user-provided values are the same as the API defaults
## What happened?

When creating a Kafka topic with the `config` block containing values which match the API default values, these values are ignored from the API response during the Read process. This results in the same changes being applied during the next run.

For example, when applying the following `min_insync_replicas` and `retention_ms` config values, these repeatedly show as requiring changes.

```hcl
  # aiven_kafka_topic.topic["instance"] will be updated in-place
  ~ resource "aiven_kafka_topic" "topic" {
        id                     = "..."

      + config {
          + min_insync_replicas = "1"
          + retention_ms        = "604800000"
        }
    }
```

The values are in fact persisted in state, they just are ignored and the (flattened) API response is preferred.

```
tf state pull | jq -r '.resources[] | select(.type=="aiven_kafka_topic" and .name == "topic").instances[] | select(.index_key=="instance").attributes.config[0] | with_entries(select(.key | IN("min_insync_replicas","retention_ms")))'
```
```json
{
  "min_insync_replicas": "1",
  "retention_ms": "604800000"
}
```

Values are persisted in state, even when they match the default values for that config parameter.

eklesel:

For Kafka Topic config, forces TF to use user-provided values even if they're the same as the default values, stopping persistent changes during apply jobs/

Resolves: #2954.

## Why this way

This was re-uses values from state, and ignores the setting to null during the `Read` operation if the values match the default values.


vmyroslav (approved):

vmyroslav:
Hi @eklesel — thanks for the fix, the approach looks good and we'd like to get it merged. Two small blockers on our side:

1. **Branch is out of date with `main`** — needs a rebase (our branch protection requires branches to be up to date before merging).
2. **`commitlint` (a required check) is failing** — the commit message needs a Conventional Commits prefix. Something like: `fix(kafka_topic): stop persistent changes when config matches API defaults`

Could you rebase on the latest `main` and reword the commit?

Alternatively, if you enable **"Allow edits by maintainers"** on this PR, we can rebase and reword it for you. Either works — whichever is easier for you. Thanks again for the contribution!

eklesel:
Hi @vmyroslav, that should be done now! Cheers

PR #2966, opened 2026-07-02
docs: add project migration guide

staceysalamon-aiven:

Adds a guide on migrating to the `aiven_organization_project` resource.

### DIFF
--- a/templates/guides/update-deprecated-resources.md
+++ b/templates/guides/update-deprecated-resources.md
@@ -50,16 +50,20 @@ Import blocks are the recommended approach because you can:
 
 4. To preview the import, run `terraform plan`.
 5. To apply the import, run `terraform apply`.
-6. To remove the deprecated resource from Terraform's control, run:
+6. To remove the deprecated resource from Terraform's control, use `terraform state rm`. Preview the changes first using the `-dry-run` flag:
+
+      ```bash
+      terraform state rm -dry-run $(terraform state list | grep '^TYPE\.')
+      ```
+
+7. After verifying the previewed changes, apply the changes by removing the `-dry-run` flag:
 
       ```bash
       terraform state rm $(terraform state list | grep '^TYPE\.')
       ```
 
-   You can use the `-dry-run` flag to preview the changes before removing the deprecated resources.
-
-7. Remove the deprecated resource blocks from your configuration.
-8. To confirm that the resources were migrated and that there is no drift between the configuration and the state, run `terraform plan` again.
+8. Remove the deprecated resource blocks from your configuration.
+9. To confirm that the resources were migrated and that there is no drift between the configuration and the state, run `terraform plan` again.
 
 ## Migrate to `aiven_organization_billing_group`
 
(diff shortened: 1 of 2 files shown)


templates/guides/update-deprecated-resources.md
```diff
@@ -245,6 +245,63 @@ resource "aiven_organization_group_project" "example" {
...
+
+3. To preview the import, run `terraform plan`.
+4. To apply the import, run `terraform apply`.
+5. To remove the deprecated resource from Terraform's control, run:
+
+      ```bash
+      terraform state rm $(terraform state list | grep '^aiven_project\.')
+      ```
+  
+   You can use the `-dry-run` flag to preview the changes before removing the deprecated resources.
```
byashimov: I think, it's worth it to put `dry-run` in the snippet, and then suggest to remove it when the result looks legit. Feels like a safer path to me.

templates/guides/update-deprecated-resources.md
```diff
@@ -245,6 +245,63 @@ resource "aiven_organization_group_project" "example" {
...
+        parent_id        = "org1a23f456789"
+        billing_group_id = "00ab1234-5678-9cd0-1ef2-345678g9012a"
+        technical_emails = ["ops@example.com"]
+      }
+      ```
+
+      Where:
+       - `parent_id` is the ID of the organization or organizational unit that the project is in.
+       - `billing_group_id` is the same as `billing_group` in the `aiven_project` resource.
+       - `project_id` is the same as `project` in the `aiven_project` resource.
```
byashimov, line 301: I think, we should explain where one can find the `organization_id`. It is a new field.

templates/guides/update-deprecated-resources.md
```diff
@@ -245,6 +245,63 @@ resource "aiven_organization_group_project" "example" {
...
+This example shows you how to migrate a project to the new resource using the following `aiven_project` resource:
+
+```hcl
+resource "aiven_project" "example" {
+  project       = "example-project"
+  parent_id     = "org1a23f456789"
+  billing_group = "00ab1234-5678-9cd0-1ef2-345678g9012a"
+  technical_emails = ["ops@example.com"]
+}
+```
```
byashimov, line 270: This might fail with formatting (not aligned). Please try to run `task fmt`.

byashimov (changes requested):

byashimov (dismissed):

templates/guides/update-deprecated-resources.md
byashimov: I think this one wasn't resolved. It is not a blocker though. My thinking is: users often copy-paste-run, then realize it's gone. So better dry-run first.

staceysalamon-aiven: Yes, good call. I added this in and also a separate step for removing the flag just so it's clear. I think this might be in some other guides, so I'll look into that and update there, if so. Thanks!

byashimov (approved):

PR #2968, opened 2026-07-06
refactor(organization_vpc): migrate to plugin framework

ndk:
Resolves: NEX-2670


definitions/aiven_organization_vpc.yml
```diff
@@ -0,0 +1,48 @@
...
+flattenModifier: true
+operations:
+  - id: OrganizationVpcCreate
+    type: create
+    disableView: true
+  - id: OrganizationVpcGet
+    type: read
+  - id: OrganizationVpcDelete
+    type: delete
+    disableView: true
```
byashimov, line 26: This is weird, but user can have the display name updated https://api.aiven.io/doc/#tag/Organization_Vpc/operation/OrganizationVpcUpdate
And it might gain features in the future. That said, I propose to support the handler.

definitions/aiven_organization_vpc.yml
```diff
@@ -0,0 +1,48 @@
...
+  - display_name
+  - peering_connections
+  - pending_build_only_peering_connections
+schema:
+  cloud_name:
+    type: string
+    required: true
+    forceNew: true
+    example: aws-eu-west-1
+    description: The cloud provider and region where the service is hosted in the format `CLOUD_PROVIDER-REGION_NAME`. For example, `google-europe-west1` or `aws-us-east-2`.
```
byashimov, line 37: I'm wondering if we should support "clouds" array instead of the flat cloud+cidr pair? Deprecate the cloud name, add the "clouds".

I mean, if the clouds is an array, then terraform does not deliver the whole feature. This can be implemented as a separate PR.

UPD: I see maxItems=1, then there is no rush for sure, but I would ask the BE team what the plan was on this field. https://api.aiven.io/doc/#tag/Organization_Vpc/operation/OrganizationVpcCreate

Image

internal/plugin/service/vpc/organizationvpc/organizationvpc.go
```diff
@@ -0,0 +1,129 @@
...
+		dto["clouds"] = []any{
+			map[string]any{
+				"cloud_name":   d.Get("cloud_name"),
+				"network_cidr": d.Get("network_cidr"),
+			},
+		}
+		dto["peering_connections"] = []any{}
+		delete(dto, "cloud_name")
+		delete(dto, "network_cidr")
+		return nil
```
byashimov, line 32: nit: fields that are not in the request struct will be ignored by the json marshaler. You don't have to remove them manually.

internal/plugin/service/vpc/organizationvpc/organizationvpc.go
```diff
@@ -0,0 +1,129 @@
...
+func validateConfig(_ context.Context, _ avngen.Client, d adapter.ResourceData) error {
+	cidr := d.Get("network_cidr").(string)
+	if cidr == "" {
+		return nil
+	}
+
+	if _, _, err := net.ParseCIDR(cidr); err != nil {
+		return fmt.Errorf("expected \"network_cidr\" to be a valid CIDR Value, got %q: %w", cidr, err)
+	}
+	return nil
```
byashimov: I recommend to extend `.schemas.yml` to support `format=cidr` and then implement it as we had in SDKv2 https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation#IsCIDR
This can be reused then in may other resources.

byashimov: Here is an example of a custom validator https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/validators-custom#implement-a-custom-validator

docs/resources/organization_vpc.md
```diff
@@ -1,61 +1,61 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
 page_title: "aiven_organization_vpc Resource - terraform-provider-aiven"
 subcategory: ""
 description: |-
-  Creates and manages a VPC for an Aiven organization.
```
vmyroslav, line 6: If the API is still experimental I would keep it as beta for now, until further notice

ndk: Not anymore :)
https://github.com/aiven/go-client-codegen/pull/480/changes#diff-e84991228165a46c7172a58bade02e16960c1f3de16d2e4d585c4ff9b5b4b530R14

definitions/aiven_organization_vpc.yml
ndk: https://github.com/aiven/go-client-codegen/pull/516

definitions/aiven_organization_vpc.yml
ndk: Yeah, I'll create a task so we don't forget to clarify the future `clouds` with the BE team.

UPD: NEX-2674

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ nav_order: 1
 
 ## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
 
+- Migrate `aiven_organization_vpc` resource and data source to the Plugin Framework.
```
byashimov: Please add it's no longer beta.

generators/plugin/imports.go
```diff
@@ -60,6 +60,7 @@ const (
 	validatorPackage      = "github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	utilPackage           = "github.com/aiven/terraform-provider-aiven/internal/plugin/util"
 	adapterPackage        = "github.com/aiven/terraform-provider-aiven/internal/plugin/adapter"
+	validatorsPackage     = "github.com/aiven/terraform-provider-aiven/internal/plugin/validators"
```
byashimov: nit: let's go with customValidatorsPackage, because two lines above there is "validatorPackage".

internal/plugin/service/vpc/organizationvpc/organizationvpc.go
```diff
@@ -0,0 +1,121 @@
...
+
+func expandModifier(_ context.Context, _ avngen.Client) adapter.MapModifier {
+	return func(d adapter.ResourceData, dto map[string]any) error {
+		dto["clouds"] = []any{
+			map[string]any{
+				"cloud_name":   d.Get("cloud_name"),
+				"network_cidr": d.Get("network_cidr"),
+			},
+		}
+		dto["peering_connections"] = []any{}
```
byashimov, line 31: nit: this field is weird. I would leave a short comment telling it is required on the API, but we don't have it in the resource (and why if possible).

vmyroslav (approved):

byashimov (approved):

PR #2971, opened 2026-07-06
fix(kafka_topic): stop persistent changes when config matches defaults

vmyroslav:

Resolves: NEX-2671


byashimov (approved):

byashimov:
Though, would update the Changelog just in case.

ndk (approved):

PR #2979, opened 2026-07-10
refactor(aws_privatelink): migrate to plugin framework

ndk:
Resolves: NEX-2677


docs/resources/aws_privatelink.md
```diff
@@ -1,62 +1,64 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
 page_title: "aiven_aws_privatelink Resource - terraform-provider-aiven"
 subcategory: ""
 description: |-
-  Creates and manages an AWS PrivateLink for Aiven services https://aiven.io/docs/platform/howto/use-aws-privatelinks in a VPC.
+  Creates and manages an AWS PrivateLink for Aiven services https://aiven.io/docs/platform/howto/use-aws-privatelinks in a VPC. If this resource is missing (for example, after a service power off), it's removed from the state and a new create plan is generated.
```
byashimov, line 5: I know, this is the way it used to work, but it sounds a bit off to me.
If the ~VPC~ private link is missing, we probably shouldn't just ignore that.
Deletion might take time, though, and fail with a timeout error.
Which probably is an expected behaviour.

So I'm still suggesting to remove `removeMissing`, and let 404 raise the error.

UPD: vpc -> private link

definitions/aiven_aws_privatelink.yml
```diff
@@ -0,0 +1,25 @@
...
+    type: create
+  - id: ServicePrivatelinkAWSGet
+    type: read
+  - id: ServicePrivatelinkAWSUpdate
+    type: update
+  - id: ServicePrivatelinkAWSDelete
+    type: delete
+schema:
+  aws_service_name:
+    example: my-service
```
byashimov:
```suggestion
    example: my-aws-service-name
```
Just to emphasize it is in AWS, in contrast to the one in Aiven.

byashimov (changes requested):

docs/resources/aws_privatelink.md
ndk: `removeMissing` preserves the original behavior. It only handles 404 on `Read`.

byashimov: I mean, the `removeMissing` is for "sub" resources. Like a database in a service that might get deleted when the service has been restarted. Then TF would recreate those "sub" resources so users don't need to clean the state of hundreds of those "subs". That what the description states for.
~VPC~ private link is a different thing. There is no reason why it might get gone.

UPD: vpc -> private link

byashimov (dismissed):

internal/plugin/service/vpc/awsprivatelink/awsprivatelink_test.go
```diff
@@ -0,0 +1,213 @@
...
+	Principal string `envconfig:"AWS_PRIVATE_LINK_PRINCIPAL" required:"true"`
+	Region    string `envconfig:"AWS_DEFAULT_REGION" default:"eu-west-3"`
+
+	// Don't need to pass to tf file, this must be in env for the AWS provider.
+	AccessKeyID     string `envconfig:"AWS_ACCESS_KEY_ID" required:"true"`
+	SecretAccessKey string `envconfig:"AWS_SECRET_ACCESS_KEY" required:"true"`
+	SessionToken    string `envconfig:"AWS_SESSION_TOKEN"`
+}
+
+func TestAccAivenAWSPrivatelink_basic(t *testing.T) {
```
vmyroslav, line 22: We create to identical VPCs in parallel for the same project, which may collide and lead to the conflicts

byashimov (approved):

PR #2985, opened 2026-07-14
fix(database): enable pagination in pg and mysql database

vmyroslav:

-  Fixed `aiven_pg_database` to correctly read databases from services that have more than one page of databases.

Resolves: NEX-2683

### DIFF
--- a/internal/schemautil/database.go
+++ b/internal/schemautil/database.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/aiven/aiven-go-client/v2"
 	avngen "github.com/aiven/go-client-codegen"
+	"github.com/aiven/go-client-codegen/handler/service"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 
 	"github.com/aiven/terraform-provider-aiven/internal/common"
@@ -47,6 +48,9 @@ func (w *DatabaseDeleteWaiter) Conf(timeout time.Duration) *retry.StateChangeCon
 	}
 }
 
+// serviceDatabaseListPageSize is the page cap for the cursor-pagination (API max is 250).
+const serviceDatabaseListPageSize = 250
+
 var (
 	seenDatabases       sync.Map
 	serviceDatabaseList DoOnce[bool]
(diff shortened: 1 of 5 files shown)


internal/schemautil/database.go
```diff
@@ -58,6 +62,34 @@ func ForgetDatabase(projectName, serviceName, dbName string) {
...
+		// Guard against an API returning a non-advancing or cycling cursor.
+		if _, ok := seen[*rsp.Next]; ok {
+			return nil, fmt.Errorf("database list pagination cursor repeated, aborting to avoid an infinite loop: %q", *rsp.Next)
+		}
+		seen[*rsp.Next] = struct{}{}
+		query = [][2]string{maxItems, service.ServiceDatabaseListAfter(*rsp.Next)}
+	}
+	return databases, nil
+}
+
```
byashimov, line 92: A `ServiceDatabaseList` call might take seconds depending on what's going on on the backend (connecting the service, etc). Within those Terraform might remove/drop resources in parallel. By doing that it might shift the pagination: page 2 items appear on page 1, which has been cached, and won't be seen.

I don't have a better solution than having a mutex on _any_ mutating operation.
Or creating a "repository" like we did for the topics, which essentially is a mutex based solution.

Highlighting this as a potential _non-blocking_ issue, which _might_ be irrelevant, as I don't remember the whole flow.

vmyroslav: Good flag, but I don't see this case here.
The endpoint is using the ordering by name. Of course it behaviour is not guaranteed per contract, but...

Page 1 returns the smallest names, so `cursor` < `target` always → the `target` is always covered by a later page. A create can push it to a later page, but not skip it.

byashimov: Here how I see it:
1. read page 1 (the target db is on page 2)
2. delete databases on page 1, shift the pagination 
3. read page 2 (the target db is on page 1 now, so the database can't be found).

Granted each request takes ≈2—3 seconds there is a room for the race.
But as I said, I barely can imagine the case.

byashimov (approved):

PR #2995, opened 2026-07-21
WIP feat(adapter): add declarative deleteStateDesired delete poller

ndk:
Resolves: NEX-2693

The main idea is to separate sending a delete request from waiting for its result. We retry DELETE only when the API returns a conflict. After the API accepts the request, we stop sending DELETE and only wait for the resource to disappear or reach its final state. Other DELETE errors are returned immediately instead of being hidden by a timeout.

The polling also doesn't use the normal Terraform Read flow. A resource being deleted may return incomplete data, while a normal Read expects a complete object. Instead, a small generated reader checks only the state or whether the resource still exists.

Resources can describe a successful state, known failure states, and whether DELETE conflicts can be retried. Unknown states continue waiting, so adding a new intermediate backend state doesn't break the provider. This keeps the common logic small while still supporting differences between VPC, Azure Private Link, and BYOC APIs.


definitions/.schema.yml
```diff
@@ -66,18 +66,40 @@ properties:
...
-        additionalProperties:
-          type: string
+        additionalProperties: false
+        properties:
+          desired:
+            type: string
+            minLength: 1
+            $comment: |
+              Terminal value returned by the generated state-only observer. The delete poller
+              completes once it matches this value, or when the observer reports 404.
```
byashimov: It used to be a map, because "state" is not guaranteed, i.e. can be "status".

internal/plugin/adapter/resource.go
```diff
@@ -434,85 +435,112 @@ func (a *resourceAdapter) Delete(
...
+			if !deleteAccepted {
+				err := a.resource.Delete(ctx, a.client, rd)
+				switch {
+				case err == nil:
+					deleteAccepted = true
+				case IsNotFound(err):
+					return nil
+				case opts.RetryConflict && isAivenError(err, http.StatusConflict):
+					tflog.Debug(ctx, fmt.Sprintf("Delete of %q returned a conflict (will retry): %s", a.resource.TypeName, err))
+					return fmt.Errorf("%w: Delete returned 409 Conflict", ErrDeleteStatePending)
```
byashimov: I'm not sure if we should handle this situation. I mean, why it is not enabled by default? In which case you would not retry a conflict if you _know_ this is normal for long processed resources?

internal/plugin/adapter/resource.go
```diff
@@ -434,85 +435,112 @@ func (a *resourceAdapter) Delete(
...
 
-			// A 404 (gone) bubbles up; Delete's outer guard treats it as success.
-			if err != nil {
+			state, err := opts.Observe(ctx, a.client, rd)
+			switch {
+			case IsNotFound(err):
+				return nil
+			case err != nil:
 				return err
 			}
```
byashimov, line 514: You can bubble the `err` up and handle the "not found" in the outer scope. Both Delete() 404 and Get() 404 exit the loop, so no need to emphasize it's management. I mean, there is no case when you would ignore 404 in this function.

internal/plugin/adapter/resource.go
```diff
@@ -434,85 +435,112 @@ func (a *resourceAdapter) Delete(
...
+			defer func() {
+				drainWarnings()
+				lastAttemptWarnings = attemptWarnings
+			}()
+
+			if !deleteAccepted {
+				err := a.resource.Delete(ctx, a.client, rd)
+				switch {
+				case err == nil:
+					deleteAccepted = true
```
byashimov, line 502: We can have this one, though the Delete() must to be idempotent, hence make no harm. Usually, though.

generators/plugin/views.go
```diff
@@ -174,6 +199,84 @@ func genNewResource(entity entityType, def *Definition, item *Item, hasConfigVal
...
 }
 
+// genDeleteStateObserve generates a direct API read used only by the delete poller. It deliberately
+// bypasses readView, Flatten, and modifiers so a partial terminal response can still be observed.
+func genDeleteStateObserve(def *Definition, item *Item) (jen.Code, error) {
+	readOperations := make([]*Operation, 0, 1)
+	for _, op := range def.Operations {
+		if op.Type == OperationRead && !op.DatasourceLookup {
+			readOperations = append(readOperations, op)
+		}
```
byashimov, line 208: I would avoid this till we really run into it. Or what is the reason behind it?

byashimov (changes requested):

internal/plugin/adapter/resource.go
ndk: Just to reduce deletion-noise.

generators/plugin/views.go
ndk: We already need this for `aiven_organization_vpc`. During deletion we only need the state, so there is no reason to flatten the whole response and add another possible failure.

PR #2998, opened 2026-07-23
refactor(organization_user_group): migrate to plugin framework

byashimov:
Allow `name` to be updated in place instead of forcing resource recreation, as supported by the [UserGroupUpdate API](https://api.aiven.io/doc/#tag/Groups/operation/UserGroupUpdate).
Additional test step with `name` update has been provided.

Resolves NEX-2756.


byashimov:
The `description` in fact didn't have `ForceNew: true` but its description claimed that. That said, I logged the changed in the `CHANGELOG.md`.

vmyroslav (approved):

PR #3001, opened 2026-07-24
refactor(kafka_native_acl): migrate to plugin framework

byashimov:
Migrate `aiven_kafka_native_acl` to the Plugin Framework. Along the way, improve the generator: backtick `*` wildcards, preserve blank-line paragraph breaks in root descriptions, and drop unusable defaults on required non-computed fields (needed for the required `pattern_type`/`permission_type`).

Resolves NEX-2758.


byashimov:
Also adds `acl_id` computed field.

generators/plugin/main.go
```diff
@@ -962,6 +962,15 @@ func recalcDeep(def *Definition, item *Item) error {
 	item.Computed = ptrOrDefault(item.OverrideComputed, item.Computed)
 	item.ForceNew = ptrOrDefault(item.OverrideForceNew, item.ForceNew)
 
+	// The Plugin Framework rejects a default on a non-computed attribute, and an
+	// attribute cannot be both required and computed. Some OpenAPI schemas mark a
+	// field as required while also giving it a default (e.g. Kafka-native ACL
+	// pattern_type/permission_type). Honor the "required" contract and drop the
+	// unusable default rather than silently downgrading the field to optional.
+	if item.Required && !item.Computed {
+		item.Default = nil
```
ndk, line 971: Looks like we silently dropping the default that masks an invalid definition. I'd rather throw an error here.

byashimov: So if the spec is invalid, the error will always fail the pipeline.

internal/plugin/service/kafka/nativeacl/zz_resource.go
```diff
@@ -0,0 +1,125 @@
...
+			"acl_id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Kafka ACL ID.",
+			},
+			"host": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "the host or `*` for all hosts. Maximum length: `256`. Changing this property forces recreation of the resource.",
+				Optional:            true,
+				PlanModifiers:       []planmodifier.String{stringplanmodifier.RequiresReplace(), stringplanmodifier.UseStateForUnknown()},
+				Validators:          []validator.String{stringvalidator.LengthAtMost(256)},
```
ndk, line 31: The migration drops the non-empty constraint for host, principal, and resource_name. It means that we can set "" (an empty string) now.

byashimov: I mean, if that's the idea of the API authors, let it be.

ndk (approved):

PR #3004, opened 2026-07-27
chore(serviceuser): fixing password issues

vmyroslav:
## Changes

- Add `RefreshStateCheck` hook to the plugin adapter: a post-Create/Update validator for conditions that can't be expressed as exact attribute values in `RefreshStateDesired`
- Add `serviceuser.PasswordIsReady`: fails while `password` is empty, so the refresh keeps polling until the password propagates
- Wire `PasswordIsReady` into the service user resources
- Remove `refreshStateDelay: 15s` from the `kafka_user`, `mysql_user`, and `pg_user` definitions
- `slack-notify.yml`: set `retries: "0"` to stop the Workflow Builder trigger from re-delivering the webhook and posting duplicate messages
- `pg/connectionpool` acceptance test: bump plan `startup-4` → `startup-8`.

Resolves: NEX-2755

### DIFF
--- a/internal/plugin/serviceuser/serviceuser.go
+++ b/internal/plugin/serviceuser/serviceuser.go
@@ -72,8 +72,10 @@ func ResetPassword(ctx context.Context, client avngen.Client, d adapter.Resource
 }
 
 // PasswordIsReady returns an error if the password is empty while not in write-only mode.
-// The backend resets credentials asynchronously, so a freshly read user may have
-// an empty password until the change is fully propagated.
+// A freshly created service user can briefly be read back before the backend has populated
+// its password, so the first post-create read may observe an empty password. Retrying the
+// read until the password is present lets that settle.
+// Updates may see this as a blip but never a stale or old password.
 //
 // This must run against the same Read() whose result is written to state.
 // It cannot be a standalone pre-check.
--- a/internal/plugin/serviceuser/serviceuser.go
+++ b/internal/plugin/serviceuser/serviceuser.go
@@ -10,6 +10,7 @@ package serviceuser
 
 import (
 	"context"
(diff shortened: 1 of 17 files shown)


internal/plugin/serviceuser/serviceuser.go
```diff
@@ -70,6 +71,20 @@ func ResetPassword(ctx context.Context, client avngen.Client, d adapter.Resource
...
+	if _, ok := d.Schema().Properties["password_wo_version"]; ok && d.Get("password_wo_version").(int) != 0 {
+		// Write-only mode: PasswordFlatten removes the password from the state.
+		return nil
+	}
+	if d.Get("password").(string) == "" {
+		return errors.New("password is not available yet")
+	}
+	return nil
+}
+
```
byashimov, line 92: I think, it is worth to mention, that refreshState is the only place, where we can or should do this trick. Because a pre-check might hit the master replica, and then the Read() hit a read-replica.

byashimov, line 92: The refreshStateDelay also solved another issue: the remote state might still be not updated (due to replica-master lag).
So I guess, this now might fail with a stale password drift.

vmyroslav: Added the comment

vmyroslav: I never had this case. The main inconsistence was not due to the replication lag, but due to the internal password reset logic. During the update their can be time lags when the password is empty, but not stale.
UPD: I edited the comment to be more precise about real flow

byashimov (approved):

ndk (approved):

PR #3005, opened 2026-07-27
refactor(valkey_user): generate

byashimov:
- Migrate `aiven_valkey_user` resource and data source to the Plugin Framework
- Improve examples for array types

Resolves NEX-2760.


vmyroslav (dismissed):

internal/plugin/service/valkey/user/zz_resource.go
```diff
@@ -0,0 +1,154 @@
...
+			},
+			"type": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Account type.",
+			},
+			"username": schema.StringAttribute{
+				MarkdownDescription: "Service username. Maximum length: `64`. Changing this property forces recreation of the resource.",
+				PlanModifiers:       []planmodifier.String{stringplanmodifier.RequiresReplace()},
+				Required:            true,
+				Validators:          []validator.String{stringvalidator.LengthAtMost(64)},
```
ndk: It seems we've missed one more validation: https://github.com/aiven/terraform-provider-aiven/pull/3005/changes#diff-b0ab00992129801f72e2523623ccf746e0b15f0344e6a92d7ed287045bb9fd21L25
```
ValidateFunc: schemautil.GetServiceUserValidateFunc(),
```

Plus, I guess we let ppl set this field as an empty string.

internal/plugin/service/valkey/user/user.go
```diff
@@ -0,0 +1,102 @@
...
+
+func flattenACL(_ adapter.ResourceData, dto map[string]any) error {
+	accessControl, ok := dto["access_control"].(map[string]any)
+	delete(dto, "access_control")
+	if !ok {
+		return nil
+	}
+
+	for key, v := range accessControl {
+		if strings.HasPrefix(key, valkeyACLPrefix) {
```
ndk, line 87: Looks like the field retains a stale value when the API omits it. I think it would be safe to initialize all ACL fields to empty slices first.

ndk (dismissed):

byashimov:
Added an extra test for `foo = []` cases where prior the provider failed to take apart `[]` and `null` (not set).
Tested locally with a Valkey instance.

ndk (approved):

PR #3006, opened 2026-07-27
refactor(kafka_acl): generate

byashimov:
Migrate `aiven_kafka_acl` to the Plugin Framework.
    
Resolves NEX-2763.


ndk (dismissed):

definitions/aiven_kafka_acl.yml
```diff
@@ -0,0 +1,46 @@
...
+      Topic: topic
+      Username: username
+  - id: ServiceKafkaAclDelete
+    type: delete
+    resultKey: acl
+schema:
+  acl_id:
+    useStateForUnknown: true
+  username:
+    pattern: ^[-._*?A-Za-z0-9]+$
```
ndk, line 46:
```suggestion
    pattern: ^[-._*?A-Za-z0-9]+$
  project:
    pattern: ^[a-zA-Z0-9_-]*$
  service_name:
    pattern: ^[a-zA-Z0-9_-]*$
```

byashimov: I think, this is too much. These regexp might turn obsolete anytime, and we won't know that. I would rather stick to the spec.

ndk (approved):

PR #3014, opened 2026-07-30
test(serviceuser): pining current password behavior in test

vmyroslav:

- Added acceptance tests pinning how `aiven_pg_user` behaves when a service user's password is changed outside of Terraform

Resolves: NEX-2779


internal/plugin/service/pg/user/user_test.go
```diff
@@ -311,6 +315,312 @@ func TestAccAivenPGUser_basic(t *testing.T) {
...
+					// Pinned behavior: the plan stays empty because PasswordFlatten keeps
+					// the state password over the rotated one the API now returns.
+					PreConfig: func() {
+						newPassword := rotatedPassword
+						_, err := client.ServiceUserCredentialsModify(t.Context(), projectName, serviceName, userName,
+							&service.ServiceUserCredentialsModifyIn{
+								Operation:   service.ServiceUserCredentialsModifyOperationTypeResetCredentials,
+								NewPassword: &newPassword,
+							})
+						require.NoError(t, err)
```
byashimov: nit:
```
newPassword := rotatedPassword
```
this var is confusing, it does nothing, though adds some complexity for the reader

internal/plugin/service/pg/user/user_test.go
```diff
@@ -311,6 +315,312 @@ func TestAccAivenPGUser_basic(t *testing.T) {
...
+					// An explicit config change is the only path that resets credentials.
+					Config: testAccPGUserNoDatasource(projectName, serviceName, userName, healedPassword),
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr(resourceName, "password", healedPassword),
+						func(*terraform.State) error {
+							require.Eventually(t, func() bool {
+								password, err := apiPassword(userName)
+								return err == nil && password == healedPassword
+							}, 3*time.Minute, 2*time.Second, "password reset did not propagate to the API")
+							return nil
```
byashimov: Can probably wrap the whole func into a helper, so it easier to read.
I mean, it is always:
```golang
checkPassword(expectedPassword, "foo bar")
```

byashimov: I guess, you can also create `rotatePassword(foo)`. It is always the same.

byashimov (dismissed):

internal/plugin/service/pg/user/user_test.go
vmyroslav: Updated and extracted all into separate funcs

byashimov (approved):

PR #3015, opened 2026-07-30
refactor(flink_jar_application): generate

byashimov:
Migrate `aiven_flink_jar_application` to the Plugin Framework.
    
Read-only blocks, here `application_versions` and `current_deployment`, are now
rendered as attributes, because the Plugin Framework has no computed blocks: it
plans a block from the configuration, which a read-only value never has. Reading
them doesn't change, so the state and the expressions written for the blocks keep
working. That is what makes the migration backward compatible.
    
The `flink_user_config.custom_code` requirement moves from a comment in the
example into the resource description, so it shows up in the docs.
    
Resolves NEX-2782.


vmyroslav (dismissed):

ndk:
I fixed two computed state issues:
1. A new application has no versions, so the API returns `application_versions: []`, but the provider stored it as `null`. It now stays an empty list.
2.  After a deployment is deleted, the API omits `current_deployment`, but the provider kept the old deployment in state. The stale value is now removed.

vmyroslav (approved):

PR #3017, opened 2026-07-31
feat(plugin): support desired and failed refresh states

ndk:
Resolves: NEX-2789

This PR replaces `refreshState` and `refreshStateDesired` with one `refreshState` block. By default, the block checks the Terraform `state` field, but a resource can choose another read-only field. It can define several desired and failed values. A desired value completes the refresh, a failed value stops it with an error, and all other values stay pending.

Existing resources are moved to the new format. The adapter also keeps the resource state after a successful `Create` when the following refresh fails, so Terraform doesn't lose the created resource. Custom refresh checks can return `ErrRefreshStateFailed` to stop polling immediately.

### DIFF
--- a/definitions/.schema.yml
+++ b/definitions/.schema.yml
@@ -40,10 +40,36 @@ properties:
           Set to true to enable termination protection validation for this resource.
           The `termination_protection` attribute generated automatically.
       refreshState:
-        type: boolean
+        type: object
+        additionalProperties: false
+        properties:
+          attribute:
+            type: string
+            minLength: 1
+            default: state
+          desired:
+            type: array
+            minItems: 1
+            uniqueItems: true
+            items:
+              type: string
+          failed:
+            type: array
+            minItems: 1
+            uniqueItems: true
+            items:
+              type: string
+        dependentRequired:
+          attribute:
+            - desired
+          failed:
+            - desired
         $comment: |
-          Set to true to refresh the state after the resource is created/updated/deleted.
-          Example: refreshState: true
+          Refreshes state after Create and Update. An empty object performs one successful Read,
+          retrying transient 404/403 errors. When conditions are configured, `attribute` selects
+          the computed-only Terraform attribute to check and defaults to `state`. The attribute must
+          reach one of the `desired` values. A match against any `failed` value fails immediately.
+          Values outside both lists are treated as pending and retried until the operation timeout.
       refreshStateDelay:
         type: string
         example: 10s
@@ -52,20 +78,6 @@ properties:
           This is useful to avoid race conditions with the backend.
           For instance, when a new user is created, the backend might hit a read-replica.
           So we sleep for a few seconds to give the backend time to catch up.
-        dependencies:
-          - refreshState
-      refreshStateDesired:
-        type: object
-        additionalProperties:
-          type: string
-        $comment: |
-          Map of Terraform attribute names to the desired value the adapter must observe before
-          treating the post-Create/Update Read as successful. Each entry is checked after a Read
-          attempt during refreshState; mismatches retry under the same backoff as the standard
-          404/403 cases. Use this for resources whose API reports a transient state
-          (e.g. `state: ACTIVE`) without custom waiter logic.
-        dependencies:
-          - refreshState
       deleteStateDesired:
         type: object
         additionalProperties:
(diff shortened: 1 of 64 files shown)


generators/plugin/item.go
```diff
@@ -131,15 +131,19 @@ type Scope struct {
...
-	Description           string            `yaml:"description"`
-	DeprecationMessage    string            `yaml:"deprecationMessage,omitempty"`
-	TerminationProtection bool              `yaml:"terminationProtection,omitempty"`
-	RefreshState          bool              `yaml:"refreshState,omitempty"`
-	RefreshStateDelay     time.Duration     `yaml:"refreshStateDelay,omitempty"`
-	RefreshStateDesired   map[string]string `yaml:"refreshStateDesired,omitempty"`
+	Description           string                           `yaml:"description"`
+	DeprecationMessage    string                           `yaml:"deprecationMessage,omitempty"`
+	TerminationProtection bool                             `yaml:"terminationProtection,omitempty"`
+	RefreshState          map[string]RefreshStateCondition `yaml:"refreshState,omitempty"`
```
byashimov:
```suggestion
	RefreshState          *RefreshStateCondition `yaml:"refreshState,omitempty"`
```
I think, we don't need here a map.

internal/plugin/adapter/resource.go
```diff
@@ -37,19 +46,16 @@ type ResourceOptions struct {
 	// Example: ["project_id", "instance_name"] == "project-123/instance-456"
 	IDFields []string
 
-	// Whether to call Read after Create and Update operations.
-	// Implemented by refreshState; see that function for retry behavior.
-	RefreshState bool
+	// RefreshState enables Read after Create and Update when non-nil. An empty map performs a
+	// post-operation Read without checking attributes. Otherwise, every configured attribute must
+	// reach one of its Desired values; any Failed value terminates refresh with an error.
+	RefreshState map[string]RefreshStateCondition
```
byashimov: I think, this is what left from the previous change.

byashimov (changes requested):

internal/plugin/adapter/resource.go
ndk: Yeah, exactly. That's why it's sometimes tempting to reuse the generator types.

internal/plugin/adapter/resource.go
```diff
@@ -210,7 +216,7 @@ func (a *resourceAdapter) Create(
 		return
 	}
 
-	if a.resource.RefreshState {
+	if a.resource.RefreshState != nil {
```
byashimov, line 223: There was an issue, now, I think, it even worse: when desired state not reached, terraform exits with an error, and state is never recorded, next plan will show "create" plan. I propose to remove return on line 217-223 and set an explicit comment why it doesn't exit.

internal/plugin/adapter/resource.go
```diff
@@ -310,10 +327,24 @@ func (a *resourceAdapter) refreshState(ctx context.Context, rd ResourceData) err
...
-				state := rd.Get(key)
-				if !Equal(state, want) {
-					return fmt.Errorf("%w: expected %q to be `%v`, got `%v`", ErrRefreshStateDesired, key, want, state)
+			if hasStateCondition {
+				state, ok := rd.GetOk(stateField)
+				if ok && slices.ContainsFunc(condition.Failed, func(failed string) bool {
+					return Equal(state, failed)
+				}) {
+					return fmt.Errorf("%w: %q is `%v`", ErrRefreshStateFailed, stateField, state)
+				}
```
byashimov, line 382: So when "true", it returns `ErrRefreshStateFailed` which is in `isRefreshStateRetryable`.
I guess, there is an issue with tests as well.

definitions/.schema.yml
```diff
@@ -40,10 +40,30 @@ properties:
...
+              type: string
+          failed:
+            type: array
+            minItems: 1
+            uniqueItems: true
+            items:
+              type: string
+        dependentRequired:
+          failed:
+            - desired
```
byashimov, line 66: I think this is an equivalent of "dependencies", we not just use a new syntax?

byashimov (changes requested):

internal/plugin/adapter/resource.go
ndk: I think you might confuse `ErrRefreshStateFailed` with `ErrRefreshStateDesired` :)

byashimov: Indeed.

internal/plugin/adapter/resource.go
byashimov, line 402: I have pointed to the wrong place. ErrRefreshStateFailed is wrapped with RefreshStateCheck.

definitions/.schema.yml
ndk: At the moment `dependencies` is considered as "old syntax": https://json-schema.org/draft/2019-09/release-notes#semi-incompatible-changes

byashimov (approved):

PR #3033, opened 2026-08-11
fix(ci): checkout passed ref in lint jobs

ndk:
Resolves: NEX-2798

Prevent `Update Schemas` and `Prepare Release` workflows from validating main and reporting false failures. Otherwise we have such problems: https://github.com/aiven/terraform-provider-aiven/actions/runs/31463076571/job/93690888921

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -59,6 +59,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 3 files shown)


vmyroslav:
nit: Can we replace please: 
`if: needs.create_update_pr.outputs.pr_created` to
`if: needs.create_update_pr.outputs.pr_created == 'true'`
as part of this PR please

vmyroslav (approved):

vmyroslav (approved):

PR #3036, opened 2026-08-17
feat(flink): migrate jar application deployment and version

byashimov:
Resolves NEX-2803.


byashimov:
rebased

internal/plugin/service/flink/jarversion/jarversion.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return err
+	}
+
+	url, _ := fileInfo(d)["url"].(string)
+	if url == "" {
+		return fmt.Errorf("jar application version created without an upload url")
+	}
+
+	return uploadFile(ctx, d.Get("source").(string), d.Get(sourceChecksumField).(string), url)
```
vmyroslav, line 78: nit: this problem existed before the migration!
In case of failed upload we still have the stalled application version. Maybe we can do the compensation deletion

internal/plugin/service/flink/jarversion/jarversion.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	defer rsp.Body.Close()
+
+	b, err := io.ReadAll(rsp.Body)
+	if err != nil {
+		// This is a connection error or something else, not an API error
+		return fmt.Errorf("failed to read response: %w", err)
+	}
+
+	if len(b) > 0 {
```
vmyroslav, line 174: nit: this problem existed before the migration!
Maybe we should also check the status code

internal/plugin/service/flink/jardeployment/jardeployment.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return err
+	}
+
+	// A value stored by a previous apply is the one the deployment was created with.
+	if _, ok := d.GetOk("restart_enabled"); ok {
+		return nil
+	}
+
+	return d.Set("restart_enabled", true)
```
vmyroslav, line 53: nit: I guess it may bite in case of importing the resource with `restart_enabled=false`. The next apply will recreate it.

docs/resources/flink_jar_application_deployment.md
```diff
@@ -1,70 +1,61 @@
...
-  application_id = aiven_flink_jar_application.example.application_id
-  version_id     = aiven_flink_jar_application_version.example.application_version_id
+  project        = "my-project" // Force new
+  service_name   = "my-application" // Force new
+  application_id = "foo" // Force new
+  version_id     = "543e420d-aa63-43e8-b8e8-294a78c600e7" // Force new
+
+  // OPTIONAL FIELDS
+  entry_class        = "com.example.MyFlinkJob" // Force new
+  parallelism        = 1 // Force new
```
vmyroslav, line 26: During the import if this field is not mentioned in the config, it may lead to recreation of the resource.
I guess we should override it in config:
```
parallelism:
    computed: true
    dropDefault: true
    useStateForUnknown: true
```

internal/plugin/service/flink/jarversion/jarversion.go
byashimov: I hope it is handled somehow on the backend with a janitor. Would leave it this simple for now.

vmyroslav (dismissed):

vmyroslav (approved):

PR #3040, opened 2026-08-19
feat(vpc): migrate organization VPC peering resources to Plugin Framework

ndk:
Resolves: NEX-2750

Migrate `aiven_aws_org_vpc_peering_connection`, `aiven_azure_org_vpc_peering_connection`, `aiven_gcp_org_vpc_peering_connection` to the Plugin Framework. Mostly, the code is generated, but small custom parts are required to read cloud-specific values from `state_info`, show the `PENDING_PEER` setup warning, and delete connections using the backend peering ID, which is not part of the Terraform resource ID.


definitions/aiven_aws_org_vpc_peering_connection.yml
```diff
@@ -0,0 +1,71 @@
...
+    disableView: true
+remove:
+  - create_time
+  - peer_azure_app_id
+  - peer_azure_tenant_id
+  - peer_resource_group
+  - state_info
+  - update_time
+  - user_peer_network_cidrs
+  - vpc_peering_connection_type
```
byashimov, line 53: You can use globs here:
```
- "*_azure_*"
- "*_gcp_*"
```
This will help to filter out new fields, that might appear in the future.

Same for other resources.

definitions/aiven_aws_org_vpc_peering_connection.yml
```diff
@@ -0,0 +1,71 @@
...
+schema:
+  organization_vpc_id:
+    example: 1a2b3c4d-5e6f-7a8b-9c0d-1e2f3a4b5c6d
+  aws_account_id:
+    description: AWS account ID.
+  aws_vpc_id:
+    description: AWS VPC ID.
```
byashimov, line 60: It is better to stick to the spec, until you really need to override things, and this improvement doesn't look like an improvement. What was the original description?

definitions/aiven_aws_org_vpc_peering_connection.yml
```diff
@@ -0,0 +1,71 @@
...
+  aws_vpc_region:
+    required: true
+    description: The AWS region of the peered VPC. For example, `eu-central-1`.
+  aws_vpc_peering_connection_id:
+    type: string
+    computed: true
+    example: pcx-1234567890abcdef0
+    description: The ID of the AWS VPC peering connection.
+  state:
+    description: State of the peering connection.
```
byashimov, line 70: Here too. "State is state".

definitions/aiven_azure_org_vpc_peering_connection.yml
```diff
@@ -0,0 +1,74 @@
...
+    failed:
+      - DELETED
+      - DELETED_BY_PEER
+      - DELETING
+      - ERROR
+      - INVALID_SPECIFICATION
+      - REJECTED_BY_PEER
+  refreshStateDelay: 1s
+  deleteStateDesired:
+    state: DELETED
```
byashimov: I think, this can be an improvement for the generator:
- deleteStateDesired.state could be a list
- refreshState.failed could use it to fail
"Delete" statuses can be considered as failed, because in fact they just provide early exit condition.

definitions/aiven_gcp_org_vpc_peering_connection.yml
```diff
@@ -0,0 +1,68 @@
+# yaml-language-server: $schema=.schema.yml
+beta: true
```
byashimov: These are old resources. Consider to remove "beta" here and let it to rely on the spec's "experimental".

definitions/aiven_gcp_org_vpc_peering_connection.yml
```diff
@@ -0,0 +1,68 @@
...
+  peer_vpc:
+    example: my-vpc
+    description: Google Cloud VPC network name.
+  state:
+    description: State of the peering connection.
+  self_link:
+    type: string
+    computed: true
+    example: https://www.googleapis.com/compute/v1/projects/my-gcp-project/global/networks/my-vpc
+    description: Computed Google Cloud network peering link.
```
byashimov, line 67: UseStateForUnkown I guess might improve the UX here.

internal/plugin/service/vpc/awsorgvpcpeeringconnection/awsorgvpcpeeringconnection.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+			if stateInfo, ok := dto["state_info"].(map[string]any); ok {
+				if connectionID, ok := stateInfo["aws_vpc_peering_connection_id"].(string); ok {
+					dto["aws_vpc_peering_connection_id"] = connectionID
+				}
+			}
+
+			return nil
+		},
+		pluginvpc.PendingPeerWarning(ctx, "AWS"),
```
byashimov, line 27: The warning won't be published until Create/Update return the response. So if thi s meant to give an intermediate feedback, it won't work.
Eventually, this would produce "success" + bunch of outdated warnings, or "error message" + bunch of outdated warnings.
In order to give a feedback tflog must be used.

internal/plugin/service/vpc/azureorgvpcpeeringconnection/azureorgvpcpeeringconnection.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	_, err := client.OrganizationVpcPeeringConnectionDeleteById(
+		ctx,
+		d.Get("organization_id").(string),
+		d.Get("organization_vpc_id").(string),
+		connectionID,
+	)
+	return err
+}
```
byashimov, line 36: I'm suggesting to remove this delete views and add an improvement to the backlog about validating the fields. Sounds like every view should have done field validation.

byashimov (changes requested):

definitions/aiven_aws_org_vpc_peering_connection.yml
ndk: I would recommend keeping the override text, otherwise we have a too generic one: "AWS VPC ID, GCP VPC network name, Azure Virtual network name of the peered VPC, or UpCloud VPC ID"

definitions/aiven_aws_org_vpc_peering_connection.yml
ndk: Alas, the spec text is "Project VPC peering connection state" which cloud mislead users

definitions/aiven_azure_org_vpc_peering_connection.yml
ndk: Make sence, I'll do it in a separate PR. But I don't think we should mix it with `refreshState.failed`, as they cover different lifecycle stages.

definitions/aiven_gcp_org_vpc_peering_connection.yml
ndk: I checked the backend implementation. `self_link` isn't stable across all replacement paths: `to_vpc_network` contains the organization VPC ID, so replacing `organization_vpc_id` produces a different link. `UseStateForUnknown` would keep the old link in the replacement plan and could cause an inconsistent result after apply. I think keeping it unknown is safer here.

internal/plugin/service/vpc/awsorgvpcpeeringconnection/awsorgvpcpeeringconnection.go
ndk: Yeah, Terraform shows diagnostics only after `Create` finishes. But this isn't a progress message. `PENDING_PEER` is a successful final state for `Create`: Aiven has created the connection, but the user still needs to finish the setup in their cloud. We only return warnings from the last `Read`, so old warnings don't collect. SDKv2 worked in the same way. `tflog` is useful for progress or debug messages, but it wouldn't replace this user-facing warning.

internal/plugin/service/vpc/azureorgvpcpeeringconnection/azureorgvpcpeeringconnection.go
ndk: Yeah, let's add it to the backlog. But the generator cannot reliably order `peering_connection_id` because it's not a part of the composite Terraform ID. Also, `GCP` doesn't store this ID in the state, so it must read the VPC and find the API ID before `Delete`. I suggest keeping the custom views for now.

byashimov (approved):

PR #3041, opened 2026-08-19
refactor(plugin): flatten resource state polling YAML

byashimov:
Replace nested refreshState and deleteStateDesired maps with stateAttribute, refreshStateExists/Desired/Failed, and deleteStateGone/Desired so exists vs value waits are explicit and 404 handling is implied by the desired lists.

Resolves NEX-2808.

### DIFF
--- a/internal/plugin/adapter/resource.go
+++ b/internal/plugin/adapter/resource.go
@@ -48,9 +48,9 @@ type ResourceOptions struct {
 	// Example: ["project_id", "instance_name"] == "project-123/instance-456"
 	IDFields []string
 
-	// RefreshState enables Read after Create and Update when non-nil. An empty condition performs a
-	// post-operation Read without checking an attribute. Otherwise, Attribute must reach one of its
-	// Desired values; any Failed value terminates refresh with an error.
+	// RefreshState enables Read after Create and Update when non-nil. 404/403 are retried until
+	// the resource exists. An empty condition then completes; otherwise Attribute must reach one
+	// of its Desired values, and any Failed value terminates refresh with an error.
 	RefreshState *RefreshStateCondition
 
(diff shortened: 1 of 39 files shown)


internal/plugin/adapter/resource.go
```diff
@@ -530,13 +530,16 @@ func (a *resourceAdapter) Delete(
...
 type DeleteStateOptions struct {
-	// Desired maps attribute names to the terminal value the poller must observe after Delete. Empty
-	// means there are no attribute conditions, so the poll completes only once the resource is gone (404).
-	Desired map[string]string
+	// Attribute is the Terraform attribute to check after Delete. Required when Desired is non-empty.
+	Attribute string
+	// Desired contains successful terminal attribute values. A match against any value completes
+	// the delete. Empty means there are no attribute conditions, so the poll completes only once
+	// the resource is gone (404).
+	Desired []string
```
vmyroslav, line 542: nit: now that `DeleteStateOptions` has the same `Attribute/Desired` shape as `RefreshStateCondition`, we could mirror the guard `refreshState` has. With `Desired` set but `Attribute` empty, `rd.GetOk("")`, may panic. This isn't reachable from generated specs but may be a footgun for handwritten `ResourceOptions`.

vmyroslav (dismissed):

ndk (dismissed):

internal/plugin/adapter/resource.go
byashimov: Agreed. `deleteState` now rejects `Desired` without `Attribute` the same way `refreshState` does, so handwritten `ResourceOptions` fail before any Delete/Read instead of hitting `GetOk("")`.

vmyroslav (approved):

